PR #33, opened 2015-02-02
Some minor changes

ravenblue12:
* Change the function name from enable to enable_mode, because according to the function name config_mode.
* Changed the default enable prompt check string.


ravenblue12:
* Change the function name from enable to enable_mode, because according to the function name config_mode.
* Changed the default enable prompt check string.
* Added code to add respond to uncommited changes in IOS XR.
* speedup commands

ktbyers:
Rene,

I am going to reject this pull request.
1. enable() to enable_mode() change I don't want to do because it would break existing code and IMO is not worth doing.

2. I will manually make this change and push it up (for the '(config)' check_mode problem).
     >>> def check_config_mode(self, check_string=')#'):

3. I will manually make the delay_factor change on find_prompt and disable_paging.  On send_command, left the max_loops and time.sleep(2) inside the while loop the same.  The max_loops should almost never occur since it is a catch all for very large output results (so changing this doesn't help the performance in a meaningful way).  I think some devices will break if I change time.sleep to 1 (in the while loop).  You can always call send_command() and overwrite the defaults.

4. On Cisco XR, not sure that is the behavior we should do.  Will discuss in the GitHub issue.

Issue #62, opened 2015-03-07
Strip Additional Juniper Prompts


supertylerc:
There are Juniper prompts that are currently not stripped from output.  See below.

```
 ❯ ssh 10.10.1.1
--- JUNOS 12.1X44-D25.5 built 2013-10-24 20:59:21 UTC
No alarms currently active

{primary:node0}
tyler@fw1.example.com> start shell
% rlogin -T node1
�
--- JUNOS 12.1X44-D25.5 built 2013-10-24 20:59:21 UTC

No alarms currently active

{secondary:node1}
tyler@fw1.example.com>
```

The method `strip_context_items()` does not currently strip these from output.  I will submit a PR for this.

supertylerc:
It looks like the prompt isn't being stripped after running a command as well:

```
>>> ip_interfaces = dev.send_command("show interfaces terse | grep inet")
>>> print(ip_interfaces)

ge-0/0/0.0              up    up   inet     10.0.2.15/24
sp-0/0/0.0              up    up   inet
                                   inet6
sp-0/0/0.16383          up    up   inet     10.0.0.1            --> 10.0.0.16
ge-0/0/1.0              up    up   inet     172.16.0.1/31
ge-0/0/2.0              up    up   inet     192.168.0.200/24
lo0.16384               up    up   inet     127.0.0.1           --> 0/0
lo0.16385               up    up   inet     10.0.0.1            --> 0/0

ansible@apples>
>>>
```

supertylerc:
This should've been mostly fixed by #63.  The only thing that isn't fixed is the presence of the final prompt.

Ragsboss:
Hi Tyler, I'm seeing for EOS, the final prompt is present - but not always. Is this a known issue? I'm issuing a command that's producing JSON output and when the prompt is present, JSON parsing fails. We are using version 0.4.3 and target device is 

```
Arista DCS-7050QX-32-F
Hardware version:    02.11
Software image version: 4.16.6M
```

Any help is appreciated

ktbyers:
@Ragsboss Please open a separate issue on this. You need to test the issue using Netmiko 1.1.

Issue #55, opened 2015-03-07
Juniper doesn't implement `check_config_mode()`


supertylerc:
The Juniper methods do not implement `check_config_mode()`.  This causes unexpected behavior and unnecessary duplication of code when trying to implement other features.  The current `config_mode()` method in the Juniper file, for example, could be eliminated and a `config_mode(config_command="configure")` (from **ssh_connection.py**) could be used instead.  I can work on this later, but here are a few notes.

The existing `check_config_mode()` method should be usable like this: `check_config_mode(check_string='[edit')`.  That should enable certain methods to be written like this:

```python
def config_mode(self):
    '''
    First check whether currently already in configuration mode.

    Enter config mode (if necessary)
    '''

    if not self.check_config_mode(check_string='[edit'):
        output += self.send_command('configure\n', strip_prompt=False, strip_command=False)
        if 'unknown command' in output:
            raise ValueError("Failed to enter configuration mode")
        if not '[edit' in output:
            raise ValueError("Failed to enter configuration mode")

    return output
```

Other methods could be written as well to take advantage of the `check_config_mode()` method, such as `exit_config_mode()`.  `commit()` could probably be rewritten to use `check_config_mode()` as well as it doesn't make much sense to check to see if you're in config mode, then enter config mode if you're not, then commit the config.  That's typically not a valid workflow, and there typically isn't a use case for committing nothing.  Maybe there's one I haven't considered, though.

supertylerc:
This is done.  I'll submit a PR later.

ktbyers:
Okay, I think is done. The commit() workflow issue we can just address in the commit() design issue.

PR #63, opened 2015-03-08
refactor, pep257, pep8

supertylerc:
Testing all methods:

```python
 ❯ python                                                                                                                                                                                       [15:59:03]
Python 2.7.9 (default, Dec 13 2014, 15:13:49)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.56)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import netmiko
>>> params = {"device_type": "juniper",
...           "ip": "ngfw1",
...           "username": "ansible",
...           "password": "Ansible"}
>>> dispatcher = netmiko.ssh_dispatcher(device_type=params["device_type"])
>>> dev = dispatcher(**params)
SSH connection established to ngfw1:22
>>> results = dev.config_mode()
>>> print(results)
configure
Entering configuration mode
Users currently editing the configuration:
  root terminal p0 (pid 1295) on since 2015-03-07 23:10:57 UTC, idle 00:19:03
      [edit]

[edit]
ansible@bananas#
>>> results = dev.send_command("set system host-name apples")
>>> print(results)


>>> results = dev.commit()
>>> print(results)
commit and-quit
commit complete
Exiting configuration mode

ansible@apples>
>>> results = dev.config_mode()
>>> print(results)
configure
Entering configuration mode
Users currently editing the configuration:
  root terminal p0 (pid 1295) on since 2015-03-07 23:10:57 UTC, idle 00:19:37
      [edit]

[edit]
ansible@apples#
>>> results = dev.send_command("run show interfaces terse | match inet")
>>> print(results)

ge-0/0/0.0              up    up   inet     10.0.2.15/24
sp-0/0/0.0              up    up   inet
                                   inet6
sp-0/0/0.16383          up    up   inet     10.0.0.1            --> 10.0.0.16
ge-0/0/1.0              up    up   inet     172.16.0.1/31
ge-0/0/2.0              up    up   inet     192.168.0.200/24
lo0.16384               up    up   inet     127.0.0.1           --> 0/0
lo0.16385               up    up   inet     10.0.0.1            --> 0/0

[edit]
ansible@apples#
>>> results = dev.exit_config_mode()
>>> print(results)
exit
Exiting configuration mode

ansible@apples>
>>> results = dev.disconnect()
>>> print(results)
None
>>> ^D
```

### DIFF
--- a/netmiko/juniper/juniper_ssh.py
+++ b/netmiko/juniper/juniper_ssh.py
@@ -1,96 +1,111 @@
 import time
 import re
 
-from netmiko.base_connection import BaseSSHConnection
+from netmiko.ssh_connection import SSHConnection
 
-class JuniperSSH(BaseSSHConnection):
+class JuniperSSH(SSHConnection):
+    """Implement methods for interacting with Juniper Networks devices.
+
+    Subclass of SSHConnection.  Disables `enable()` and `check_enable_mode()`
+    methods.  Overrides several methods for Juniper-specific compatibility.
+    """
+
+    def enable(self):
+        """Disable superclass `enable()` method."""
+        raise AttributeError("'JuniperSSH' object has no attribute 'enable'")
+
+    def check_enable_mode(self):
+        """Disable superclass `check_enable_mode()` method."""
+        raise AttributeError("'JuniperSSH' object has no attribute \
+                             'check_enable_mode'")
 
     def session_preparation(self):
-        '''
-        Prepare the session after the connection has been established
-        '''
+        """Prepare the session after the connection has been established.
 
+        Ensure we're in CLI mode.  Disable paging (the '--more--' prompts).  Set
+        the base prompt for interaction ('>').
+        """
         self.cli_mode()
         self.disable_paging(command="set cli screen-length 0\n")
         self.set_base_prompt()
 
 
     def cli_mode(self, delay_factor=1):
-        '''
-        Enter Juniper cli
-        '''
+        """Enter CLI mode."""
         self.clear_buffer()
         self.remote_conn.send("cli\n")
         time.sleep(1*delay_factor)
         self.clear_buffer()
-
         return None
 
 
     def config_mode(self):
-        '''
-        First check whether currently already in configuration mode.
+        """Enter configuration mode.
 
-        Enter config mode (if necessary)
-        '''
-
-        output = self.send_command('\n', strip_prompt=False, strip_command=False)
-        if not '[edit' in output:
-            output += self.send_command('configure\n', strip_prompt=False, strip_command=False)
+        Checks to see if the session is already in configuration mode first.
+        Raises `ValueError` if the session was unable to enter configuration
+        mode.
+        """
+        output = ""
+        configure_marker = "[edit"
+        if not self.check_config_mode(check_string=configure_marker):
+            cmd = "configure\n"
+            output = self.send_command(cmd,
+                                       strip_prompt=False,
+                                       strip_command=False)
             if 'unknown command' in output:
                 raise ValueError("Failed to enter configuration mode")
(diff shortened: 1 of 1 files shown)


supertylerc:
I was unable to run the tests in `tests/test_juniper.*` due to it requiring modules that are not included, such as `DEVICE_CREDS`.

```
 ❯ python tests/test_juniper_srx.py                                                                                                                                                             [16:03:57]
  File "tests/test_juniper_srx.py", line 7, in <module>
    from DEVICE_CREDS import *
ImportError: No module named DEVICE_CREDS
```

supertylerc:
This was done primarily for #55, but it also resolves #62, which was simple.

ktbyers:
Yes, DEVICE_CREDS.py contains actual device credentials so it is intentionally excluded using .gitignore.

You can create it as such with your IP, user/password

juniper_srx = {
    'device_type': 'juniper',
    'ip':   '10.10.10.10',
    'username': 'user',
    'password': 'password'
}

You can then run py.test as follows:

$ py.test -s -v test_juniper_srx.py 
================================ test session starts ================================
platform linux2 -- Python 2.6.9 -- py-1.4.26 -- pytest-2.6.4 -- /usr/bin/python26

test_juniper_srx.py::test_disable_paging PASSED
test_juniper_srx.py::test_verify_ssh_connect PASSED
test_juniper_srx.py::test_verify_send_command PASSED
test_juniper_srx.py::test_base_prompt PASSED
test_juniper_srx.py::test_strip_prompt PASSED
test_juniper_srx.py::test_strip_command PASSED
test_juniper_srx.py::test_normalize_linefeeds PASSED
test_juniper_srx.py::test_clear_buffer PASSED

supertylerc:
In that case, `check_config_mode()` needs to be removed from `SSHConnection()` and placed in `BaseSSHConnection()`.  It's an extremely generic method that accepts an argument for the correct string to check and can thus be used by any vendor without rewriting the method in every vendor's specific class.

If you really wanted end users to be able to just call `device.check_config_mode()`, then you could override the method and return a `super()` call to the parent class.  I don't think it's too much to ask end users to use the method correctly, though, and call the parent class's `check_config_mode()` properly by specifying the expected string (`device.check_config_mode(check_string="]")`, for example).  That would deduplicate code and remove the need for an overridden method in `JuniperSSH`.  It would also remove it from `HPComwareSSH` and `SSHConnection`, two places where it already exists.

supertylerc:
Let me know what you think.  I can refactor that bit of code pretty quickly and easily and submit a PR.  Then nothing breaks, code duplication is reduced, and it's easily extended to other classes (`JuniperSSH`, for example) when it's needed.

supertylerc:
And having `cli_mode()` doesn't make much sense.  It's only used when someone is in the shell, and a person should only be in the shell if they log in as root (or manually start the shell).  If you have to enter the shell in Junos, you're probably doing it wrong or doing manual troubleshooting.

Although the `cli_mode()` method won't return an error, the text return from Junos will be an error.  See below.

```
╭─tyler at deathstar-vi in ~ using ‹ruby-2.1.1› 15-03-08 - 9:47:16
╰─○ ssh fw1
--- JUNOS 12.1X44-D25.5 built 2013-10-24 20:59:21 UTC
No alarms currently active

{primary:node0}
tyler@fw1.example.com> cli
                       ^
unknown command.

{primary:node0}
tyler@fw1.example.com>
```

ktbyers:
Yes, cli_mode() should be removed.

ktbyers:
Merging and I will modify.

ktbyers:
Okay, I just moved check_config_mode() to BaseSSHConnection. I created a Juniper specific version of this that users super().

Issue #68, opened 2015-03-09
Cisco IOS has no exit enable mode (that actually does anything)


ktbyers:

supertylerc:
PR #73 opened for this issue.

supertylerc:
#73 was merged.  @ktbyers, this can probably be closed.

PR #73, opened 2015-03-10
add exit_enable_mode method

supertylerc:
this is for #68.


ktbyers:
FYI, 'exit' from Cisco enable mode will close the SSH session. Will update to use 'disable' command.

PR #77, opened 2015-03-12
Issue with HP Comware Switches

lamiskin:
Start of the prompt in my tested starts with "<".

Included test for this - as it was failing previously.

Also formats were referencing non existing variables.

### DIFF
--- a/netmiko/hp/hp_comware_ssh.py
+++ b/netmiko/hp/hp_comware_ssh.py
@@ -47,7 +47,7 @@ class HPComwareSSH(SSHConnection):
         return super(HPComwareSSH, self).check_config_mode(check_string=check_string)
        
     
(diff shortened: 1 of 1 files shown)


ktbyers:
Can you send me the SSH output of what you are referring to on "<".

lamiskin:
After logging in I get the following prompt before entering any commands:
"<ste1-as01>"

"ste1-as01" is of course the host name.

I am not sure if this is expected or not - I don't often deal with HP switches.

lamiskin:
Forgot to mention the software version number of the HP Switch:
H3C Comware Platform Software
Comware Software, Version 5.20, Release 1211

ktbyers:
I think GitHub markdown might have hid the prompt name above? Can you you re-post and escape the '>' and '<' characters?

lamiskin:
`<ste1-as01>`

ktbyers:
Okay, I see the issue and manually fixed. I fixed it in a slightly different way.

PR #101, opened 2015-07-03
Use sendall() instead of send().

### LINKED ISSUE #100: Send() method used without checking that all data was sent
The [send()](http://paramiko-docs.readthedocs.org/en/latest/api/channel.html#paramiko.channel.Channel.send) method of Paramiko channels doesn't guarantee that all the data gets sent, but there is a [sendall()](http://paramiko-docs.readthedocs.org/en/latest/api/channel.html#paramiko.channel.Channel.sendall) method that does. Using sendall() saves you from having to write your own loop.

This is an easy swap - I'll submit a PR for review.

yourteamneedsadam:
Paramiko's send() method doesn't guarantee that all data will be
sent, but sendall() does.

Fixes #100

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -111,7 +111,7 @@ class BaseSSHConnection(object):
         Disable paging default to a Cisco CLI method
         '''
 
-        self.remote_conn.send(command)
+        self.remote_conn.sendall(command)
         time.sleep(1*delay_factor)
 
         # Clear the buffer on the screen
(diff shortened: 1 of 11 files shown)


yourteamneedsadam:
Build failed but it looks like an unrelated error - let me know if you see anything in this that needs fixing and I'll update it.

ktbyers:
Yes, this should be fine. I looked at Paramiko docs and sendall() looked to be a superior way to do this. Will integrate into code and double check the Travis CI build.

ktbyers:
@yourteamneedsadam FYI, I am planning on converting the license to Apache2 (as it provides a better framework for contributions from a set of contributors). Let me know that you are okay with this.

yourteamneedsadam:
@ktbyers Apache2 license is great - thanks for checking!

Issue #103, opened 2015-07-09
Installing into an empty virtualenv throws an ImportError on paramiko


yourteamneedsadam:
Looks like the setup.py imports netmiko, which fails because the dependencies in install_requires haven't been installed yet. There are some recommended ways around this and I'll submit a PR in a second.

ktbyers:
This should be fixed now.

PR #104, opened 2015-07-09
Fix setup errors

### LINKED ISSUE #91: Unknown distribution option: 'install_requires'
/usr/lib64/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'

yourteamneedsadam:
Fixes #91 and #103.

Uses a PyPa recommended way to fix race conditions in dependency install and uses setuptools instead of distutils (distutils is deprecated). There are a couple ways to do these, just let me know if you think there's a better one and I'll update the PR.

I tested that the install succeeds in empty Python 2 virtualenvs and Python 3 pyvenvs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ tests/etc/responses.yml
 examples/SECRET_DEVICE_CREDS.py
 build
 dist
+*\.egg-info/
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 import os
 import re
 
-from distutils.core import setup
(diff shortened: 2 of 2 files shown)


ktbyers:
@yourteamneedsadam 

I incorporated your find_version fix manually to remove dependency on netmiko package for version info.

I didn't include the other two (egg-info in .gitignore or use of setuptools). I will probably revisit packaging when I try to get Netmiko up on PyPI. I will open a separate issue for setuptools/distutils.

PR #106, opened 2015-07-31
Fortinet device support for netmiko

brutus333:
Hi!
I've added support for fortinet device support. There are some failed tests since Fortinet does not support enable and configure mode. Also, since paging cannot be disabled on VDOM but only system-wide, I had to find a way to get all output from the session. I introduced a little hack in session_preparation method by telling invoke_shell that the terminal has 1000 lines (well, the value is hard-coded which is ugly).
Please merge this work if you find it useful.
Virgil


ktbyers:
Thanks Virgil.

Let me look at it.


Kirk

On Fri, Jul 31, 2015 at 2:57 AM, brutus333 <notifications@github.com> wrote:

> Hi!
> I've added support for fortinet device support. There are some failed
> tests since Fortinet does not support enable and configure mode. Also,
> since paging cannot be disabled on VDOM but only system-wide, I had to find
> a way to get all output from the session. I introduced a little hack in
> session_preparation method by telling invoke_shell that the terminal has
> 1000 lines (well, the value is hard-coded which is ugly).
> Please merge this work if you find it useful.
> Virgil
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/ktbyers/netmiko/pull/106
> Commit Summary
>
>    - Added fortinet vendor
>    - Fixed paging and other bugs
>
> File Changes
>
>    - *A* netmiko/fortinet/__init__.py
>    <https://github.com/ktbyers/netmiko/pull/106/files#diff-0> (1)
>    - *A* netmiko/fortinet/fortinet_ssh.py
>    <https://github.com/ktbyers/netmiko/pull/106/files#diff-1> (76)
>    - *M* netmiko/ssh_dispatcher.py
>    <https://github.com/ktbyers/netmiko/pull/106/files#diff-2> (2)
>    - *M* setup.py
>    <https://github.com/ktbyers/netmiko/pull/106/files#diff-3> (3)
>    - *M* tests/etc/commands.yml.example
>    <https://github.com/ktbyers/netmiko/pull/106/files#diff-4> (11)
>    - *M* tests/etc/responses.yml.example
>    <https://github.com/ktbyers/netmiko/pull/106/files#diff-5> (10)
>
> Patch Links:
>
>    - https://github.com/ktbyers/netmiko/pull/106.patch
>    - https://github.com/ktbyers/netmiko/pull/106.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ktbyers/netmiko/pull/106>.
>



-- 
Kirk Byers
CCIE #6243 emeritus
ktbyers@twb-tech.com
*http://pynet.twb-tech.com <http://pynet.twb-tech.com/blog>*
(415) 264-6270
Simplify your network management through automation

ktbyers:
@brutus333 So you end up seeing no paging because Fortinet thinks it has a large terminal to display to? Have you tested this fix with extended output and does it work properly?

brutus333:
yes, basically this is the idea. Even if it is only an quick and dirty workaround, it works. I tested with extended output.
One sensible setting is the number of lines - I put 1000 thinking that it should be enough for all cases. However, on large installations, the number of line in the output may be larger and in this case it will fail. 
But I don't know what are the consequences of setting a very large number - maybe increased consumption on Fortinet device. I'll try to test it on a virtual instance.

ktbyers:
@brutus333 I haven't forgot about this just haven't got to it. Don't super like the paging solution so I want to dig into it more.

brutus333:
I don't like this solution either but I did not have a better idea. I'll be really happy if you can find something better.

ktbyers:
It looks like Fortinet defaults to paging disabled 'set output standard'. See:

http://docs-legacy.fortinet.com/fweb/5-3-2/cli/index.html#page/FortiWeb%20CLI%20Reference/system_console.html

Is paging commonly enabled/disabled in the Fortinet? Do you have any knowledge/data on this?


Here is what RANCID did (or at least what someone proposed i.e. going into global config if VDOMs are enabled).

http://www.gossamer-threads.com/lists/rancid/users/6729

I propose we do something similar to what RANCID did...i.e. check if VDOMs enabled and then go into global config and disable it). Are there any issues in doing this (for example, does the user that logs into the VDOM have permission to go into global config typically)?

We should also track paging state and restore it to its original state when the script exits.

Thoughts?

brutus333:
Hello!
In our case the paging is enabled - most probably our network admins are using this function.
On the other hand, we are working with the least privileges needed principle, so we have access only on VDOM. I will check with network admins if raising the privilege level is acceptable.
Thank you,
Virgil

brutus333:
Hello!
Raising the privilege level is not an option for us. However, for other sites this may be acceptable.
kind regards,
Virgil

brutus333:
Another idea would be to try the nice approach first (disable paging system wide) and if access is not allowed, revert to the ugly hack with line numbers.
What do you think?

ktbyers:
Yes, let's do this. It will try to disable paging system-wide and give you the option of passing in the terminal parameters in. So by default it will use the system-wide method, but will give people an option for doing the terminal fix.

brutus333:
OK, I will try to implement it in two weeks (now I'm a little busy with other projects) and come back with a PR.

ktbyers:
Okay, sounds good.



Kirk

On Mon, Sep 14, 2015 at 5:01 AM, brutus333 <notifications@github.com> wrote:

> OK, I will try to implement it in two weeks (now I'm a little busy with
> other projects) and come back with a PR.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ktbyers/netmiko/pull/106#issuecomment-140049390>.
>



-- 
Kirk Byers
CCIE #6243 emeritus
ktbyers@twb-tech.com
*http://pynet.twb-tech.com <http://pynet.twb-tech.com/blog>*
(415) 264-6270
Simplify your network management through automation

brutus333:
I started today to work on fortinet implementation. I have a ugly issue anyway: since the terminal hack is done in remote_conn_pre.invoke_shell call, which is called itself by establish_connection() method, I couldn't run any commands before setting the terminal (like doing nice thing first) unless I'm closing and re-opening the connection after trying the nice approach (in case the nice approach fails).
What do you think? Should I implement this way?

brutus333:
I've just commit a version with the discussed functionality. However, I'm not sure how to add terminal options without overriding class __init__ method...
Please look at the code and comment it.
Virgil

ktbyers:
@brutus333 Okay, let me look at it.

ktbyers:
@brutus333 Sorry for the delayed response. I have merged this in. 

I still want to review some of the disable paging items, but I will add an issue to track this.

PR #118, opened 2015-09-29
fix enable() for slow tacacs+ servers

nitzmahone:
Added expect for the password prompt to enable() to allow it to work with slow tacacs+ setups (eg, remote LDAP-integrated). This *should* always work- I can't find any documented examples of enable where it doesn't have a password prompt (you're either in enable immediately at login, or you get a password prompt).

### DIFF
--- a/netmiko/ssh_connection.py
+++ b/netmiko/ssh_connection.py
@@ -17,7 +17,7 @@ class SSHConnection(BaseSSHConnection):
         Enter enable mode
         '''
 
-        output = self.send_command('enable')
+        output = self.send_command_expect('enable', expect_string="password:")
         if 'password' in output.lower():
             output += self.send_command(self.secret)
 


ktbyers:
@nitzmahone Let me think about this a little bit. I am a bit reluctant to use the _expect method, but let me think if there is a better solution (or not).

nitzmahone:
Yeah, I'm not totally sold on it either, but need more reliable way to deal with pokey auth than just a delay. The setup I'm dealing with can take anywhere from 3-20s to come back with the password prompt. :( Just remembered too, that if you blindly do ```enable``` when you're already there, it won't prompt... So I guess it should call check_enable internally first, bleh. I suppose I could pass through the kwarg for expect behavior on the enable method, but that's kinda gross too...

ktbyers:
Yeah, the ways I thought of are:
1. something using _expect (with the caveat you mentioned regarding already being in enable-mode)
2. add a delay_factor argument to the enable method definition which then gets passed to send_command(). Not great because then you have to plan for longest delays.
3. Code a custom enable method that doesn't rely on send_command / send_command_expect (i.e use a while loop and .sendall() / recv() directly).

I am leaning towards option #3.

ktbyers:
Closing as I am going to look into this in the Netmiko 1_0 release. There is a separate issue for it.

PR #119, opened 2015-10-02
Remove last hard-coded reference to flash: filesystem

bernhardschmidt:
This commit removes the last hardcoded reference to the flash: filesystem and uses the self.file_system attribute (which defaults to flash:).

With this change (and basic support for another filesystem in the cisco_file_transfer ansible module which I will push shortly) I was able to manage files on a 7200 VXR with disk2: as file system


bernhardschmidt:
The ansible module PR is at https://github.com/ktbyers/scp_sidecar/pull/10

PR #125, opened 2015-11-04
Add support for Alcatel-Lucent SR-OS based devices

arnaudcoquelet:


ktbyers:
@arnaudcoquelet I apologize about the slow integration of this pull request.

Can you explain to me the fix you put in for the ">" prompt changes that Alcatel has?

Issue #98, opened 2015-05-31
Create a global delay_factor


ktbyers:

jfilliben:
Is there a way to pass a timeout/delay value to ConnectHandler? I am attempting to manage some slow IOS devices on the other side of the globe and I believe the default 0.5 second delay is causing me issues:

Traceback (most recent call last):
  File "dns_crawl.py", line 159, in <module>
    interfacesDict[router] = get_interfaces(routerDict)
  File "dns_crawl.py", line 66, in get_interfaces
    with ssh_manager(routerDict) as netConnect:
  File "C:\Users\308001282\AppData\Local\Continuum\Anaconda2\lib\contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "dns_crawl.py", line 23, in ssh_manager
    conn = SSHClient(**net_device)
  File "build\bdist.win-amd64\egg\netmiko\base_connection.py", line 45, in __init__
  File "build\bdist.win-amd64\egg\netmiko\base_connection.py", line 61, in session_preparation
  File "build\bdist.win-amd64\egg\netmiko\base_connection.py", line 201, in set_base_prompt
ValueError: Router prompt not found: e

ktbyers:
@jfilliben I added a new global_delay_factor argument, you use it when you call the ConnectHandler as follows:

cisco_881 = {
   'device_type': 'cisco_ios',
   'ip': ‘1.1.1.1’,
   'username': ‘admin’,
   'password': ‘password’,
   'global_delay_factor': 3,
}

net_connect = ConnectHandler(**cisco_881)

Netmiko will choose the larger of the two (global_delay_factor or the delay_factor specified in an individual method). By default, they usually are set the same at a value of .5 second.

You will need to use the master branch on GitHub to use the code.

Let me know if this doesn't fix your issue (or isn't working properly).

jfilliben:
Awesome, thanks! I'll test it out and let you know if it helps on Monday.

jfilliben:
Kirk,

I couldn't wait to test :)  I tried it out this morning and it worked perfectly... more slowly, of course, but it worked through all the distant routers without error.

Now I need to work on parallel processing to cut down on the run time.

Thanks!
Jeremy

ktbyers:
Nice :-)

Either using threads or parallel processes should be fine. Python's GIL
(global interpreter lock) won't be an issue here since so much of time is
spent waiting on IO.

Here is an example using Netmiko + Threads:
[...]

PR #127, opened 2015-11-10
Initial delay factor

### LINKED ISSUE #126: delay_factor for connection initialization
Hi Kirk,

We intermittently get this stack trace when connecting to Nexus devices:

```
SSH connection established to X.X.X.X:22
  File "/Users/mzb/.ansible/tmp/ansible-tmp-1447101573.0-35515167451694/ntc_show_command", line 1843, in <module>
    main()
  File "/Users/mzb/.ansible/tmp/ansible-tmp-1447101573.0-35515167451694/ntc_show_command", line 211, in main
    password=password
  File "/Library/Python/2.7/site-packages/netmiko/ssh_dispatcher.py", line 54, in ConnectHandler
    return ConnectionClass(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/netmiko/base_connection.py", line 45, in __init__
    self.session_preparation()
  File "/Library/Python/2.7/site-packages/netmiko/base_connection.py", line 61, in session_preparation
    self.set_base_prompt()
  File "/Library/Python/2.7/site-packages/netmiko/base_connection.py", line 187, in set_base_prompt
    raise ValueError("Router prompt not found: {0}".format(prompt))
ValueError: Router prompt not found:
```

Our belief is that increasing the ``delay_factor`` in ``set_base_prompt()`` would solve this to account for the latency within the switch itself and whatever network latency there may be. Another option would be pushing the ``delay_factor`` param up the stack to ``ConnectHandler()``. If you like either of those options, we'd be happy to submit a PR.

Thanks,
Michael

mzbenami:
Hey Kirk,

This PR resolves #126 and also similar problems I found in ``send_command()`` on higher-delay links with Cisco switches. Namely, ``send_command()`` would return the empty string because there was no data in the buffer yet. The PR includes:

* the addition of a ``wait_for_recv_ready()`` function that loops and sleeps using ``delay_factor`` and ``max_loops`` until data is in the buffer
* the application of ``wait_for_recv_ready()`` in ``set_base_prompt``
* the application of ``wait_for_recv_ready()`` in ``send_command()``

Feel free to use it if it's helpful.

Thanks,
Michael

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -164,7 +164,11 @@ class BaseSSHConnection(object):
 
         self.clear_buffer()
         self.remote_conn.sendall("\n")
-        time.sleep(1 * delay_factor)
+        
+        i = 0
+        while not self.remote_conn.recv_ready() and i < 10:
+            time.sleep(1 * delay_factor)
+            i += 1
 
         prompt = self.remote_conn.recv(MAX_BUFFER).decode('utf-8')
 
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -143,6 +143,20 @@ class BaseSSHConnection(object):
         return output
 
 
+    def wait_for_recv_ready(self, delay_factor, max_loops):
+        '''
+        Wait for data to be in the buffer so it can be received.
+
+        delay_factor can be used to increase the delays.
+
+        max_loops can be used to increase the number of times it reads the data buffer
+        '''
+        i = 0
+        while not self.remote_conn.recv_ready() and i < max_loops:
+            time.sleep(1 * delay_factor)
+            i += 1
+
+    
     def set_base_prompt(self, pri_prompt_terminator='#',
                         alt_prompt_terminator='>', delay_factor=.5):
         '''
@@ -165,10 +179,7 @@ class BaseSSHConnection(object):
         self.clear_buffer()
         self.remote_conn.sendall("\n")
         
-        i = 0
-        while not self.remote_conn.recv_ready() and i < 10:
-            time.sleep(1 * delay_factor)
-            i += 1
+        self.wait_for_recv_ready(delay_factor, 10)
 
         prompt = self.remote_conn.recv(MAX_BUFFER).decode('utf-8')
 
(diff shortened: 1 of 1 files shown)


ktbyers:
@mzbenami I incorporated this change in. I reverted the behavior back in the send_command() section, however. I did this because it changed the behavior of what send_command was doing (and send_command is pretty core to the library). I still like that it is a function because there will be other places where this kind of behavior is probably needed.

Let me know if this change fixes your original issue.

mzbenami:
@ktbyers The original issue was intermittent and believed to be caused by delay. So to test, I artificially increased the bi-directional delay on my NIC, and that did consistently reproduce the issue. The fix in ``set_prompt()`` consistently solved the issue in this artificial environment, so I assume it will also solve the original intermittent issue. However, in this artificial environment, ``send_command()`` would consistently return the empty string. The fix I included in ``send_command()`` solved that problem in the artificial environment, but I'm not sure that problem would even come up under normal circumstances. So the way it is now is great, and I'll let you know if I ever I see ``send_command()`` return the empty string when it shouldn't under normal network circumstances.

ktbyers:
@mzbenami Yes, just let me know if you see issues with send_command. That code is pretty core so I am more reluctant to change it (due to potential side effects).

ktbyers:
@mzbenami FYI, I do sometimes see intermittent failures in my test environment. I am going to open an issue in GitHub regarding this. When I ran the unit tests, I got a failure on Juniper (I think with commit). I am running that test a bunch of times in succession to see if I can reproduce it.

mzbenami:
@ktbyers Sounds good. On another note, are changes to the GitHub master branch automatically synced to pypi, or should we use ``git clone`` to install these changes?

ktbyers:
No, they won't be up on pypi until I do my next release. Until then you will need to sync with master.

ktbyers:
@mzbenami FYI, I also just added a new global_delay_factor to the master branch see issue #98.

mzbenami:
@ktbyers Awesome, that should definitely help if we see that issue again. Thanks!

Issue #131, opened 2015-11-17
ProxyCommand Support


coxley:
Should be a no-brainer since Paramiko supports this directly.

coxley:
There are a couple options:

1. Have user provide raw ProxyCommand string via `proxy` kwarg
2. Have user provide path to ssh config (`/etc/ssh/ssh_config` or `~/.ssh/config`) and conform to connection settings they have configured.

I vote for no. 2 because it allows much more flexibility. With Paramiko you can load an openssh config and do `conf.lookup(hostname)` to get all the settings that would be required for that host. Useful because most users of ProxyCommand have it set up with wildcards, eg `*prod.example.com`.

Result would be smoother experience for user

ktbyers:
I will have to look into this a bit. 

If you have a proposed solution, definitely free to submit a PR.

coxley:
@ktbyers Coming up shortly, will like to get your feedback. :)

ktbyers:
Sounds good.

ktbyers:
This is done.

PR #133, opened 2015-11-18
OpenSSH Configuration Support (#131)

coxley:
This has been implemented by outsourcing paramiko.SSHClient.connect(**kwargs)
to a private method ._return_config()

Not only will this give a more central place to tweak client configuration in
the future, but it serves a nice place to do some conditions and openssh cfg
file reading thanks to paramiko.SSHConfig

As of this commit, ssh_config_path is an __init__ arg defaulting to ~/.ssh/config.
If an unparsable ssh config is passed, Exception will be thrown. We can easily
capture this and error something better if needed, but it's a straightfwd error.

On top of proxycommand support, there is now support for Host Aliasing within
a user SSH config. The only drawback/discussion might need of this is whether
to log the provided hostname or the resolved one (and port).

Username will also be favored from the config if provided under the Host block.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ examples/SECRET_DEVICE_CREDS.py
 build
 dist
 netmiko.egg-info
+.cache
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -30,7 +30,7 @@ class BaseSSHConnection(object):
 
     def __init__(self, ip, username, password, secret='', port=22, device_type='',
                  verbose=True, use_keys=False, key_file=None, global_delay_factor=.5,
-                 ssh_config_path='~/.ssh/config'):
+                 ssh_config_path=None):
 
         self.ip = ip
         self.port = port
@@ -69,22 +69,30 @@ class BaseSSHConnection(object):
     def _return_config(self, use_keys=False, key_file=None, timeout=8):
         '''Return dict of kwargs that Paramiko will accept for connect'''
 
-        # Fully expand ssh_config_path and test its existence
-        # If it exists, use Paramiko SSHConfig to generate source content.
-        # Otherwise use empty dict
-        full_path = os.path.abspath(os.path.expanduser(self.ssh_config_path))
-        if os.path.exists(full_path):
-            ssh_config_instance = paramiko.SSHConfig()
-            with open(full_path) as f:
-                ssh_config_instance.parse(f)
-                source = ssh_config_instance.lookup(self.ip)
+        # If ssh_config_path is given, go through the following process to
+        # validate and pull selected values if they exist. Otherwise set source
+        # to empty dict
+        #
+        # This lets the return code be the same by utilizing source.get
+        if self.ssh_config_path:
+            # Fully expand ssh_config_path and test its existence
+            # If it exists, use Paramiko SSHConfig to generate source content.
+            # Otherwise use empty dict
+            full_path = os.path.abspath(os.path.expanduser(self.ssh_config_path))
+            if os.path.exists(full_path):
+                ssh_config_instance = paramiko.SSHConfig()
+                with open(full_path) as f:
+                    ssh_config_instance.parse(f)
+                    source = ssh_config_instance.lookup(self.ip)
+            else:
+                source = {}
+
+            if source.get('proxycommand'):
+                proxy = paramiko.ProxyCommand(source['proxycommand'])
+            else:
+                proxy = None
         else:
             source = {}
-
-        if source.get('proxycommand'):
-            proxy = paramiko.ProxyCommand(source['proxycommand'])
-        else:
-            None
 
         # Contents of source aren't reliable, except for hostname and even
         # then we're setting source to empty dict sometimes. When these fields
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -15,6 +15,7 @@ import time
 import socket
 import re
 import io
+import os
 
 from netmiko.netmiko_globals import MAX_BUFFER
 from netmiko.ssh_exception import NetMikoTimeoutException, NetMikoAuthenticationException
@@ -27,8 +28,9 @@ class BaseSSHConnection(object):
     Otherwise method left as a stub method.
     '''
 
-    def __init__(self, ip, username, password, secret='', port=22, device_type='', verbose=True,
-                 use_keys=False, key_file=None, global_delay_factor=.5):
+    def __init__(self, ip, username, password, secret='', port=22, device_type='',
+                 verbose=True, use_keys=False, key_file=None, global_delay_factor=.5,
+                 ssh_config_path='~/.ssh/config'):
 
         self.ip = ip
         self.port = port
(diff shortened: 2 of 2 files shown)


coxley:
Linking in title didn't work. This corresponds to #131 

Even though it says tests fail, that appears to be an issue with travis. Also I noticed tests are broken in the project because they can't find cisco_881 key in device yml.

ktbyers:
@coxley The test failing is fine. I think the test issue pertains to I have an encrypted file to protect passwords and when other people do pull requests Travis doesn't know how to decrypt the file.

Not sure if there is a workaround for this (while still keeping passwords secure).

Anyways when I merge the pull request, it will test normally as Travis builds pushed by me can decrypt the encrypted file (containing the passwords).

ktbyers:
@coxley Sorry I have got back to you on this and the other async item you posted on. Will try to get back to you probably early next week.

coxley:
@ktbyers All is good, enjoy your holidays and we can talk soon.

ktbyers:
@coxley So what are we trying to accomplish here?

Are we trying to:

1. Add SSH Proxy support using a 'config' file.
2. Add support of pulling router/switch parameters from ssh 'config' file
3. Both of the the above.

I am skeptical that item 2 will be used very much (it strikes me as a very rare use case and not very useful). You could already store your router parameters in a DB, YAML file, JSON file--why would you want to store them here?

Adding support for SSH Proxy sounds valuable to me, however.

Let me know your thoughts on this. I am happy to hear an alternate opinion.

Kirk

coxley:
@ktbyers, This implements basic OpenSSH configuration support for:

* ProxyCommand
* Hostname
* Port

Those are useful so if you have an .ssh/config like:

```
Host routerA
  Hostname real.hostname.here

Host *.prod.example.com
  Port 222

Host *.reallyprod.example.com
  ProxyCommand [...]
```

It'll do the right thing as a user would expect. Certainly not for storing device parameters, just SSH parameters - the former would be whacky! :)

ktbyers:
Okay, I am thinking we should have the following in __init__.py

ssh_config_path=None

Then do a later check in establish_connection along the lines of:

if ssh_config_path:
      # Grab params from the specified config file
else:
      Do normal behavior.


Basically, I want the normal behavior to be unchanged, but then if someone
wants to use ssh_config_path they could just specify the 'config' file in
the Connection objection creation process.


Kirk


On Wed, Dec 2, 2015 at 8:37 PM, Codey Oxley <notifications@github.com>
wrote:

> @ktbyers <https://github.com/ktbyers>, This implements basic OpenSSH
> configuration support for:
>
>    - ProxyCommand
>    - Hostname
>    - Port
>
> Those are useful so if you have an .ssh/config like:
>
> Host routerA
>   Hostname real.hostname.here
>
> Host *.prod.example.com
>   Port 222
>
> Host *.reallyprod.example.com
>   ProxyCommand [...]
>
> It'll do the right thing as a user would expect. Certainly not for storing
> device parameters, just SSH parameters - the former would be whacky! :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ktbyers/netmiko/pull/133#issuecomment-161512257>.
>



-- 
Kirk Byers
CCIE #6243 emeritus
ktbyers@twb-tech.com
*http://pynet.twb-tech.com <http://pynet.twb-tech.com/blog>*
(415) 264-6270
Simplify your network management through automation

coxley:
Sounds reasonable to me! I've updated my branch:

```diff
$ git diff 99ccad7d^ 99ccad7d 

diff --git a/netmiko/base_connection.py b/netmiko/base_connection.py
index e603534..ca6529b 100644
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -30,7 +30,7 @@ class BaseSSHConnection(object):
 
     def __init__(self, ip, username, password, secret='', port=22, device_type='',
                  verbose=True, use_keys=False, key_file=None, global_delay_factor=.5,
-                 ssh_config_path='~/.ssh/config'):
+                 ssh_config_path=None):
 
         self.ip = ip
         self.port = port
@@ -69,22 +69,30 @@ class BaseSSHConnection(object):
     def _return_config(self, use_keys=False, key_file=None, timeout=8):
         '''Return dict of kwargs that Paramiko will accept for connect'''
 
-        # Fully expand ssh_config_path and test its existence
-        # If it exists, use Paramiko SSHConfig to generate source content.
-        # Otherwise use empty dict
-        full_path = os.path.abspath(os.path.expanduser(self.ssh_config_path))
-        if os.path.exists(full_path):
-            ssh_config_instance = paramiko.SSHConfig()
-            with open(full_path) as f:
-                ssh_config_instance.parse(f)
-                source = ssh_config_instance.lookup(self.ip)
-        else:
-            source = {}
+        # If ssh_config_path is given, go through the following process to
+        # validate and pull selected values if they exist. Otherwise set source
+        # to empty dict
+        #
+        # This lets the return code be the same by utilizing source.get
+        if self.ssh_config_path:
+            # Fully expand ssh_config_path and test its existence
+            # If it exists, use Paramiko SSHConfig to generate source content.
+            # Otherwise use empty dict
+            full_path = os.path.abspath(os.path.expanduser(self.ssh_config_path))
+            if os.path.exists(full_path):
+                ssh_config_instance = paramiko.SSHConfig()
+                with open(full_path) as f:
+                    ssh_config_instance.parse(f)
+                    source = ssh_config_instance.lookup(self.ip)
+            else:
+                source = {}
 
-        if source.get('proxycommand'):
-            proxy = paramiko.ProxyCommand(source['proxycommand'])
+            if source.get('proxycommand'):
+                proxy = paramiko.ProxyCommand(source['proxycommand'])
+            else:
+                proxy = None
         else:
-            None
+            source = {}
 
         # Contents of source aren't reliable, except for hostname and even
         # then we're setting source to empty dict sometimes. When these fields

```

ktbyers:
@coxley Things to override from SSH 'config' file.

proxy --> Yes

hostname --> has to override 'ip' argument already provided. We need to do this to support the wildcard in the SSH 'config file'

port --> defaults to 22. So override provided that the object creation used the default port 22. If object creation used a different port, then object 'port' attribute should take precedence.

username --> Probably should change __init__ to username='' by default. Then SSH config username will be used if username == '' else object username will be used.

I am working on integrating what you did and on the above.

Thoughts?

ktbyers:
@coxley I pushed up a possible solution to this. Please check as I changed things meaningfully. I haven't tested it against an SSH 'config' file.

I am regression testing it now.

ktbyers:
Added this issue #145 to track what you had put in a comment (i.e. that some of this behavior will not be intuitive and needs documented for anyone that wants to use an SSH config file).

coxley:
@coxley, I think this is fine - installed from your branch and confirmed (Seems Force10 works fine with IOS driver... only on S4810s but not S55 switches :\) I think that mutating a dict by passing it through a function vs. new dict via return value is odd, but that's just a personal gripe with Python. :)

Good call on the separate issue for documentation, only right for me to tackle that one.

PR #135, opened 2015-11-18
Add ovs_linux device_type

jumpojoy:
ovs_linux - allows to interact with ovs switch on linux
 via bash commands

### DIFF
--- a/netmiko/ovs/__init__.py
+++ b/netmiko/ovs/__init__.py
@@ -1,0 +1,1 @@
+from netmiko.ovs.ovs_linux_ssh import OvsLinuxSSH
--- a/netmiko/ovs/ovs_linux_ssh.py
+++ b/netmiko/ovs/ovs_linux_ssh.py
@@ -1,0 +1,32 @@
+from netmiko.ssh_connection import SSHConnection
+
+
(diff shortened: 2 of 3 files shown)


ktbyers:
@jumpojoy 

Sounds good. 

I will try to integrate today/tomorrow.

jtdub:
Am I missing something? The code looks to be just a generic Linux SSH connector. I didn't see anything that was ovs specific.

ktbyers:
@jtdub Yeah, you are right. It is just a generic Linux connector. My error on posting to Twitter earlier (I misinterpreted part of it). I should probably move it to a generic Linux class and then inherit OVS from that.

jumpojoy:
@jtdub: managing openvswitch on linux is performed via bash commands ('ovs-vsctl' in general but other also can be used [0] ). @ktbyers: Should I move all code to generic linux class and inherit ovs from it? 

[0] http://openvswitch.org/support/dist-docs-2.4/

ktbyers:
@jumpojoy Yes, I think we should do that (i.e. have a generic Linux class and then inherit OVS from this). There will probably be other use cases for just Linux and that will let us decouple Linux from OVS.

PR #139, opened 2015-11-26
Add linux_ssh class

jumpojoy:
This change introduce new class LinuxSSH which is
generic linux ssh connector.
update OvsLinuxSSH class, inherit it from LinuxSSH


ktbyers:
@jumpojoy I will probably add the password default so that SSH keys can directly be used. Just need to check no side-effects (which there probably aren't).

PR #140, opened 2015-12-01
Added support for enterasys and extreme switches.

nitrocode:
I read through the VENDORS.md file but I noticed when testing the extreme switches that sometimes my output would be 

    '\n\r'

or something similar to that but when I ran the script again or ran the test again, I would get the output I desired. I also checked the prompt and netmiko was discovering that correctly too.

Increasing the global delay factor in the config json from 0.5 to 1 seemed to do the trick. 

    extremehandle = {'device_type':'extreme', 'ip' : '10.54.149.80', 'username':'admin', 'password':'',
		 'global_delay_factor': 1}

Is there a way to better debug than by increasing the global delay factor?

My TODO is to write a few tests for these switches too. I'll try to add that in the next commit.

### DIFF
--- a/netmiko/enterasys/__init__.py
+++ b/netmiko/enterasys/__init__.py
@@ -1,0 +1,1 @@
+from netmiko.enterasys.ent_ssh import ENTSSH
--- a/netmiko/enterasys/ent_ssh.py
+++ b/netmiko/enterasys/ent_ssh.py
@@ -1,0 +1,17 @@
+'''
+ENT support
+'''
+from netmiko.ssh_connection import SSHConnection
+
+
+class ENTSSH(SSHConnection):
+    '''
+    ENT support
+    '''
+    def session_preparation(self):
+        '''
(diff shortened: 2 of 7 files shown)


ktbyers:
@nitrocode On the debugging question, I usually look at the nature of the exception and you can print output of the current context to see what happened when you failed.

ktbyers:
@nitrocode Note, I changed some of the naming of this to make it more consistent with existing naming standard and with Python standard. Also I modified slightly your test files so please make sure they still run properly.

nitrocode:
@ktbyers Awesome, thank you. 

Just wondering, how come I didn't get added to the list of contributers? Issue with my git?

ktbyers:
@nitrocode Strange, I don't understand that. 

The thing that is a bit odd is that the commits come from 'rbhatia' and I see this in the list of commits, but then the pull request came from you.

Maybe an issue with your git user.name/git user.email for when you submitted the commits?

PR #142, opened 2015-12-02
Disable enable_mode for linux

jumpojoy:
We can't mirror enable and config mode.
It is impossible to exit from config_mode in this case
since both enable and config modes are using the same
check_string='#'


ktbyers:
@jumpojoy I did run into an issue which was if you did enable() and then config_mode(), then 'sudo su' would be executed multiple times.

If you then did 'exit' you would still be at the '#' prompt.

I put in a fix for this here 838297f5c3dbd7e849ecd36d08f37d8198cadab1

I also added a quick set of test code in tests/test_linux.py. I need to convert this to pytest but it works for a quick test.

PR #148, opened 2015-12-14
Update cisco_asa enable method

lamiskin:
Raise exception if enable password is not accepted. 
Throw an longer warning if enable secret is not set (personally I normally expect the login password to be tried, but instead I thought it would be best to give a longer warning).

I had issues where if the secret was not set, it would throw an issue about not setting base prompt, which wasn't very intuitive.


ktbyers:
@lamiskin Merged in...I removed the check for enable secret and separate exception, however (seemed unnecessary).

I have a separate issue related to implementing a login() method for the ASA (which we could potentially integrate to enable() method). In other words, check if enable secret is blank and if it is attempt to do login() method.

lamiskin:
Sounds good to me :+1:

PR #171, opened 2016-02-02
added _ssh to existing device_type names

jedelman8:
I saw you had `_ssh` added to `cisco_wlc` supporting both `cisco_wlc` and `cisco_wlc_ssh` as device types.

Would you mind supporting this for all device types? For the ntc work, we are trying to stick to vendor_os_api, so this would be great to have natively in netmiko too, so it's consistent.  If not, I understand, and we can add this directly to ntc modules.

Let me know.


ktbyers:
@jedelman8 I did it a slightly different way as I don't want to break existing code.

So it supports both the old_name and old_name_ssh. 

This was pushed in this change https://github.com/ktbyers/netmiko/commit/d035f6d7116dfb794fd2f5c8009f9c3e7c644f76

ktbyers:
Let me know if that doesn't work for some reason.

jedelman8:
Sounds good, should be fine.  Thanks.

Issue #164, opened 2016-01-19
find_prompt fails if Cisco Nexus devices have enabled "terminal color persist"


smnmtzgr:
Hi Kirk, at first thanks for your great work!
I've found an issue that happens when the user which is used for netmiko connects to a Cisco NX-OS device and has enabled "terminal color persist" on this device. 
The prompt of the device cant be found correctly in this situation because it doesnt end with the primary or secondary character (# or >). The colored prompt is formatted as described here: https://en.wikipedia.org/wiki/ANSI_escape_code

Logically your function doesnt find the prompt. I think there are 2 solutions for this:
1) Dont use colors for the user running ansible ("no terminal color persist"). :)
2) Change the function find_prompt to work with colored prompts (I think this would be too much overhead).

At last I want to give you an example (VSM-LAN-K has enabled "terminal color persist" for my user):

python (interactive shell):
<pre>
>>> from netmiko import ConnectHandler
>>> net_connect = ConnectHandler(device_type='cisco_nxos', ip='10.10.10.1', username='admin', password='password')
SSH connection established to 10.10.10.1:22
Interactive SSH session established
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/netmiko/ssh_dispatcher.py", line 65, in ConnectHandler
    return ConnectionClass(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/netmiko/base_connection.py", line 63, in __init__
    self.session_preparation()
  File "/usr/lib/python2.7/site-packages/netmiko/base_connection.py", line 77, in session_preparation
    self.set_base_prompt()
  File "/usr/lib/python2.7/site-packages/netmiko/base_connection.py", line 281, in set_base_prompt
    raise ValueError("Router prompt not found: {0}".format(prompt))
ValueError: Router prompt not found: VSM-LAN-K#
</pre>

ktbyers:
@smnmtzgr And the problem goes away when 'terminal color persist' is disable?

smnmtzgr:
Yes it goes away. Thats definitively the problem. My co-worker made a fix to it: https://github.com/Benedikt1992/netmiko/commit/71e50105569977bbeb62c280ae00030264900a16

The "terminal color persist" is a per user-based command. If you enter "no terminal color persist" (thats default) there is no colored prompt in NX-OS and all is fine.

ktbyers:
[...]

Issue #168, opened 2016-01-23
[NX-OS] terminal color persist


NetSecParano:
Hello,
Terminal prompt cannot be found if the user activated "terminal color persist" on a Nexus. 
On the function set_base_prompt, when prompt[-1] is checked, it fails because prompt[-1] return a char used to display the color. 

You could maybe sanitaze and withdraw those additionnal chars before checking the prompt.
Or, easier, just send "no terminal color" before (maybe at the same time as the "terminal lenght 0").

Anyways, thank you for the great work :+1

ktbyers:
Fix here https://github.com/ktbyers/netmiko/commit/dcf323c9d9e1c9b138e1ba654b4d09ac2c832de8

PR #178, opened 2016-02-21
Regex for stripping most ansi escape codes

Benedikt1992:
As @smnmtzgr already metioned in Issue #164 the function find_prompt() can't find the prompt if there are colors enabled. Technically speaking the function strip_ansi_escape_codes() doesn't even get called and can't remove all escape codes.

Therefore I enhanced the function to remove the most of possible escape codes (maybe there are some special codes that are not matched) and introduced a new argument to enable the stripping of escape codes. Default is not to strip escape codes (as it is today).

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -30,7 +30,7 @@ class BaseSSHConnection(object):
     def __init__(self, ip, username, password='', secret='', port=22, device_type='', verbose=True,
                  global_delay_factor=.5, use_keys=False, key_file=None, ssh_strict=False,
                  system_host_keys=False, alt_host_keys=False, alt_key_file='',
-                 ssh_config_file=None):
+                 ssh_config_file=None, ansi_escape_codes=False):
 
         self.ip = ip
         self.port = int(port)
@@ -38,7 +38,7 @@ class BaseSSHConnection(object):
         self.password = password
         self.secret = secret
         self.device_type = device_type
-        self.ansi_escape_codes = False
+        self.ansi_escape_codes = ansi_escape_codes
 
         # Use the greater of global_delay_factor or delay_factor local to method
         self.global_delay_factor = global_delay_factor
@@ -267,7 +267,6 @@ class BaseSSHConnection(object):
         # Some platforms have ANSI escape codes
         if self.ansi_escape_codes:
             prompt = self.strip_ansi_escape_codes(prompt)
-
         prompt = self.normalize_linefeeds(prompt)
 
         try:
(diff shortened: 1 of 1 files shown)


ktbyers:
@Benedikt1992 Yes, historically strip_ansi_escape_codes wasn't for Cisco (only cases where needed mainly HP ProCurve).

The NEXUS color issue is a fairly new issue.

Is this NEXUS color enabled by default? I seem to recall no (could be wrong on this)? Also why not just turn it off (at least for scripts)? Do many people use this?

Reluctant to change this without a better reason as stripping data from output is not that great of a practice. 

Also probably wouldn't accept this change anyways (at a minimum you would have to document the regex). 

I think the previous way is better i.e. explicitly identify the ANSI escape codes to strip. So if we want to fix the NEXUS color issue then would have to look at the ANSI codes it generates. 

If this is impractical due too many ANSI escape codes or there isn't some simpler pattern that we could match on, then we could possibly go to a more complex regex pattern. I am pretty strongly against complex regex patterns, however, as they are really hard to understand, maintain, and test.

Benedikt1992:
@ktbyers As @smnmtzgr wrote this color feature is disabled by default. If you want to use this feature you can enable it per user. So if you want to disable it for scripts you have 3 options
1. You run the script with an different user
2. You don't use the feature
3. Each script checks if there is the color feature enabled. If appplicable the script turns off the feature in the beginning and reenables it when finished.

The last two option feels a little bit dirty to me. The first option may works in many cases. Nevertheless filtering those problematic escape codes seems the clearst solution to me. Basically every command line tool has to filter those escape codes. Since every command line tool strips those escape codes it's not necessarily trivial to determine which Codes are used. Filtering every escape code seems to be much easier.

The regex isn't easy at all. How would you imagine such a documentation?

ktbyers:
@Benedikt1992 @smnmtzgr 1 or 2 sound fine to me (especially since this feature is not enabled by default). 

> Basically every command line tool has to filter those escape codes. Since every command line tool strips those escape codes it's not necessarily trivial to determine which Codes are used.

We only need to solve the NEXUS color escape codes, not every possible ANSI escape code. This is probably a pretty small set of escape codes. It is pretty rare for me to run into ANSI escape codes in router output.

The flip side is filtering output from the output stream is not a great idea (in general). I also don't know all of the uses for ANSI escape codes so I am reluctant to have a broad filter (especially when it rarely applies to router output).

Can someone post the output that you get back from the Nexus with the ANSI escape codes embedded in it?

ktbyers:
I think we are at an impasse on this. 

I am open to other solutions on this, my preferred solution is just to add the ANSI color codes that that the Nexus uses in its output and strip them specifically. I don't have access to a Nexus so I can't test right now.

I have been working on getting a virtual Nexus working on KVM, but am still running into issues. I will probably try to fix this once I get that working.

In the interim, I will track this as an issue in GitHub.

ktbyers:
Tracking this issue here https://github.com/ktbyers/netmiko/issues/168

Issue #181, opened 2016-02-25
SCP: Validate md5 without comparing to a local file


ogenstad:
It should be possible to retrieve the md5 checksum of a file transfered through scp. Currently you can only compare to the local file (which might be corrupted).

ktbyers:
So you are looking for a method that returns the remote file's MD5?

ogenstad:
Yes something in the lines of

```Python
    def show_remote_md5(self, base_cmd='verify /md5', delay_factor=8):
        '''
        Calculate remote MD5 and compare to source MD5
        Default command is Cisco specific
        This command can be CPU intensive on the remote device
        Return boolean
        '''
        remote_md5_cmd = "{0} {1}{2}".format(base_cmd, self.file_system, self.dest_file)
        dest_md5 = self.ssh_ctl_chan.send_command_expect(remote_md5_cmd)
        dest_md5 = self.process_md5(dest_md5)
        return dest_md5
```

ktbyers:
@ogenstad Probably should just submit this as a PR. I will get to it faster that way.

PR #184, opened 2016-03-11
Added function to show remote md5 #181

ogenstad:
PR to show remote md5 as per https://github.com/ktbyers/netmiko/issues/181

Perhaps this function should be used in compare_md5 so that the functionality isn't duplicated.


ktbyers:
@ogenstad I made some minor modifications to this commit. I did rename the method, however, to just remote_md5.

PR #185, opened 2016-03-21
Use a command after connection instead of invoking a shell

erdnaxeli:
Hi,

I have a special use case. For security reason, I have to jump on a first host before connecting to the real destination. The connections looks like this:

Me --SSH--> Secure_Host --SSH--> final destination

I connect with SSH to Secure_Host, executing a special command (not a shell) that starts a new SSH connection between Secure_Host and my final destination. This looks like  "ssh me@Secure_Host -- other_user@final_destination".

This means that I can't use any solution based on ProxyCommand (like [this one](http://www.cyberciti.biz/faq/linux-unix-ssh-proxycommand-passing-through-one-host-gateway-server/)). ProxyCommand is useful when you want to make a tunnel through a jump host, but with a SSH connection between you and the destination.

In my case I need to run a command on the jump host to establish a second SSH session between it and the final destination. With paramiko, I need to use _exec_command()_ instead of _invoke_shell()_.

This patch basically allows to run any command right after the connection, by adding two new parameters to _BaseSSHConnection_'s constructor: _cmd_ and _cmd_sleep_:

- The _cmd_ while be launch after the connection, on his own channel, and this channel will be used after instead of a shell channel.

- The _cmd_sleep_ is the duration in second to sleep after running the command. If the command returns some text too slowly before returning the actual destination's prompt _find_prompt()_ fails because it gets a fragment of the command output without the prompt it is looking for.

With my modification I can use netmiko with these parameters:

```python
args = {
    'cmd': 'other_user@final_destination,
    'cmd_sleep': 1,
    'ip': 'Secure_Host',
    'device_type': 'cisco_ios',
    'username': 'me',
    'use_keys': True,
    'verbose': True,
}
```

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -30,7 +30,7 @@ class BaseSSHConnection(object):
     def __init__(self, ip=u'', host=u'', username=u'', password=u'', secret=u'', port=22,
                  device_type=u'', verbose=False, global_delay_factor=.1, use_keys=False,
                  key_file=None, ssh_strict=False, system_host_keys=False, alt_host_keys=False,
-                 alt_key_file='', ssh_config_file=None):
+                 alt_key_file='', ssh_config_file=None, cmd=u'', cmd_sleep=.1):
 
         if ip:
             self.host = ip
@@ -45,6 +45,8 @@ class BaseSSHConnection(object):
         self.secret = secret
         self.device_type = device_type
         self.ansi_escape_codes = False
+        self.cmd=u''
+        self.cmd_sleep = int(cmd_sleep)
 
         # Use the greater of global_delay_factor or delay_factor local to method
         self.global_delay_factor = global_delay_factor
@@ -176,8 +178,14 @@ class BaseSSHConnection(object):
         if verbose:
             print("SSH connection established to {0}:{1}".format(self.host, self.port))
 
-        # Use invoke_shell to establish an 'interactive session'
-        self.remote_conn = self.remote_conn_pre.invoke_shell()
+        if self.cmd:
+            self.remote_conn = self.remote_conn_pre.get_transport().open_channel("session")
+            self.remote_conn.exec_command(self.cmd)
+            time.sleep(self.cmd_sleep)
+        else
+            # Use invoke_shell to establish an 'interactive session'
+            self.remote_conn = self.remote_conn_pre.invoke_shell()
+
         self.special_login_handler()
         if verbose:
             print("Interactive SSH session established")


ktbyers:
@erdnaxeli Why can't you use SSH proxy and then run a different command on the intermediate server? SSH proxy doesn't limit you to just 'nc [host] [port]', you should be able to run other commands.

See https://pynet.twb-tech.com/blog/automation/netmiko-proxy.html

erdnaxeli:
SSH proxy does not work in the same way. SSH runs the ProxyCommand, then tries to open an SSH connection using the proxy's STDIN and STDOUT like a socket. With "ssh me@jumphost nc destination 22" we have the following:

```
      +----+    +--------+    +-----------+
      | me |    |jumphost|    |destination|
      |    |    |        |    |           |
Proxy |SSH +--> |  nc    +--> | SSHD      |
      |    |    |        |    |   ^       |
      |SSH +------------------|---+       |
      |    |    |        |    |           |
      +----+    +--------+    +-----------+
```

This works because nc opens the socket and wait for data to write in it.

What I want is:

```
      +----+    +--------+    +-----------+
      | me |    |jumphost|    |destination|
      |    |    |        |    |           |
      |SSH +--> |  black +--> | SSHD      |
      |    |    |  box   |    |    +      |
      |    |    |        |    |    v      |
      |    |    |        |    | cisco cli |
      +----+    +--------+    +-----------+

```
I don't want a direct SSH connection between me and my destination, I don't even have the credentials to do it.

If I set a ProxyCommand like "ssh me@jumphost -- other_user@destination", what is done is:

```
      +----+    +--------+    +-----------+
      | me |    |jumphost|    |destination|
      |    |    |        |    |           |
Proxy |SSH +--> |  black +--> | SSHD      |
      |    |    |  box   |    |    +      |
      |    |    |        |    |    v      |
      |SSH +----------------> | cisco cli |
      +----+    +--------+    +-----------+
```

Here my jumphost does not just open the socket, it replies some garbage before openning the SSH connection to the destination, and then open it. So my initial SSH process on my host cannot make an SSH connection to destination. It is not talking to a SSH daemon, it is talking to a cisco cli.

Actually I don't event know what runs on the jumphost (and I can't open a shell on it). When I run "ssh me@Secure_Host -- other_user@final_destination", "other_user@final_destination" is passed as a parameter to the black box (probably a "command" parameter in the authorized_keys file).

This is obvious when enabling verbose mode:

```
me@src_host:~$ cat .ssh/config 
Host thisdoesnotwork
    IdentityFile /home/me/.ssh/id_rsa
    ProxyCommand /usr/bin/ssh me@jumphost -- other_user@final_destination
me@src_host:~$ ssh -v thisdoesnotwork
OpenSSH_6.7p1 Debian-5, OpenSSL 1.0.1k 8 Jan 2015
debug1: Reading configuration data /home/me/.ssh/config
…
debug1: Executing proxy command: exec /usr/bin/ssh me@jumphost -- other_user@final_destination
…
debug1: Local version string SSH-2.0-OpenSSH_6.7p1 Debian-5
debug1: ssh_exchange_identification: src_host:47624 => me@jumphost:22 => other_user@final_destination:22 ...
debug1: ssh_exchange_identification: 
debug1: ssh_exchange_identification: 
debug1: ssh_exchange_identification: \033[0m allowed ... log on(/home/me/ttyrec/final_destination/2016-03-21.17-52-57.12869)
debug1: ssh_exchange_identification: 
debug1: ssh_exchange_identification: \033[0mSSH-2.0-OpenSSH_6.7p1 Debian-5
debug1: ssh_exchange_identification: 
debug1: ssh_exchange_identification: Failed to add the host to the list of known hosts (/home/me/.ssh/known_hosts).
debug1: ssh_exchange_identification: final_destination#SSH-2.0-OpenSSH_6.7p1 Debian-5
debug1: ssh_exchange_identification:             ^
debug1: ssh_exchange_identification: % Invalid input detected at '^' marker.
debug1: ssh_exchange_identification: 
debug1: ssh_exchange_identification: final_destination#
```

On one side we have an SSH process tring to do the ssh_exchange_identification, on the other we have the script on the jumphost talking about logs and then the final destination's cli, wich does not understand what SSH is trying to tell him ("% Invalid input detected at '^' marker.").

ktbyers:
@erdnaxeli I probably disagree with you on how SSH proxy works (though definitely could be wrong), but would need more time to experiment. Regardless, it looks like given your restricted setup you probably can't make SSH proxy work.

Unfortunately, your situation is too rare to make a general change (albeit a very minor change). Few people even have SSH proxies and then an ultra-restricted one like yours.

Consequently, I am going to reject this change.

I would recommend you just make a child class and overwrite the \_\_init\_\_ method. You can then sub-in your code and still use the other Netmiko methods.

erdnaxeli:
Ok I understand. I wil do some monkey patching for my specific case :)

PR #191, opened 2016-04-15
Set base_prompt for linux to '@'

jumpojoy:
On linux host base prompt depends on username and may be customized.
By default it is set to the following:
for user: user@hostname:~$
for root: root@hostname:#

This patch sets base prompt for linux to '@'.
It fixes issue when it is not possible to get to configuration mode from general user.
Since after 'sudo su' command, output prompt from host is changed to root@hostname...
While base_prompt left user@hostname...

### DIFF
--- a/netmiko/linux/linux_ssh.py
+++ b/netmiko/linux/linux_ssh.py
@@ -7,10 +7,7 @@ class LinuxSSH(SSHConnection):
     def set_base_prompt(self, pri_prompt_terminator='$',
                         alt_prompt_terminator='#', delay_factor=.1):
         """Determine base prompt."""
-        return super(SSHConnection, self).set_base_prompt(
-            pri_prompt_terminator=pri_prompt_terminator,
-            alt_prompt_terminator=alt_prompt_terminator,
-            delay_factor=delay_factor)
+        self.base_prompt = '@'
 
     def send_config_set(self, config_commands=None, exit_config_mode=True, **kwargs):
         """Can't exit from root (if root)"""


ktbyers:
@jumpojoy Problems I see are...some things would break anytime '@' was in the output. Also presupposes what the nature of a given Linux prompt is (as opposed to dynamically determining the prompt).

I think a better solution is to have the library code call self.set_base_prompt() any time a 'sudo su' is executed in the library (executed by user would have to be done manually).

I think that we can fix this by updating some of the methods in the linux_ssh.py module to call set_base_prompt() after a 'sudo su' action.

jumpojoy:
@ktbyers 
'sudo su' will never be completed successfully if base prompt is not correctly set before executing the command.
For example to fix (update) base_prompt for enable mode. We need to call set_base_prompt() after we send a 'sudo su' command L514 but before searching base_prompt L515'
https://github.com/ktbyers/netmiko/blob/master/netmiko/base_connection.py#L514-L516
The same for check_enable_mode, config_mode, check_config_mode.

ktbyers:
@jumpojoy We should be able to make this work without too much difficulty (i.e. make 'sudo su' complete properly even though the prompt changes). Then call self.set_base_prompt() to reset the prompt.

jumpojoy:
@ktbyers 
I have created a new pull request https://github.com/ktbyers/netmiko/pull/194.
After executing ''sudo su" command checking for sudo password prompt 'password'. If not able to find assume that sudo doesn't requires password and setting base_prompt.

The same for exit command. After executing 'exit' updating base_prompt.

Thanks in advance!

ktbyers:
@jumpojoy I will try to look at this (hopefully today or tomorrow).

ktbyers:
@jumpojoy I merged other PR for this issue. I might modify some of that PR to reduce repetitive code (need to see if this is possible or not). 

I also need to add automated tests for Linux into my unit tests.

jumpojoy:
@ktbyers thank you for assistance.  Do you have a guide or how-to for setuping/launching unit tests?

ktbyers:
See https://github.com/ktbyers/netmiko/blob/master/TESTING.md

PR #193, opened 2016-04-18
Add support for Quanta switches

matejv:
This adds support for the default OS that comes with Quanta whitebox switches. Tested on:

```
(Quanta) #sh ver

Switch: 1

System Description............................. LY2R, Runtime Code 1.4.21.00, Linux 2.6.35
Machine Type................................... LY2R
Machine Model.................................. QUANTA LY2R
FRU Number..................................... 1
Part Number.................................... 1LY2BZZ000N
Maintenance Level.............................. 1
Manufacturer................................... 0x0100
Software Version............................... 1.4.21.00
Operating System............................... Linux 2.6.35
Network Processing Device...................... BCM56846_A1

```

### DIFF
--- a/netmiko/quanta/__init__.py
+++ b/netmiko/quanta/__init__.py
@@ -1,3 +1,3 @@
-from netmiko.quanta.quanta_ssh import QuantaSSH
+from netmiko.quanta.quanta_mesh_ssh import QuantaMeshSSH
 
-__all__ = ['QuantaSSH']
+__all__ = ['QuantaMeshSSH']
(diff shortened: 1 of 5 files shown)


ktbyers:
@matejv I am trying to standardize on a naming convention of

vendor_platform_ssh

Does quanta have the equivalent of a platform name? For example, Cisco has 'cisco_ios_ssh', 'cisco_asa_ssh'?

This would help better differentiate between one quanta device_type from another (that run different software).

matejv:
Well there is nothing in the CLI commands and in boot messages that would indicate platform type or OS name.

Looking at their website and docs they have a line of switches called QuantaMesh. These all seem to be running the same OS. So I suppose we could use quanta_mesh_ssh and QuantaMeshSSH.

ktbyers:
@matejv Will try to follow up/merge this PR this week.

PR #222, opened 2016-06-03
send_command_expect loop stops when connection is closed

xavierhardy:
The `send_command_expect` function does not stop on connection loss. This pull request addresses this issue by checking the state of `remote_conn.closed` at each iteration.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -382,7 +382,7 @@ class BaseSSHConnection(object):
         pattern = re.escape(pattern)
         if debug:
             print("Pattern is: {}".format(pattern))
-        while True:
+        while self.remote_conn.closed:
             try:
                 output += self.remote_conn.recv(MAX_BUFFER).decode('utf-8', 'ignore')
                 if re.search(pattern, output, flags=re_flags):
@@ -399,7 +399,7 @@ class BaseSSHConnection(object):
             pattern = self.base_prompt
         pattern = re.escape(pattern)
         base_prompt_pattern = re.escape(self.base_prompt)
-        while True:
+        while self.remote_conn.closed:
             try:
                 output += self.remote_conn.recv(MAX_BUFFER).decode('utf-8', 'ignore')
                 if re.search(pattern, output, flags=re_flags) or re.search(base_prompt_pattern,
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -391,6 +391,7 @@ class BaseSSHConnection(object):
                     return output
             except socket.timeout:
                 raise NetMikoTimeoutException("Timed-out reading channel, data not available.")
+        return output
 
     def read_until_prompt_or_pattern(self, pattern='', re_flags=0):
         """Read until either self.base_prompt or pattern is detected. Return ALL data available."""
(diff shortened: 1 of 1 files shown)


xavierhardy:
You can test by simply sending the `exit` command. It will close the connection, the `send_command_expect` must then return.

ktbyers:
@xavierhardy What is the behavior that you see when the connection is closed?


Kirk

xavierhardy:
It loops until the timeout (max_loops), try with `exit` on any shell that closes the connection on that command, you should be able to reproduce it.

xavierhardy:
I think most people would expect an IOError on connection loss.

xavierhardy:
One more thing: this problem has also repercussions in another case. If you execute a command that results in a connection loss (`exit` in my case), and try to use `disconnect`, it also loops in `read_until_pattern`. My PR would fix that too.

Here is the traceback:
```python
  File "/home/xhardy/test.py", line 99, in disconnect
    connection.disconnect()
  File "/home/xhardy/env/local/lib/python2.7/site-packages/netmiko/base_connection.py", line 703, in disconnect
    self.cleanup()
  File "/home/xhardy/env/local/lib/python2.7/site-packages/netmiko/ssh_connection.py", line 52, in cleanup
    self.exit_config_mode()
  File "/home/xhardy/env/local/lib/python2.7/site-packages/netmiko/ssh_connection.py", line 48, in exit_config_mode
    pattern=pattern)
  File "/home/xhardy/env/local/lib/python2.7/site-packages/netmiko/base_connection.py", line 580, in exit_config_mode
    if self.check_config_mode():
  File "/home/xhardy/env/local/lib/python2.7/site-packages/netmiko/ssh_connection.py", line 30, in check_config_mode
    pattern=pattern)
  File "/home/xhardy/env/local/lib/python2.7/site-packages/netmiko/base_connection.py", line 564, in check_config_mode
    output = self.read_until_pattern(pattern=pattern)
  File "/home/xhardy/env/local/lib/python2.7/site-packages/netmiko/base_connection.py", line 387, in read_until_pattern
    output += self.remote_conn.recv(MAX_BUFFER).decode('utf-8', 'ignore')
  File "/home/xhardy/env/lib/python2.7/encodings/utf_8.py", line 15, in decode
    def decode(input, errors='strict'):
KeyboardInterrupt
```

xavierhardy:
Mea culpa, it wouldn't solve it, because the `disconnect` problem comes from another function (but basically, the same problem): `read_until_pattern`. The fix would be quite simple as well:
```
--while True:
++while not self.remote_conn.closed:
```

xavierhardy:
This PR update fixes them all.

ktbyers:
I am probably not going to make time to look at this for few weeks (currently moving office).

ktbyers:
I don't think this is worth fixing...

If the SSH session is disconnected by a firewall (which it shouldn't do as the session is active and passing data), then there are a whole bunch of ways Netmiko could fail as it is a very unexpected event.

I am open to hearing about real situations when you encounter this issue, however.

xavierhardy:
This is actually worth fixing. Imagine you setup a one minute timeout, if I get disconnected (e.g. the network device on the other side just got reloaded by someone else), this function would fail after one minute, where you could simply have checked the status of the socket and raised the error right away.

Another real situation is when you execute a command that disconnects you from the device by mistake (a wrong route that makes you lose L3 connectivity on the management interface).

This is not the expected behavior, it makes the program using Netmiko lose time, so it is worth fixing.

ktbyers:
@xavierhardy Yes, but this is just one action where Netmiko could encounter this...there are dozens across 20+ drivers.

And it generates an exception, but slower than you would like.

Yes, I disagree with you on the expected behavior as you're constructing scenarios where the device/SSH sessions fails due to something else. In other words, you are saying this very unexpected thing happened, and you are not handling it in the expected way.

PR #226, opened 2016-06-16
Fix for configure mode on Arista 7500 models

rehlers:
Need to look for config), as Arista 7500 has supervisor # (s1) / (s2) in prompt.
This confuses the default check_string, which only looks for ')#'.
Arista 7500 output:
loc1-core01(s1)#configure terminal 
loc1-core01(s1)(config)#

### DIFF
--- a/netmiko/arista/arista_ssh.py
+++ b/netmiko/arista/arista_ssh.py
@@ -2,4 +2,6 @@ from netmiko.ssh_connection import SSHConnection
 
 
 class AristaSSH(SSHConnection):
-    pass
+    def check_config_mode(self, check_string='config)'):
+    	"""Checks if the device is in configuration mode or not."""
+        return super(SSHConnection, self).check_config_mode(check_string=check_string)
--- a/netmiko/arista/arista_ssh.py
+++ b/netmiko/arista/arista_ssh.py
@@ -2,6 +2,6 @@ from netmiko.ssh_connection import SSHConnection
 
 
 class AristaSSH(SSHConnection):
-    def check_config_mode(self, check_string='config)'):
+    def check_config_mode(self, check_string='(config'):
     	"""Checks if the device is in configuration mode or not."""
         return super(SSHConnection, self).check_config_mode(check_string=check_string)


ktbyers:
@rehlers Why does ')#' confuse the default check_string. It should still properly match the patterns shown above?

Also 'config)' won't work because of things like 'config-if-Et6'

```
pynet-sw3(config)#interface ethernet 6
pynet-sw3(config-if-Et6)#
```

rehlers:
The confusion is because the check_config_mode function always returns True on Arista 7500 since the prompt always has a )# at the end of it when in enable mode, even when not in config mode. This causes the exit_config_mode function to fail with ValueError("Failed to exit configuration mode").

Oops didn't think of the config mode, perhaps check_string='(config' instead would work?

ktbyers:
Okay, I see...bad choice Arista...okay, I will have to experiment some in the CLI and see what a good pattern is. I will post what I see here.

ktbyers:
@rehlers FYI, I have recollections that there are contexts where '(con' won't work. This is why I am holding off on this.

rehlers:
ok. don't want to break anything. It (the '(config') has been working for me on our gear, but I am not doing exhaustive testing.

ktbyers:
@rehlers I have been tied up with an office move so I haven't made time to look at this.

ktbyers:
@rehlers I am thinking of using the following (or equivalent not using regular expressions).

config.*)#

Just posted to an Arista channel to see if any reason this won't work.

rehlers:
sounds good, I thought the check for config mode uses a if confg_string in prompt, so would regex .* work? I could be remembering wrong though. Thanks for keeping with this issue, I use your package for a lot of my code.

ktbyers:
@rehlers Yes, I think the underlying method uses a substr in string pattern, but I can overwrite the method in the Arista code if I need to.

ktbyers:
Notes from RANCID as this might help:

```
test = 'loc1-core01(s1)(config)#'
test.replace("(s1)", "")
test.replace("(s2)", "")


import re
ws1 = 'loc1-core01(s1)(config)#'
ws2 = 'loc1-core01(s2)(config)#'
wo = 'loc1-core01(config)#'

regmatch = re.search( '(\(s[1|2]\)|)\(config\)#', ws1)
print regmatch.groups()

regmatch = re.search( '(\(s[1|2]\)|)\(config\)#', ws2)
print regmatch.groups()

regmatch = re.search( '(\(s[1|2]\)|)\(config\)#', wo)
print regmatch.groups()
```

ktbyers:
@rehlers Fix is here:

https://github.com/ktbyers/netmiko/commit/96a6fce610710077dc6b9959a7371dd42b0ea5cb


Please test (if you can)

rehlers:
Confirmed working on Arista 7508 with 2 supervisors.
I downloaded netmiko v1.0.0, attempted config mode change and it failed. I then added the check_config_mode section to arista_ssh.py and it worked.

PR #229, opened 2016-06-23
Add support for private keys

ghmj2417:
Added support to be able to pass in a private key file.  Modified base_connection.py and scp_handler.py to support this feature.  I successfully tested this feature connecting to a Cisco router using a private key as well as I made sure a username and password would still continue to work.  I will also be opening a pull request for scp_sidecar to support private keys.

Let me know if there are any issues with my changes.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -42,6 +42,7 @@ class BaseSSHConnection(object):
         self.port = int(port)
         self.username = username
         self.password = password
+        self.key_file = key_file
         self.secret = secret
         self.device_type = device_type
         self.ansi_escape_codes = False
--- a/netmiko/scp_handler.py
+++ b/netmiko/scp_handler.py
@@ -31,6 +31,7 @@ class SCPConn(object):
                               port=self.ssh_ctl_chan.port,
                               username=self.ssh_ctl_chan.username,
                               password=self.ssh_ctl_chan.password,
+                              key_filename=self.ssh_ctl_chan.key_file,
                               look_for_keys=False,
                               allow_agent=False,
                               timeout=8)


ghmj2417:
Is anyone available to review this change?

ktbyers:
Yes, I will review. I will try to get this done by end of the weekend.

ktbyers:
@ghmj2417 I had to make some minor modifications to this particularly relating to look_for_keys/use_keys. You probably should test it and make sure it works.

ghmj2417:
Yup, I see your changes.  I know from testing and reading paramiko's documentation that when pkey or key_filename is set it will automatically use the key or file.  I believe what you are allowing now is to specify use_keys which will look at pkey, key_filename, id_rsa, etc. Is that correct? 

Thank you for getting this merged. I will test it out this week.  I also put in a PR for the scp sidecar project. Take a gander at that.

ktbyers:
Yes, if you set use_keys it will use the keys defined in ~/.ssh (id_rsa, id_dsa). This also makes it more consistent with the connect definition in the Netmiko-ssh connection (i.e. supports all the same arguments).

ghmj2417:
Nice!  Thank you for the explanation.

I use pip.  For the new update, are you required to manually upload a package or will it do it automatically?

ktbyers:
@ghmj2417   Right now, you would have to 'git clone' the master branch on GitHub (as I have not made a release to pypi that contains this fix).

ghmj2417:
Understood.  Is uploading a new version to pypi on your roadmap?  I just need to figure out if I should proceed with rewriting my code.  If you were going to do it soon then I won't worry about it and will wait for an update to pypi.

ktbyers:
@ghmj2417 I probably won't do it for a while as I released 0.5.6 not too long ago.

PR #234, opened 2016-07-09
Brocade ICX/FastIron support

LindsayHill:
First commit for ICX support. Key point is that the disable_paging command is `skip-page-display`, and it must be entered from enable mode.

I've also provided examples for the test definitions. All tests passing so far.

### DIFF
--- a/netmiko/brocade/brocade_fastiron_ssh.py
+++ b/netmiko/brocade/brocade_fastiron_ssh.py
@@ -2,6 +2,9 @@ from netmiko.ssh_connection import SSHConnection
 
 
 class BrocadeFastironSSH(SSHConnection):
-    '''Placeholder for Brocade FastIron'''
-    def __init__(self, *args, **kwargs):
-        raise NotImplementedError
+    """Brocade FastIron aka ICX support."""
+    def session_preparation(self):
+        """FastIron requires to be enable mode to disable paging."""
+        self.set_base_prompt()
+        self.enable()
+        self.disable_paging(command="skip-page-display")
--- a/tests/brocade_fastiron_commands.txt
+++ b/tests/brocade_fastiron_commands.txt
@@ -1,0 +1,3 @@
+logging buffered 4000
+logging buffered 3000
+no logging console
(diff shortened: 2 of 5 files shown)


LindsayHill:
One thing I have noticed with this is that there's an extra newline in some of the output. 

Assuming I have a connection set up, then run this:
```
config_commands = ['logging buffered warn',
                   'logging buffered notifications',
                   'no logging console']
output = net_connect.send_config_set(config_commands)
print(output)
```
The output contains this:
```
>>> print(output)
config term

SSH@ICX7250-24P Router(config)#logging buffered warn

SSH@ICX7250-24P Router(config)#logging buffered notifications

SSH@ICX7250-24P Router(config)#no logging console

SSH@ICX7250-24P Router(config)#end

SSH@ICX7250-24P Router#
>>>
```
I'm not sure why that extra newline is getting in there. Thoughts?

ktbyers:
I try to normalize the newlines to always be '\n' (as different vendors do different things here).

Here is the default method for this:

```
    @staticmethod
    def normalize_linefeeds(a_string):
        """Convert '\r\r\n','\r\n', '\n\r' to '\n."""
        newline = re.compile(r'(\r\r\r\n|\r\r\n|\r\n|\n\r)')
        return newline.sub('\n', a_string)
```

Try to do the following:

print(repr(output))

So we can see what the characters are. This might give us a hint on what we need to do.

LindsayHill:
@ktbyers here's what I see with that output:
```
>>> config_commands = ['logging console',
...                    'no logging console']
>>> output = net_connect.send_config_set(config_commands)
>>> print(output)
config term

SSH@ICX7250-24P Router(config)#logging console

SSH@ICX7250-24P Router(config)#no logging console

SSH@ICX7250-24P Router(config)#end

SSH@ICX7250-24P Router#
>>> print (repr(output))
u'config term\n\nSSH@ICX7250-24P Router(config)#logging console\n\nSSH@ICX7250-24P Router(config)#no logging console\n\nSSH@ICX7250-24P Router(config)#end\n\nSSH@ICX7250-24P Router#'
>>>
```

I'll start playing around with over-riding `normalize_linefeeds()`, see if I can sort it out

Issue #173, opened 2016-02-04
Break out Brocade adapter for different devices?


LindsayHill:
Brocade has 3 main IP lines - MLX, ICX, VDX. Each run a different OS (NetIron, FastIron, NOS) respectively. There is of course also the FC line, running FabricOS

What's your thoughts on breaking the Brocade adapter up into separate adapters? Something like Brocade_ssh_mlx, or similar?

If I can, I might be able to help with getting initial adapters going for these. Funnily enough, I can get access to some test Brocade gear :)

ktbyers:
@LindsayHill Yes, I want to standardize on the following naming convention:

vendor_platform     

For example,

cisco_ios
cisco_asa
cisco_nxos
hp_procurve

In this file https://github.com/ktbyers/netmiko/blob/master/netmiko/ssh_dispatcher.py

I have:

```python
CLASS_MAPPER_BASE = {
    'cisco_ios': CiscoIosSSH,
    'cisco_xe': CiscoIosSSH,
    'cisco_asa': CiscoAsaSSH,
    'cisco_nxos': CiscoNxosSSH,
    'cisco_xr': CiscoXrSSH,
    'cisco_wlc': CiscoWlcSSH,
    'arista_eos': AristaSSH,
    'hp_procurve': HPProcurveSSH,
    'hp_comware': HPComwareSSH,
    'huawei': HuaweiSSH,
    'f5_ltm': F5LtmSSH,
    'juniper': JuniperSSH,
    'brocade_vdx': BrocadeVdxSSH,
    'a10': A10SSH,
    'avaya_vsp': AvayaVspSSH,
    'avaya_ers': AvayaErsSSH,
    'linux': LinuxSSH,
    'ovs_linux': OvsLinuxSSH,
    'enterasys': EnterasysSSH,
    'extreme': ExtremeSSH,
    'alcatel_sros': AlcatelSrosSSH,
    'fortinet': FortinetSSH,
    'dell_force10': DellForce10SSH,
}
```

So currently I have brocade_vdx.

It sounds like there should also be:

brocade_icx
brocade_mlx

Is it more logical to name it this way or brocade_netiron, brocade_fastiron, etc?

You can see this file for some details on adding a vendor:

https://github.com/ktbyers/netmiko/blob/master/VENDOR.md


Note, the brocade_ssh.py file should probably be renamed to brocade_vdx_ssh.py. This is a pretty easy fix since it should be transparent to people that use Netmiko.

[...]

PR #235, opened 2016-07-09
Enable Brocade MLX/NetIron support

LindsayHill:
Enabled MLX support., related to https://github.com/ktbyers/netmiko/issues/173

Tests all passing. I haven't included my responses/commands.yml files here. Can do if you like


ktbyers:
@LindsayHill Yes, it would be nice if you could add responses/commands.yml. 

This makes it easier for me to know what to expect for unit tests.

Issue #257, opened 2016-08-24
Timing out on authentication if primary auth is unavailable


ghmj2417:
In certain circumstances a user may use TACACS, RADIUS, etc as their primary authentication service (with local being their backup).  In normal operation this should work fine.  In the event TACACS or RADIUS is unavailable there will be a delay in authentication.  Even though we can set a timeout for SSH being established, we cannot set a timeout for authentication success.  I have tested this a few times and I always receive an exception of NetMikoTimeoutException("Timed out waiting for recv_ready").  

base_connection.py, on establish_connection, this gets called; if self.wait_for_recv_ready_newline():
The problem is there is no way for a user to customize the wait time for initial authentication.  I would ask either use the timeout value set, or allow us to use global_delay_factor.

Using Cisco device configured for two RADIUS servers with a timeout of 2 and retransmit of 2.  I use RADIUS as the primary and local as the backup.

ktbyers:
Good idea. Let's add this to the dev_1_0 issues and see if it can be fixed there.

ktbyers:
@ghmj2417 Where exactly do you think a delay needs introduced to fix your above problem?

ghmj2417:
@ktbyers Hmm.  I know the issue occurs in establish_connection at (if self.wait_for_recv_ready_newline():) so it's in that function.  I doubt modifying max_loops would do anything because your max wait time appears to be 8 seconds (I don't think it ever hits this amount with the max_loops being only 20).  Maybe we should focus on delay_factor.  This is where I thought global_delay_factor could be used.

One option; use user assigned global_delay_factor at init, at the beginning of the function (wait_for_recv_ready_newline()) add  delay_factor = self.select_delay_factor(delay_factor) and increase your if delay_factor >= 8: to 20 (or get rid of this check entirely because I don't see why you would want to cap this.)

Second option; Increase max_loops to something much larger and increase the if delay_factor >= 8 to 20, or remove this check entirely.

Third option; remove delay_factor and use the timeout set by the user.  Since this function is only called from establish_connection, it may make sense to use the timeout.

ktbyers:
@ghmj2417 Okay, that code is pretty much all changed in the new dev_1_0 branch (the entire wait_for_recv_ready_newline() method doesn't exist).

https://github.com/ktbyers/netmiko/blob/dev_1_0/netmiko/base_connection.py

So the question is...where should extra sleep be added for devices that take excessively long times to log in. Probably in the session_preparation() method.

[...]

PR #264, opened 2016-09-05
Set retry time to 60 seconds for auth timeouts.

ghmj2417:
This allows for up to 60 seconds of trying to read data from the device during the initial connection.  This is in relation to issue #257  where TACACS, RADIUS, etc are not available and the device may not respond in a timely fashion.  I tested this in my environment and I was able to connect to my devices that have RADIUS as primary and the servers were unavailable.

Couple things: changed the max loop to 60 to resemble 60 seconds.  Changed i += 1 to using main_delay, reason for this; we should probably wait a consistent amount of time and not rely on someone setting global_delay_factor. Using main_delay mimics actual wait time. With this change only affecting the establishment of the connection, I feel 60 seconds is ample amount of time to wait for authentication.  I don't know if I feel allowing a user to specify a high amount for global_delay_factor is the best idea for fixing the timeout issue especially since a global delay factor affects other items.  I know it's best to try and keep things dynamic but I have no problem with this setting being static.

Thoughts?

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -381,7 +381,7 @@ class BaseConnection(object):
         delay_factor = self.select_delay_factor(delay_factor=0)
         main_delay = delay_factor * .1
         time.sleep(main_delay)
-        while i <= 60:
+        while i <= 40:
             new_data = self._read_channel()
             if new_data:
                 break
@@ -391,7 +391,7 @@ class BaseConnection(object):
                 if main_delay >= 8:
                     main_delay = 8
                 time.sleep(main_delay)
-                i += main_delay
+                i += 1
         # check if data was ever present
         if new_data:
             return ""
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -381,7 +381,7 @@ class BaseConnection(object):
         delay_factor = self.select_delay_factor(delay_factor=0)
         main_delay = delay_factor * .1
         time.sleep(main_delay)
-        while i <= 20:
+        while i <= 60:
             new_data = self._read_channel()
             if new_data:
                 break
@@ -391,7 +391,7 @@ class BaseConnection(object):
                 if main_delay >= 8:
                     main_delay = 8
                 time.sleep(main_delay)
-                i += 1
+                i += main_delay
         # check if data was ever present
         if new_data:
             return ""


ghmj2417:
If you want the user to be able to customize the retry time, I can do that but we will need to add another arg into __init__, maybe call it auth_timeout?  Maybe set it to 20 by default?

ktbyers:
Increasing the count to 60 is fine.

Why do this?

                i += main_delay

I didn't understand why you wanted to do that. main_delay will start at .1 and then increase up to a maximum value of 8.

I would propose increase the:

while i <= 60:

but just do the:

                 i += 1

Does this fix your issue?

global_delay_factor will still affect it because of this:

delay_factor = self.select_delay_factor(delay_factor=0)

Basically, I want all sleeps to be affected by global_delay_factor so that ultimately the end-user can adjust this if necessary.

ktbyers:
I think the solution is just to increase the loop count (while i <= 60). It will break out whenever it encounters the first data so devices that respond faster won't be slowed down.

ghmj2417:
If you only change the loop to i <= 60 then at a minimum you could wait a total of 367.3228 seconds before you get to the exception (assuming no data is being read).  On top of the wait time, specifying 60 doesn't really give an indicator of how long you are actually waiting until the exception is thrown.

If you set i += main_delay then your i <= 60 is based off of actual wait time.  Example of setting i += main_delay, let's say main_delay is set to 3, on your first iteration through the loop you will be waiting 3.3 seconds, i now reflects the actual wait time.  Your next iteration, main_delay will equal 3.63, however i will equal 6.93 which reflects actual waiting time.  Hence, now i <= 60 reflects time in seconds of waiting.  I personally like this approach better and that's why I suggested I could add an auth_timeout variable so we can make it i <= self.auth_timeout, instead of i <= 60.  This gives the user flexibility if their environment takes awhile for authentications.

ktbyers:
@ghmj2417 How long a wait does TACACS+/RADIUS require in your case?

There is already an argument for auth_timeout in __init__ the timeout argument. This is actually a post authentication timeout.

We could possibly base the timeout off that. 

I don't think we should spend too much time here. The key thing is just making sure TACACS+ has enough time to clear with reasonable margin to account for different environments. We should give end-users a way to adjust the time if they require using global_delay_factor

ghmj2417:
My scenario, I need at ~8 seconds but I setup my authentication servers for only 2 retries and a timeout of 2.  I think the default setting on a Cisco device is 3 retries with a 5 second timeout.

Last commit should work.  I increased the loop to 40 and changed i back to i += 1.  This, by default, should allow approximately 53 seconds to read data.  With this configuration, users can extend the retry timeout by setting the global_delay_factor.

PR #272, opened 2016-09-13
modify scp_handler.py to support file transfers from remote

pvinci:
scp_handler implemented file transfers to a remote device via
the scp.put() method.  This change adds the ability to copy
files from a remote device via the scp.get() method.


ktbyers:
@pvinci This feature has already been implemented in the dev_1_0 branch which should be released this week. There is a direction argument. There are also separate get_file and put_file methods.


Kirk

PR #273, opened 2016-09-14
Aruba OS Support

mww012:
Adds initial support(enable and config) for Aruba Networks controllers.  This is not support for Aruba Instant APs.


ktbyers:
@mww012 Did you get all of the unit tests to pass on Aruba?

ktbyers:
@mww012 Second question,
```
    def check_config_mode(self, check_string='(config) #', pattern=''):
        """
        Checks if the device is in configuration mode or not.

        Aruba uses "(<controller name>) (config) #" as config prompt
        """
        if not pattern:
            pattern = self.base_prompt[:16]
        return super(ArubaSSH, self).check_config_mode(check_string=check_string,
                                                       pattern=pattern)
```

Why the first 16 characters of base_prompt?

mww012:
Yes, all the unit tests passed.  As far as the 16 character prompt, that was a mistake.  I copied that function over from cisco_base_connection and intended to delete that portion but missed it.  So far, I've not run into a situation where Aruba truncates a prompt.

PR #274, opened 2016-09-15
Add a way to handle confirmation requests

salvoventura:
Commands issued with "send_command" might prompt user for a confirmation question. In NXOS for example that looks like:
   switch# reload
   This command will reboot the system. (y/n)?  [n] 

My patch is quite simplistic, but should serve as a baseline to add more coverage. 
Optional parameter `answer` will apply default reply (if a question gets asked) or use the provided string as reply.

Tested in Cisco NXOS only.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -572,7 +572,7 @@ class BaseConnection(object):
 
     def send_command(self, command_string, expect_string=None,
                      delay_factor=1, max_loops=500, auto_find_prompt=True,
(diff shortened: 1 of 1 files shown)


ktbyers:
@salvoventura Just use send_command_timing() and capture the output and then make a decision based upon it:

```
output = net_connect.send_command_timing("reload")
if '(y/n)' in output:
    net_connect.send_command_timing("y")
```


Kirk

salvoventura:
Indeed that works. Thanks for checking this and the awesome library!

PR #277, opened 2016-09-19
remove read_data from telnet_login

Silent-Bob:
In case the switch does not ask for a username the login would not be possible. With the changed code it works with and without username.

example Cisco config:

> line vty 0 4
>  exec-timeout 0 0
>  password GUESSME
>  login
> 

example output:

> Connected to 255.255.255.255.
> Escape character is '^]'.
> 
> User Access Verification
> 
> Password:


ktbyers:
@Silent-Bob Yes, that is a bug...I missed that earlier.

Issue #279, opened 2016-09-21
Enhancement Request - Provide option to log all read/writes from device to file equivalent to putty session log


halfmetaljacket:
Kirk - First thanks for Netmiko, I've gotten a lot mileage out of it already- I was looking through the code and didn't find a way to capture all output and commands applied to the device by Netmiko, beyond just the responses from explicit send_command/config calls. The idea here being equivalent to a putty or securecrt session log that saves everything the device sees and responds with.

I think the use case here would be not only debugging (both netmiko as well as user scripts that make use of netmiko), but would also provide a sort of audit trail if that were desired by the user.

Now that most all I/O is consolidated down to the read/write_channel methods I think this might be feasible, perhaps selecting on the existing `verbose` option or maybe a new `session_log` option. If you think this is worth looking into I can try my hand at a patch.

Thoughts?

ktbyers:
@halfmetaljacket Yes, I like this idea. It would probably be good to have a way to have it write to a log file and/or standard output.

ktbyers:
This is done, here and some other commits:

https://github.com/ktbyers/netmiko/commit/a1c1c7a6c62aa3af937d591e0285e0b2c23e3a3a

Still might be improvmeents to be made.

PR #284, opened 2016-09-28
Added support for session logging under dev_1_1 branch this time.

halfmetaljacket:
Otherwise the same as the previous PR. Uses the Python logging module. Ref #279

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -147,7 +147,6 @@ class BaseConnection(object):
                 try:
                     # If no data available will wait timeout seconds trying to read
                     output += self.remote_conn.recv(MAX_BUFFER).decode('utf-8', 'ignore')
-                    self.write_to_session_log(output)
                 except socket.timeout:
                     raise NetMikoTimeoutException("Timed-out reading channel, data not available.")
             elif self.protocol == 'telnet':
@@ -155,6 +154,7 @@ class BaseConnection(object):
             if re.search(pattern, output, flags=re_flags):
                 if debug:
                     print("Pattern found: {} {}".format(pattern, output))
+                self.write_to_session_log(output)
                 return output
             time.sleep(loop_delay * self.global_delay_factor)
             i += 1
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -20,7 +20,7 @@ from os import path
 
 from netmiko.netmiko_globals import MAX_BUFFER, BACKSPACE_CHAR
 from netmiko.ssh_exception import NetMikoTimeoutException, NetMikoAuthenticationException
-from netmiko.utilities import write_bytes
+from netmiko.utilities import write_bytes, initiate_session_log
 
 
 class BaseConnection(object):
@@ -32,7 +32,7 @@ class BaseConnection(object):
     def __init__(self, ip='', host='', username='', password='', secret='', port=None,
                  device_type='', verbose=False, global_delay_factor=1, use_keys=False,
                  key_file=None, ssh_strict=False, system_host_keys=False, alt_host_keys=False,
-                 alt_key_file='', ssh_config_file=None, timeout=8):
+                 alt_key_file='', ssh_config_file=None, timeout=8, session_log=None):
 
         if ip:
             self.host = ip
@@ -55,6 +55,10 @@ class BaseConnection(object):
         self.ansi_escape_codes = False
         self.verbose = verbose
         self.timeout = timeout
+        if session_log:
+            self.session_log = initiate_session_log(session_log)
+        else:
+            self.session_log = None
 
         # Use the greater of global_delay_factor or delay_factor local to method
         self.global_delay_factor = global_delay_factor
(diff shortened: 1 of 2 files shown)


netmiko/base_connection.py
```diff
@@ -147,6 +154,7 @@ def _read_channel_expect(self, pattern='', re_flags=0):
             if re.search(pattern, output, flags=re_flags):
                 if debug:
                     print("Pattern found: {} {}".format(pattern, output))
+                self.write_to_session_log(output)
```
halfmetaljacket: This line can be removed if line#149 referenced self.read_channel(). I'm not sure if there was a good reason for that though so I didn't touch it.

ktbyers:
@halfmetaljacket I think this is a good idea...just haven't had time to review / merge it in.

ktbyers:
@halfmetaljacket What is the issue that CustomFileHandler is solving?

halfmetaljacket:
The default file handler unconditionally prints a newline at the end of
every log message. This distorts the device output as any required newlines
are already included. Although py3.4+ modified the logging API to
optionally not print newline, there is no such ability currently for
py2.6/2.7. The only backwards-compatible way to prevent printing the
newline is to sub-class the default `FileHandler` class and override the
relevant method.

ktbyers:
@halfmetaljacket Haven't forgotten about this. I am out traveling this week so it might be hard for me to review. I will try to review it next week.

ktbyers:
@halfmetaljacket I think I might want to do it a different way than you did...so I want to look into this some more.

halfmetaljacket:
Ok. If there's any way I can assist, let me know. This was really just my
first go at this, and I wasn't too sure about the best way to present the
API for it.

ktbyers:
@halfmetaljacket It is all good...mostly haven't had time to work on it :-(

ktbyers:
@halfmetaljacket Okay, I did a first past at logging in the four commits on Dec 7 (to develop branch). Here is one of them.

https://github.com/ktbyers/netmiko/commit/a1c1c7a6c62aa3af937d591e0285e0b2c23e3a3a

Right now I kept it pretty small. I need to look some more at what you did.

halfmetaljacket:
Looked through the commits... I like the more library-centric configuration, but that does come with two tradeoffs:
 Not setting up a handler to output to means that it is up to the users of your library to learn logging and how to roll their own handler -not very friendly-, but that does mean it is more flexible for those who wish to figure out what to do. 
Adding in the additional information to each log entry (i.e. your fourth commit) makes debugging much easier, but it kills the "CLI session-log" look that you would get from CLI tools like putty/securecrt/etc.

I like this setup for creating a library debugging log (and might be a very desirable feature just for that), but for the specific use-case of a more traditional "session log" I'm not sure I would use this. Thoughts?

halfmetaljacket:
BTW, if you're still interested in my original take on this, I can modify it to point it to the right branch and make a few minor improvements.

halfmetaljacket:
I have an idea of how to implement both a debug log and an optional session log using a similar setup to what you put together, if you're interested.

ktbyers:
@halfmetaljacket It is pretty easy for a Netmiko consumer to configure logging, they just have to do this:

```
import logging
logging.basicConfig(filename='test.log', level=logging.DEBUG)
```

Is there something I am missing here?

So the SecureCRT/PuTTY log format just makes it somewhat look like you had copied and pasted a CLI session and put it into your logfile--is that correct?

```
I like this setup for creating a library debugging log (and might be a very desirable feature just 
for that), but for the specific use-case of a more traditional "session log" I'm not sure I would 
use this. Thoughts?
```

For debugging code, it is very helpful to me to see the separate read versus write actions (as opposed to having them co-mingled). 

I am somewhat inclined to think it is better to not log like PuTTY/SecureCRT i.e. to log the separate read and writes and make this distinction clear.


What were you thinking as far as the below?

```
I have an idea of how to implement both a debug log and an optional session log using a 
similar setup to what you put together, if you're interested.
```

halfmetaljacket:
My particular use-case is to use session logging as an audit log for my netmiko scripts - a clean output of everything my scripts do to devices that they touch. For this use-case, CLI-like-logs is what I am looking for. If you forgo that scenario to make this more of a debugging log instead, then yeah configuring logging is easy for your users as you mentioned, and I agree that how you have it configured here should work well for that. You could even expand this to replace all of your "if debug" statements with debug logging, avoiding users live editing netmiko code on their systems.

My thinking for that last part would be to implement both a debug log and a separate (or sub-) logger that is used for CLI-like session logging. If you're interested in supporting two separate use cases, the session/audit logging can be set up very similarly to how you have it set up now.

ktbyers:
@halfmetaljacket Okay that makes sense i.e. we had two different use cases.

I definitely want to keep the debug logging. 

How are you thinking of implementing the session-log?


Kirk

ktbyers:
I think I am just going to leave this as is.

If you want otherwise, let me know and we can discuss...I am not really convinced we need the other way.

Issue #295, opened 2016-10-13
Add support for Python context management (with statement)


jrschneider:
I think it would be neat to incorporate the ability to use the 'with' statement to manage connections to other devices. This would involve adding the __exit__ method to automatically close out any open connections (or maybe some other action is appropriate). 

This alleviates the need for direct connection management (having to remember to close connections), and would be especially useful for connecting to multiple devices within the same script. It can also come in handy in error handling so that we don't need to provide extraneous "close connection" code in the event of exceptions within the context, or wrap everything in massive try/except blocks.

For example:
```
with ConnectHandler(device_type='cisco_ios', ip=host, username=username, password=password) as cmd:

    config_output = cmd.send_command('show running')
    print(config_output)

# now that we've left the context, the cmd object is automatically disconnected and 
# closed out properly, if possible
sys.exit()
```

As of now, running the above code results in the following:

```
Traceback (most recent call last):
  File "/tmp/tmp.ZNdhTJ7Zva/lb-grab.py", line 24, in <module>
    with ConnectHandler(device_type='cisco_ios', ip=host, username=username, password=password) as cmd:

AttributeError: __exit__
```

ebeahan:
I see value in this as well and created a PR with a proposed implementation: https://github.com/ktbyers/netmiko/pull/299

Open to changes or alternative approaches. Here's an example of the current proposed implementation:

```
>>> from netmiko import ConnectHandler
>>> with ConnectHandler(**linux_device) as conn:
...     print("Testing netmiko context manager functionality")
...     print(conn.send_command('uname -a'))
...
Testing netmiko context manager functionality
Linux raspberrypi 4.4.13-v7+ #894 SMP Mon Jun 13 13:13:27 BST 2016 armv7l GNU/Linux
```

ktbyers:
This was completed in the above PR. It is in the dev_1_2 branch.

PR #299, opened 2016-10-23
Proposal to add context manager functionality

ebeahan:
Reference: https://github.com/ktbyers/netmiko/issues/295

I also can see value in being able to use a context manager with the `ConnectHandler` method, and here's a proposed approach to adding that functionality by adding the `__enter__` and `__exit__` methods to `BaseConnection`.

```
>>> from netmiko import ConnectHandler
>>> with ConnectHandler(**linux_device) as conn:
...     print("Testing netmiko context manager functionality")
...     print(conn.send_command('uname -a'))
...
Testing netmiko context manager functionality
Linux raspberrypi 4.4.13-v7+ #894 SMP Mon Jun 13 13:13:27 BST 2016 armv7l GNU/Linux
```

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -88,6 +88,14 @@ class BaseConnection(object):
             self.establish_connection()
             self.session_preparation()
 
+    def __enter__(self):
+        """Enter runtime context"""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Gracefully close connection on context manager exit"""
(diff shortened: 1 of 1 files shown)


ktbyers:
Sounds good. Will review.

ktbyers:
@ebeahan FYI, I also added unit tests for this:

https://github.com/ktbyers/netmiko/commit/d42f1feb4feee4d7547a8007a481d8faa99dd051

PR #307, opened 2016-11-03
Fixed condition to check if vdoms are enabled

hutchris:
When connecting to a Fortinet device with vdoms disabled I found that device.vdoms == True. The condition to check was using output.find("Virtual domain configuration: enabled") which evaluated to True even though the result was -1. This could be unique to my setup as I'm using Windows with python 3.4.2, however I thought it would be best to explicitly test for -1.

### DIFF
--- a/netmiko/fortinet/fortinet_ssh.py
+++ b/netmiko/fortinet/fortinet_ssh.py
@@ -16,7 +16,7 @@ class FortinetSSH(CiscoSSHConnection):
         self.vdoms = False
 
         # According with http://www.gossamer-threads.com/lists/rancid/users/6729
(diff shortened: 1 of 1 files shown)


ktbyers:
@hutchris Yes, that is a bug. 

I fixed it here. Note, I thought it was simpler to change it to a more common pattern and not use find().

https://github.com/ktbyers/netmiko/commit/06900ec80255d049e84c8596076298851dc7b167


Please test it if you can (as I don't have Fortinet devices). The fix is in the dev_1_2 branch.

hutchris:
@ktbyers yeah that makes sense, your search pattern is much better.

I'll test it next week and let you know

PR #315, opened 2016-11-17
update telnet code for cisco console over telnet

ragzilla:
fixes problems with console-over-telnet when not at login prompt yet

using netmiko to connect to devices connected over console port on opengear console servers, when at the initial 'Press RETURN to get started.' the \n sent by netmiko doesn't cause the device to wake, we need to send \r\n or \r&lt;nul&gt; (see Mark's comment at: http://stackoverflow.com/questions/4159342/how-to-send-carriage-return-over-telnet)

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -323,7 +323,7 @@ class BaseConnection(object):
                     if debug:
                         print("checkpoint5")
                     return return_msg
-                self.write_channel("\n")
+                self.write_channel("\r\n")
                 time.sleep(.5 * delay_factor)
                 i += 1
             except EOFError:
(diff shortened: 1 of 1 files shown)


ragzilla:
Not sure how to do this on github without making the pull name incredibly long, but this also merges in a patch to handle 'initial configuration mode' prompts on cisco devices

ktbyers:
@ragzilla On your terminal server issue...so can you explain your setup?

Are you telneting to network device directly (using a port that is mapped on the terminal server)? Or are you telneting to terminal server and then connecting to the network device?

ragzilla:
In this case it's a direct-to-port (unauthenticated telnet is opengear's name for it), you telnet to the opengear on 6000+port and you're connected to the console of the attached device

ktbyers:
@ragzilla I manually merged this into the 'develop' branch. Please test it on your terminal server in the completely blank config case as I made some changes to it.

ktbyers:
Here are the changes I made from your PR:

https://github.com/ktbyers/netmiko/commit/51a39324653c711a946351f1e7628f89d0312652

PR #324, opened 2016-11-29
Added support to aruba wireless controllers for backup

fefa2k:
Implemented a backup method that creates a backup of the flash and pushes it trough ftp, tftp or scp on the destination provided.

### DIFF
--- a/netmiko/aruba/aruba_ssh.py
+++ b/netmiko/aruba/aruba_ssh.py
@@ -1,5 +1,12 @@
 """Aruba OS support"""
 from netmiko.cisco_base_connection import CiscoSSHConnection
+from datetime import date
+import time
+import socket
+import re
+
+from netmiko.netmiko_globals import MAX_BUFFER
+from netmiko.ssh_exception import NetMikoTimeoutException
 
 
 class ArubaSSH(CiscoSSHConnection):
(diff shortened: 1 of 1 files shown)


ktbyers:
@fefa2k I assume this is a pretty standard operation for Aruba Wireless devices?

Also, why did you need to do this?

```
+        # It waits for prompt until timeout is reached
 +        output += self._read_channel_expect_timing(timeout=timeout)
```

Can't you just use?

```
output += self.send_command_timing(delay_factor=some_value)
```

fefa2k:
@ktbyers, yes, it's similar to backing up the "show running-config" output but this backs up the whole flash which contains:
- Software licenses and license database
- Configuration files
- Local user database
- Global AP database
- WMS database
- Custom upload certificates
- Custom upload CP pages
- Any other files uploaded to the controller

https://community.arubanetworks.com/t5/Controller-Based-WLANs/How-do-I-back-up-the-flash-on-the-controller/ta-p/177838

I defined the "_read_channel_expect_timing" because this way it doesn't have to timeout to finish. It waits until the expected output is received (in this case the prompt) or it stops if the timeout is reached.

I tested out with send_command_timing and _read_channel_expect_timing with a timeout value of 20.

- send_command_timing: 44 seconds to finish
- _read_channel_expect_timing: 7 seconds to finish

The latter only waits x seconds for the prompt instead of waiting in case something else is received, does this make sense?

Cheers,

Aarón

ktbyers:
@fefa2k Thanks for the contribution.

The backup method is probably something that should be outside of Netmiko as it looks like a higher level operation than Netmiko usually has.

If there is a lower level method you need for reading the channel, let me know. I think we could probably accomplish that with one of Netmiko's existing methods.

PR #328, opened 2016-11-30
HP Procurve switches new firmware new escape codes on prompt

fefa2k:
HP Procurve switches starting adding "tty=ansi " to the beginning of their prompt plus two escade codes: form feed and carriage return.  I added those to the strip_ansi_escape_codes method to remove them, this way find prompt does work.

This started on models 2530, on firmware versions YA.16.01.0011 and YA.16.01.0011. Probably more switches will follow this path.


ktbyers:
@fefa2k Okay, I merged this change into the dev_1_2 branch.

PR #331, opened 2016-12-02
Issue 325 os guesser addition

maximumG:
The guesser module contains a *Guesser* class, which is used to auto guess the Netmiko *device_type* before creating a new *ConnectionClass* using the *ConnectHandler* function.

The Guesser is using SNMP (v1/v2/v3) to reach a specific OID per device_type - most of the time it will be SysDescr. The value retrieved from the OID must be a string that will be matched by a regular expression.

Please read the code documentation for further option on SNMPv2 and SNMPv3 usage.

Usage
------
```
# SNMPv2
from netmiko.guesser import Guesser
g = Guesser("192.168.1.1", snmp_version="v2", community="read")
type = g.guess()
print type

#SNMPv3
from netmiko.guesser import Guesser
g = Guesser("192.168.1.1", snmp_version="v3", auth_key="auth", encrypt_key="cypher", auth_proto="sha", encrypt_proto="aes256")
type = g.guess()
print type

cisco = {
'device_type': type,
'ip': '192.168.1.1',
'username': 'pyclass',
'password': 'password'}
net_connect = ConnectHandler(**cisco)

### DIFF
--- a/netmiko/guesser.py
+++ b/netmiko/guesser.py
@@ -1,0 +1,269 @@
+"""
+The guesser module is used to Guess the type of a device in order to automatically create a Netmiko ConnectionClass.
+This will avoid to hard code the 'device_type' attribute when using the ConnectHandler function from Netmiko.
+"""
+
+import re
+from pysnmp.entity.rfc3413.oneliner import cmdgen
+
+"""
+This constant dict is matching a 'device type' with 3 defined objects:
+* The OID to GET
+* The regular Expression to match for the OID value you retrieve
+* The priority of the type (the highest the priority is, the better match it will be)
+"""
+SNMP_MAPPER = {
+    'cisco_ios': {"oid": ".1.3.6.1.2.1.1.1.0",
+                  "expr": re.compile(".*Cisco IOS Software,.*", re.IGNORECASE),
+                  "priority": 60},
+    'cisco_xe': {"oid": ".1.3.6.1.2.1.1.1.0",
+                 "expr": re.compile(".*IOS-XE Software,.*", re.IGNORECASE),
+                 "priority": 99},
+    'cisco_asa': {"oid": ".1.3.6.1.2.1.1.1.0",
+                  "expr": re.compile(".*Cisco Adaptive Security Appliance.*", re.IGNORECASE),
+                  "priority": 99},
+    'cisco_nxos': {"oid": ".1.3.6.1.2.1.1.1.0",
+                   "expr": re.compile(".*Cisco NX-OS.*", re.IGNORECASE),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ktbyers** [comment]
I am probably going to remove pysnmp as a dependency from requirements.txt (as pysnmp requires pycrypto which is problematic for windows). The implication will be if you want to use this guesser then you will have to pip install pysnmp separately.

**Comment by ktbyers** [comment]
I will try to review/merge this next week (maybe over the weekend).

**Comment by maximumG** [comment]
Thanks for the reply and your quick review. Hope it will help some folks. If you need further explanation agains the code and the global algorithm, just let me know.

**Comment by ktbyers** [comment]
@maximumG Okay, I manually merged this PR into the develop branch.

I ended up changing things quite a bit...I think it accomplishes the same intent of what you originally wanted. 

See new code:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/snmp_autodetect.py

Main changes I made:
1. I didn't want duplication of device_types in two places so I had the SNMP_MAPPER get checked against the device_types in ssh_dispatcher.py.
2. Removed pysnmp dependency from requirements.txt as I didn't want Netmiko to have this dependency. Mainly because I didn't want the pycrypto dependency (especially since Paramiko no longer has this as a requirement).
3. Changed some of the names: filename, class name, hostname, and changed default to SNMPv3 and AES128.
4. On the autodetect() method used to be guesser(), I sorted by priority so the first match would be used. I also made an cache of SNMP responses so SysDesr SNMP query would only happen once.


Here is an example use:

```
from netmiko.snmp_autodetect import SNMPDetect
 
my_snmp = SNMPDetect(hostname='1.1.1.70', user='pysnmp', auth_key='key1', encrypt_key='key2')
print(my_snmp.autodetect())
```

```
$ python netmiko_guess.py 
cisco_ios
```

We still need to write unit tests for this for SNMPv3 and SNMPv2.

maximumG:
Thanks for the job you did and the modification. How do you want me to start with unit tests?

ktbyers:
@maximumG You too...thanks for the work on this.

We probably need to add a new fixture into conftest.py

https://github.com/ktbyers/netmiko/blob/develop/tests/conftest.py

then call that fixture, here is an example of how I call the fixture for the SCP tests:

https://github.com/ktbyers/netmiko/blob/develop/tests/test_netmiko_scp.py

This document might help. FYI, I can probably create it pretty easily since I am pretty familiar with how the testing is setup...so if you are reverse engineering too much stuff (not a problem).

I think I would like to have at least one SNMPv2 test and one SNMPv3 test that returns an expected device_type.

PR #345, opened 2016-12-27
add 2 new devices : WindowsSSH and Cisco TelePresence endpoints (TC and CE software)

ahmadbarrin:
Hello,

I did some change on major class "base_connection.py" to manage this 2 new devices type. I attached 2 example how to use them in example folder.

I'll be happy if you can take a look on those changes and discuss about it.

Here is my change on existing files:
Add new parameter on BaseConnection object to be more versatile and handle some various device like Windows. 
new parameter on BaseConnection constuctor (__init__) :
- line_feed : specify line break used by device shell. For linux-like OS this is '\n', but on windows client it's '\r\n'. this setting doesn't change line break output returned by netmiko wich is always '\n' (thanks to "normalize_linefeeds" method). Default value : '\n'. 
- shell_encoding : specify device shell character encoding. Previous implementation was utf-8 but Windows device use other encoding (cp850). Default value : 'utf-8'

Child class of BaseConnection can optionnally set some attributes and methods to change some BaseConnection processing.
Bellow is attribute or method that you can set on BaseConnection's child class :
- self.re_prompt_pattern : use this regex pattern as prompt instead of find_prompt()/self.base_prompt. If attribute is not set use auto-finded self.base_prompt as previously. This new setting is necessary for Windows server (where a cd.. will change current prompt) or CiscoTP device where prompt is one of the following 3 values : "OK", "ERROR", "Command not recognized.".
- self.prompt_search_last_depth : when "self.re_prompt_pattern" is set, it's also possible to specify how many line from the end should search for prompt. By default, it's -1 so only search prompt on last line. For Cisco Telepresence device prompt ("OK"/"ERROR"/"Command not recognized.") is on the second last line so it need self.prompt_search_last_depth = -2. search loop begin at -1 (last line) and decrement until self.prompt_search_last_depth.
- child_clean_channel(self, output) : if this method exist, BaseConnection will call it at the end of read_channel(). This permit to do some encoding. This is method is use by Windows SSH device because MS-Dos shell return some special character that will disturb line split, prompt detection and should be display to user anyway.

Thanks,
Ahmad BARRIN


ktbyers:
@ahmadbarrin Thanks for your PR, here are some things you need to change to get this accepted:
1. Re-submit the PR against the develop branch and not against the master branch
2. Separate the WindowsSSH PR from the TelePresence PR (i.e. two independent PRs)
3. On the Windows SSH PR:
    a. Don't make any changes to BaseConnection class. I am unlikely to accept them since Netmiko is a library for supporting networking devices and not for general SSH into Windows/Linux systems. Do everything in a WindowsSSH child class.
    b. Try to avoid adding any new arguments via \_\_init\_\_, discuss this with me in advance if you really need these new arguments. Once again there is a good chance, I won't add them.
4. Both the WindowsSSH and TelePresence PR will need to pass the Netmiko unit tests, see https://github.com/ktbyers/netmiko/blob/develop/TESTING.md. You will need to send me the output from the unit tests.

ahmadbarrin:
Hi Kirk,

Thanks for taking my PR into consideration. It's good to see another CCIE with great programming skill ;)
OK I'll close it and open 2 PR instead. I'll work on it next weekend.
Sorry for all process mistake I didn't make GitHub or other opensource project contribution before :)

Like you asked I'll move all my BaseConnection change into the two child class, but this mean override many method so it will be probably difficult to maintain, but let's do it and see.
To be honest I began working with WindowsSSH class, this was like a training to get familiar with your library because I don't have Cisco TelePresence device at home :) Then I did Cisco TelePresence class, this is the topic where I have some customer need at the moment.

See you next year then
Ahmad

ktbyers:
@ahmadbarrin Only do the WindowsSSH class if you are willing to maintain it. If you aren't willing to maintain it, then don't do it.

It is outside the purpose of the library...so it is questionable if it makes sense to do.

ahmadbarrin:
@ktbyers Yes don't worry I'll be able to maintain both WindowsSSH and CiscoTelepresence if any bug raises with those devices. 
I've made many other applications before and always support them properly.

see you
Ahmad

PR #353, opened 2017-01-11
Fix proxycommand support

thomseddon:
Prior to this, the port was appended to the hostname parameter when given to paramiko.SSHConfig, this causes a config such as:

```
  ProxyCommand ssh jumphost nc %h %p
```

to be expanded to something like:

```
  ssh jumphost nc 10.0.0.1:22 22
```

Which isn't correct according to ssh_configs doc's on the usage of the replacements:

> In the command string, any occurrence of ‘%h’ will be substituted by the host name to connect, ‘%p’ by the port, and ‘%r’ by the remote user name.

This bug was introduced in 649dbcd4e1

------

The commit that introduces this isn't super clear on why this was done, I may be missing something but I'm not sure how this can be working in it's current state.

Thanks!

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -394,8 +394,7 @@ class BaseConnection(object):
             ssh_config_instance = paramiko.SSHConfig()
             with open(full_path) as f:
                 ssh_config_instance.parse(f)
-                host_specifier = "{0}:{1}".format(self.host, self.port)
-                source = ssh_config_instance.lookup(host_specifier)
+                source = ssh_config_instance.lookup(self.host)
         else:
             source = {}
 


thomseddon:
*commit message/description updated to reflect `ssh_config`'s docs

ktbyers:
@thomseddon It should cause it to be expanded to:

```
ProxyCommand ssh jumphost nc %h %p
```

so

```
ssh jumphost nc 1.1.1.1 22
```

I have tested this and this works fine, can you clarify what the issue is?

ktbyers:
Here is an article where I tested this, worked through this:

https://pynet.twb-tech.com/blog/automation/netmiko-proxy.html

thomseddon:
Hey,

Before this change, the code was:

```python
host_specifier = "{0}:{1}".format(self.host, self.port)
source = ssh_config_instance.lookup(host_specifier)
```

So the `hostname` that's being passed to [paramiko.SSHConfig.lookup](https://github.com/paramiko/paramiko/blob/master/paramiko/config.py#L99-L139) (via `host_specifier`) was actually the manually joined host and port e.g. `10.0.0.1:22`.

Although it isn't a valid hostname, this will match rules such as:

```
Hostname 10.*
```

Paramiko will expand `%h` to be whatever it was originally passed as the hostname parameter: https://github.com/paramiko/paramiko/blob/master/paramiko/config.py#L135

Hence the proxycommand string returned from paramiko is:

```
 ssh jumphost nc 10.0.0.1:22 22
```

This was confirmed by adding a print statement here: https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L401 before the change

What do you see if you print out the source object returned from paramiko?

ktbyers:
Yes, Netmiko doesn't support SSH wildcard globbing (in .ssh/config)

How would your solution handle different SSH ports? For example,

```
host 10.10.10.27:22             # Cisco IOS router pynet-rtr1
  user pyclass
  hostname 10.10.10.27
  Port 22
  ProxyCommand ssh jumphost nc %h %p    

host 10.10.10.27:8022           # Cisco IOS router pynet-rtr2
  user pyclass
  hostname 10.10.10.27
  Port 8022
  ProxyCommand ssh jumphost nc %h %p
```

The hostname and port in the above cases will be the correct value.

Your change would also break existing code...which I am pretty reluctant to do.

I think there is a separate issue/discussion somewhere regarding SSH config and wildcards...unfortunately, I think there is probably going to be a lot of work to make that work properly and to handle other meaningful use cases (like support for different SSH ports).

thomseddon:
Ah, whilst that joined hostname:port format _works_, I don't think it's actually intended/support by ssh_config and paramiko - the ssh_config man page doesn't seem to mention it.

Looking again, the reason you're format works is because of the explicit definition of the hostname again in the matching block, paramiko seems to prefer this definition when expanding: https://github.com/paramiko/paramiko/blob/master/paramiko/config.py#L172-L175

But this does mean that you _have_ to add the hostname definition in ssh_config, for every host, for proxycommand to work with netmiko (worth documenting?) - however this doesn't work in my case and doesn't seem practical on a real network with 100/1000's devices?

Looking at your example, wouldn't the following would work for both, as long as you passed the intended port to netmiko:

```
host 10.10.10.27
  user pyclass
  ProxyCommand ssh jumphost nc %h %p    
```

If the above is correct, my change would break backwords compatibility where the hostnames are being defined in ssh_config in the format given in your example (how many people do this?), we could add a flag to enable this matching (the current style) and then look at changing it in the future.

In any case, we can certainly add some more documentation which may help others

ktbyers:
@thomseddon 

Let me re-test with this:

```
host 10.10.10.27
  user pyclass
  ProxyCommand ssh jumphost nc %h %p    
```

I might have been explicitly specifying the hostname / Port arguments when I tried this previously (as opposed to just having the port pass directly to %p) so maybe that was what was causing me the problems earlier.

Anyways, let me re-test it.

Yes, if this works with port overload issue, and makes it more standard with expected way SSH config/Paramiko (and potentially fixes SSH glob problem), then it is probably worth doing.

ktbyers:
Made final change of this to fix the host_specifier issue (i.e. to make it operate in the proper way for the SSH config file).

I tested this using both an IP address and a hostname. You can just embed the port in the SSH config.

PR #357, opened 2017-01-16
Pass through all ssh_config to scp

thomseddon:
Hey, still plugging away with ProxyCommand :)

This commit adds support for passing through all ssh connection configuration to the scp_handler for use when creating a new ssh client.

I wasn't sure on the most consistent way to implement this, I've gone with returning the `ssh_connect_params` and `SSHClient` seperately, leaving the caller to then call `open` with them both. It could be neater just to call open within `build_ssh_client` (perhaps rename it).
Another option would be just copy pasta the block from base_connection too scp_handler...

Depends on 91450d1

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -394,8 +394,7 @@ class BaseConnection(object):
             ssh_config_instance = paramiko.SSHConfig()
             with open(full_path) as f:
                 ssh_config_instance.parse(f)
-                host_specifier = "{0}:{1}".format(self.host, self.port)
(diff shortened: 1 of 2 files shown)


ktbyers:
I might restructure this slightly...but definitely needs done. Thanks.

ktbyers:
@thomseddon Note, I made some changes to this here.

https://github.com/ktbyers/netmiko/commit/e786e1061a709757345e532c9e5b519e90c36128

I also backed out (temporarily) the host_specifier change as I need to test that (and I probably need to add/expand some unit tests).

PR #360, opened 2017-01-19
Develop

plejon:
adding support mt5600t.


plejon:
i have tested the following:

find_prompt() user/enable/config
enable()
config_mode()
check_config_mode()
exit_config_mode()
send_command()
send_config_set()

plejon:
Model name is MA5600T. not mt5600t
just typo..

ktbyers:
Sorry for slow response...

Structure of this needs to be: vendor_platform_transport so it should be. It also looks like we probably shouldn't have the 't' as there is a ma5603u

```
huawei_ma5600_ssh 
HuaweiMA5600SSH
```

ktbyers:
Also can you run all of the config and show unit tests against your driver.

See this document here:

Let me know if you need assistance.

ktbyers:
Closing...happy to accept this, but a few other things need done (see above).

PR #363, opened 2017-01-24
Increase default delay factor for BaseConnection.set_base_prompt()

amberik:
When establishing new connection to Cisco switch devices connection is failing with Exception: "Router prompt not found: <last line of welcome message>"
I found that on those Cisco switches there is a delay (~1 sec.) between "welcome message" and "prompt message". By increasing default delay factor for BaseConnection.set_base_prompt() I managed to connect to those devices.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -576,7 +576,7 @@ class BaseConnection(object):
         return output
 
     def set_base_prompt(self, pri_prompt_terminator='#',
-                        alt_prompt_terminator='>', delay_factor=1):
+                        alt_prompt_terminator='>', delay_factor=10):
         """
         Sets self.base_prompt
 


ktbyers:
@amberik I agree this is a problem that needs improved in Netmiko.

In other words, a generic issue that is run into is that not enough time is allowed post login to read banner and other text at login...so the set_base_prompt() message fails.

I am going to look at the code a bit more and see if where you put the fix is the most logical place to put it (it might be).

amberik:
You are right, the issue is what you've just described. 
Proposed solution probably not elegant but it looks like it is isolated only by connection use case because set_base_prompt is invoked only when connection just established. 
Thank you!

ktbyers:
@amberik I put this fix into the develop branch:

https://github.com/ktbyers/netmiko/commit/a03638c1a2447204841bbd62af2d9c8d7b8ff60c

Can you see if it fixes the problem you saw?

PR #367, opened 2017-02-06
Adds Ciena SAOS support (experimental)

rodecker:
This adds support for Ciena SAOS carrier Ethernet switches. SAOS does not have enable or config mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Enterasys
 Extreme  
 F5 LTM  
 Fortinet  
-
+Ciena SAOS
 
 <br>
 ## Tutorials:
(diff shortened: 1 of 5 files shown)


ktbyers:
@rodecker Can you run the standard unit tests against this driver? This should be both the 'show' and 'config' unit tests. There might be some that will fail...need to be skipped given that there is no enable or config mode.

See this document:

Let me know if you need assistance.

rodecker:
@ktbyers test_netmiko_show.py passes. Indeed I need to disable test_config_mode and test_exit_config_mode for test_netmiko_config.py to pass.

ktbyers:
@rodecker Okay, so all the tests in `show` pass and all the tests in `config` pass except for `test_config_mode` and `test_exit_config_mode`?

rodecker:
@ktbyers yes

PR #371, opened 2017-02-16
Added three additional ansi escape codes

fefa2k:
I stumbled upon these while upgrading the firmware of HP ProCurve/Aruba switches (2530).

The codes I found were:

 - [?6l -> Reset mode screen with option 6: 640 x 200 monochrome (graphics). The number specifies the width of the screen, it can go from 0 to 19, except 7
that resets line wrapping.
 - [?7l -> Reset line wrapping.
 - [2J -> Erase display.

I added to the code the whole range of options for resetting the width of the screen (from 0-12 -> \d+), the reset of the line wrapping and the erase display code. The reset of the line wrapping is a bit redundant but that way it's specified that it is indeed removed.


ktbyers:
@fefa2k I assume you tested this and it resolves the issues you saw with the HP ProCurve/Aruba switches?

fefa2k:
Hi @ktbyers, sorry for the delay. Yes, both platforms were tested with these escape codes. After uploading a firmware to the any flash it validates the firmware and then the screen gets wiped clean, I do believe that was the main issue here. With these additional ansi escape codes that problem went away.

PR #372, opened 2017-02-17
Telnet connection improvement. Fix check_config_mode, check_enable_mode. EdgeCore switch support.

shamanu4:
Telnet connection improvement:

- login/password detection case insensitive, banners ignored (multiline)
- fix for switches that does not produce output for the first telnet connection read attempt

Fix check_config_mode, check_enable_mode:

- replaced `check_string in output` with `output.endswith(check_string)` - check_string can be present in prompt and checks will always return True.
- replaced cehck_string for check_config_mode for cisco from `)#` to `config)#` - same reason as above.

EdgeCore support:

- Telnet connection class for EdgeCore ES series switches. (Maybe not only ES series)

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -288,7 +288,7 @@ class BaseConnection(object):
                      delay_factor=1, max_loops=60):
         """Telnet login. Can be username/password or just password."""
         TELNET_RETURN = '\r\n'
-        debug = True
+        debug = False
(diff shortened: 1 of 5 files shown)


shamanu4:
EdgeCore base prompt issue. Need to be fixed.
tty number is added to prompt e.g. switch-0> switch-1> switch-2#.

Issue #376, opened 2017-02-17
`#` symbol in hostname - check_enable_mode always returns True


shamanu4:

ktbyers:
Why does the hostname have a '#' in it?

Is this a function of the EdgeCore device or the hostname you chose?

shamanu4:
It is possible to set such hostname even in Cisco ios. And fix is pretty simple - change string check from `in` to `endswith`

ktbyers:
If you do this (i.e. use a hostname with # in it), you are setting yourself up for automation failure...

shamanu4:
Ok. Not an issue then.

Issue #375, opened 2017-02-17
Parenthesis at the end of hostname - check_config_mode always returns True


shamanu4:
Example scenario: IP address added to hostname and in result prompt looks like this:
`Switch(127.0.0.1)#` and in config mode like this `Switch(127.0.0.1)(config)#`
Checking for string `)#` returns `True` in both cases.

ktbyers:
EdgeCore is not supported. You should make any necessary adjustments in the EdgeCore driver PR.

shamanu4:
@ktbyers 
It is not function of EdgeCore. I have hundreds of Cisco switches in network configured in this way and I need to manage them. Hostname contains physical address and 2 last bytes of IP e.g. `European.76(3.11)`.

ktbyers:
@shamanu4 Yes, I don't see myself changing this. 

It is very hard to meet all the requirements and changing nature of the router prompt as it is. There are certain requirements the hostname will have to meet i.e. no '#' in it and no parenthesis in it.

PR #377, opened 2017-02-17
Fixes for issues #375 and #376. Telnet connection improvement.  EdgeCore switch support.

shamanu4:
Telnet connection improvement:

- login/password detection case insensitive, banners ignored (multiline)
- fix for switches that does not produce output for the first telnet connection read attempt

Fixes for issues #375 and #376 :

- replaced `check_string in output` with `output.endswith(check_string)` - check_string can be present in prompt and checks will always return True.
- replaced cehck_string for check_config_mode for cisco from `)#` to `config)#` - same reason as above.

EdgeCore support:

- Telnet connection class for EdgeCore ES series switches. (Maybe not only ES series)

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -288,7 +288,7 @@ class BaseConnection(object):
                      delay_factor=1, max_loops=60):
         """Telnet login. Can be username/password or just password."""
         TELNET_RETURN = '\r\n'
-        debug = True
+        debug = False
         if debug:
             print("In telnet_login():")
         delay_factor = self.select_delay_factor(delay_factor)
@@ -838,7 +838,7 @@ class BaseConnection(object):
         output = self.read_until_prompt()
         if debug:
             print(output)
-        return check_string in output
+        return output.endswith(check_string)
 
     def enable(self, cmd='', pattern='password', re_flags=re.IGNORECASE):
         """Enter enable mode."""
@@ -871,7 +871,7 @@ class BaseConnection(object):
         output = self.read_until_pattern(pattern=pattern)
         if debug:
             print("check_config_mode: {}".format(repr(output)))
-        return check_string in output
+        return output.endswith(check_string)
 
     def config_mode(self, config_command='', pattern=''):
         """Enter into config_mode."""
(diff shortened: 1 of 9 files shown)


shamanu4:
Tests runs with no errors.
py.test -v test_netmiko_show.py --test_device cisco_ios
py.test -v test_netmiko_config.py --test_device cisco_ios
py.test -v test_netmiko_show.py --test_device cisco_telnet
py.test -v test_netmiko_config.py --test_device cisco_telnet
py.test -v test_netmiko_show.py --test_device edge_core_telnet
py.test -v test_netmiko_config.py --test_device edge_core_telnet

ktbyers:
@shamanu4 First of all thanks for your contribution.

In general, you need to move almost all of these changes to the EdgeCore driver and not make changes in base_connection.py or in cisco_base_connection.py

1. I am pretty unlikely to accept new arguments to __init__ (i.e. protocol). Netmiko already has a way of determining protocol from device type. So I am not going to accept the add of `protocol`.

2. Don't change the telnet_login method in base_connection. If telnet_login doesn't work for EdgeCore than just overwrite the method in the EdgeCore driver.

3. Don't change the `check_config_mode` in cisco_base_connection. Cisco devices will do things like `config-if)#` so your change won't work (for Cisco devices).

4. Don't change names in ssh_dispatcher. I am very unlikely to accept them. Obviously, you need to add your new driver here, but minimize your changes just to what is needed to do this.

shamanu4:
OK. I'll drop all not mandatory code for EdgeCore and will make new pull request. Packing several features in single pull request was a bad idea.

shamanu4:
@ktbyers 
OK. I'll drop all not mandatory code for EdgeCore and will make new pull request. Packing several features in single pull request was a bad idea.

shamanu4:
@ktbyers
OK. I'll drop all not mandatory code for EdgeCore and will make a new pull request. Packing several features in single pull request was a bad idea.

ktbyers:
@shamanu4 No problem at all...obviously a lot of the code in `base_connection.py` and `cisco_base_connection.py` has been worked on and tested extensively so I am cautious about making significant changes to it.

PR #387, opened 2017-02-28
Add SSH autodetect feature

maximumG:
I finished my work on SSH autodetect.

The global idea:

- Having a dedicated module and class for device type detection over SSH : SSHDetect
- Each driver should implement a static method called autodetect. This method has a single argument with is the netmiko connection using 'terminal_server'. This method return an integer between 0 and 99 (confidence in autodetect)
- Currently work with IOS, Huawei and NX-OS

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,11 +47,12 @@ Vyatta VyOS
 ###### Experimental
 A10  
 Alcatel-Lucent SR-OS  
+Ciena SAOS  
+Cisco Telepresence  
 Enterasys  
 Extreme  
 F5 LTM  
 Fortinet  
-Ciena SAOS
 
 <br>
 ## Tutorials:
--- a/netmiko/__init__.py
+++ b/netmiko/__init__.py
@@ -7,6 +7,7 @@ log.addHandler(logging.NullHandler()) # noqa
 
 from netmiko.ssh_dispatcher import ConnectHandler
 from netmiko.ssh_dispatcher import ssh_dispatcher
+from netmiko.ssh_dispatcher import redispatch
 from netmiko.ssh_dispatcher import platforms
 from netmiko.scp_handler import SCPConn
 from netmiko.scp_handler import FileTransfer
@@ -22,7 +23,7 @@ __version__ = '1.2.8'
 
 __all__ = ('ConnectHandler', 'ssh_dispatcher', 'platforms', 'SCPConn', 'FileTransfer',
            'NetMikoTimeoutException', 'NetMikoAuthenticationException',
-           'NetmikoTimeoutError', 'NetmikoAuthError', 'InLineTransfer')
+           'NetmikoTimeoutError', 'NetmikoAuthError', 'InLineTransfer', 'redispatch')
 
 # Cisco cntl-shift-six sequence
 CNTL_SHIFT_6 = chr(30)
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -88,7 +88,6 @@ class BaseConnection(object):
         :type timeout: float
 
         """
-
         if ip:
             self.host = ip
             self.ip = ip
--- a/netmiko/cisco/cisco_ios.py
+++ b/netmiko/cisco/cisco_ios.py
@@ -6,6 +6,21 @@ from netmiko.cisco_base_connection import CiscoTelnetConnection
 
 class CiscoIosSSH(CiscoSSHConnection):
     """Cisco IOS SSH driver."""
+
+    @staticmethod
+    def autodetect(session):
+        """
+        """
+        matches = ["Cisco IOS Software", "Cisco Internetwork Operating System Software"]
+        try:
+            response = session.send_command("show version | inc Cisco")
+            for m in matches:
+                if m in response:
+                    return 99
+        except:
+            return 0
+        return 0
+
     def session_preparation(self):
         """Prepare the session after the connection has been established."""
         self.set_base_prompt()
--- a/netmiko/cisco/cisco_nxos_ssh.py
+++ b/netmiko/cisco/cisco_nxos_ssh.py
@@ -6,6 +6,21 @@ from netmiko.cisco_base_connection import CiscoSSHConnection
 
 
 class CiscoNxosSSH(CiscoSSHConnection):
+    
+    @staticmethod
+    def autodetect(session):
+        """
+        """
+        matches = ["Cisco Nexus Operating System", "NX-OS"]
+        try:
+            response = session.send_command("show version | inc Cisco")
+            for m in matches:
+                if m in response:
+                    return 99
+        except:
+            return 0
+        return 0
+    
     def session_preparation(self):
         """
         Prepare the session after the connection has been established.
--- a/netmiko/huawei/huawei_ssh.py
+++ b/netmiko/huawei/huawei_ssh.py
@@ -6,6 +6,18 @@ from netmiko.cisco_base_connection import CiscoSSHConnection
 
 class HuaweiSSH(CiscoSSHConnection):
 
+    @staticmethod
+    def autodetect(session):
+        """
+        """
+        try:
+            response = session.send_command("display version | inc Huawei")
+            if 'Huawei' in response:
+                return 99
+        except:
+            return 0
+        return 0
+
     def session_preparation(self):
         """Prepare the session after the connection has been established."""
         self.set_base_prompt()
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -1,0 +1,74 @@
+"""
+This module is used to auto-detect the type of a device in order to automatically create a
+Netmiko connection.
+
+The will avoid to hard coding the 'device_type' when using the ConnectHandler factory function
+from Netmiko.
+"""
+
+from netmiko.ssh_dispatcher import CLASS_MAPPER_BASE, ConnectHandler
+
+SSH_MAPPER_BASE = {}
+for k, v in CLASS_MAPPER_BASE.iteritems():
+    if getattr(v, "autodetect", None):
+        SSH_MAPPER_BASE[k] = v
+
+
+class SSHDetect(object):
+    """
+    The SSHDetect class tries to automatically guess the device type running on the SSH remote end.
+
+    Parameters
+    ----------
+    *args : list
+        The same *args that you might provide to the netmiko.ssh_dispatcher.ConnectHandler.
+    *kwargs : dict
+        The same *kwargs that you might provide to the netmiko.ssh_dispatcher.ConnectHandler.
+
+    Attributes
+    ----------
+    connection : netmiko.terminal_server.TerminalServer
+        A basic connection to the remote SSH end.
+    potential_matches: dict
+        Dict of (device type, accuracy) that is populated trough a interaction with the remote end.
+
+    Methods
+    -------
+    autodetect()
+        Try to determine the device type.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Constructor of the SSHDetect class
+        """
+        self.connection = ConnectHandler(*args, **kwargs)
+        if self.connection.device_type != "terminal_server":
+            self.connection.disconnect()
+            raise ValueError("The connection device_type must be of 'terminal_server'")
+
+        self.potential_matches = {}
+
+    def autodetect(self):
+        """
+        Try to guess the best 'device_type' based on each device_type autodetect method.
+
+        Returns
+        -------
+        bast_match : str or None
+            The device type that is currently the best to use to interact with the device
+        """
+        for k, v in SSH_MAPPER_BASE.items():
+            try:
+                accuracy = v.autodetect(self.connection)
+                self.potential_matches[k] = accuracy
+            except:
+                pass
+
+        if not self.potential_matches:
+            self.connection.disconnect()
+            return None
+
+        best_match = sorted(self.potential_matches.items(), key=lambda t:t[0])
+        self.connection.disconnect()
+        return best_match[0][0]
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -9,7 +9,7 @@ from Netmiko.
 from netmiko.ssh_dispatcher import CLASS_MAPPER_BASE, ConnectHandler
 
 SSH_MAPPER_BASE = {}
-for k, v in CLASS_MAPPER_BASE.iteritems():
+for k, v in CLASS_MAPPER_BASE.items():
     if getattr(v, "autodetect", None):
         SSH_MAPPER_BASE[k] = v
 
@@ -69,6 +69,6 @@ class SSHDetect(object):
             self.connection.disconnect()
             return None
 
-        best_match = sorted(self.potential_matches.items(), key=lambda t:t[0])
+        best_match = sorted(self.potential_matches.items(), key=lambda t:t[1], reverse=True)
         self.connection.disconnect()
         return best_match[0][0]
--- a/netmiko/ssh_dispatcher.py
+++ b/netmiko/ssh_dispatcher.py
@@ -33,6 +33,8 @@ from netmiko.vyos import VyOSSSH
 from netmiko.ubiquiti import UbiquitiEdgeSSH
 from netmiko.ciena import CienaSaosSSH
 from netmiko.cisco import CiscoTpTcCeSSH
+from netmiko.terminal_server import TerminalServerSSH
+from netmiko.terminal_server import TerminalServerTelnet
 
 
 # The keys of this dictionary are the supported device_types
@@ -75,6 +77,7 @@ CLASS_MAPPER_BASE = {
     'ubiquiti_edge': UbiquitiEdgeSSH,
     'ciena_saos': CienaSaosSSH,
     'cisco_tp': CiscoTpTcCeSSH,
+    'generic_termserver': TerminalServerSSH,
 }
 
 # Also support keys that end in _ssh
@@ -87,6 +90,10 @@ CLASS_MAPPER = new_mapper
 
 # Add telnet drivers
 CLASS_MAPPER['cisco_ios_telnet'] = CiscoIosTelnet
+CLASS_MAPPER['generic_termserver_telnet'] = TerminalServerTelnet
+
+# Add general terminal_server driver
+CLASS_MAPPER['terminal_server'] = TerminalServerSSH
 
 platforms = list(CLASS_MAPPER.keys())
 platforms.sort()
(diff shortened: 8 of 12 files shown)


ktbyers:
@maximumG Some feedback on this:

You should create an `autodetect` method in base_connection.py and eliminate your code duplication.

Move this to base_connection.py and then in the child drivers just call the parent class and pass in the search strings.

```
+    @staticmethod
 +    def autodetect(session):
 +        """
 +        """
 +        matches = ["Cisco IOS Software", "Cisco Internetwork Operating System Software"]
 +        try:
 +            response = session.send_command("show version | inc Cisco")
 +            for m in matches:
 +                if m in response:
 +                    return 99
 +        except:
 +            return 0
 +        return 0
```

So I would think the definition of autodetect in base_connection.py should be:

```
def autodetect(self, matches, default_priority=99):
```

Not sure what you want the default_priority to be on a success.

Also, we should make this be a normal method and not a static method (since the `session` is already bound to the object).

I would also vote, we don't do a bare except so at a minimum `except Exception` (or better would be to specify the exceptions we are looking for. Note, it is probably better to get some exceptions here (otherwise we are probably hiding errors by autodetect just returning 0).

ktbyers:
Note, you should also run `pylama .` on your code (from the directory containing the setup.cfg file) and fix any of the warnings/errors you get.

Thanks.

Wyko:
@maximumG, would you like me to make the edits?

maximumG:
@Wyko : It would be great if you can handle the edits. I won't be available until next week. Thanks !

@ktbyers : Thanks for your review. here are some point I want to discuss

> Also, we should make this be a normal method and not a static method (since the session is already bound to the object).

I'm not really sure that this will work as my session is of type 'terminal_server'. That is why I created static method in the first place. If I changed *autodetect* method to a class method I will end up by calling the autodetect of my TerminalServer class or will need to create a new object for each subclass (IOS, NXOS, Huawei,...).

By doing this way I only create one single basic SSH session (terminal_server). Basically when I iterate over the SSH_MAPPER_BASE items, these are just class and not object so I cannot call a class method at this point or I will need to instantiate one new object per items in the for loop. This will end with a SSH connection per item.

```python
       for k, v in SSH_MAPPER_BASE.items():
             try:
                 accuracy = v.autodetect(self.connection)
                 self.potential_matches[k] = accuracy
             except:
                 pass
```

> I would also vote, we don't do a bare except so at a minimum except Exception (or better would be to specify the exceptions we are looking for. Note, it is probably better to get some exceptions here (otherwise we are probably hiding errors by autodetect just returning 0).

This could be easily changed to catch *Exception*. For now as used a global Exception as I don't really know what the function could raise.

> So I would think the definition of autodetect in base_connection.py should be 
def autodetect(self, matches, default_priority=99):

I thought of using this method definition in the first place. However I wanted to keep this autodetect as generic as possible and let anybody use their own logic to implement the detection. If for some reason, someone prefer to use something different than matching some strings, he would be able to do so (reading the SSH banner, capturing an SSH footprint or whatever).

ktbyers:
@maximumG Okay, that makes sense on the @staticmethod and passing in the `session` (i.e. I agree with they way you did it).

We should make the autodetect method in each Netmiko class be a private method i.e. `def _autodetect` i.e. it is not expected that end-users call this. The `autodetect` in the SSHDetect class should still be a public method as end-users will call that.

Yes, lets make it `except Exception` we can possibly try to narrow it later. 

```
I thought of using this method definition in the first place. However I wanted to
keep this autodetect as generic as possible and let anybody use their own logic to
 implement the detection. If for some reason, someone prefer to use something 
different than matching some strings, he would be able to do so (reading the SSH
banner, capturing an SSH footprint or whatever).
```

I don't see the issue? Someone could still overwrite the base `autodetect` in the child class. So they can still do a completely different implementation if they want? 

Probably should do this check (before the connection):

```
if self.connection.device_type != "terminal_server":
```

Is `terminal_server` the logical device_type to use here. Probably should just make a new name in ssh_dispatcher.py `autodetect` and map it to the TerminalServerSSH class?

maximumG:
@ktbyers : Thanks for your feedback, really appreciate your comments and the discussion :smile:

Okay for changing **autodetect** public method to **_autodetect** private method. I took the wrong habit in python as everything is accessible from everywhere, but the convention '_' is still useful to indicate too people 'warning private method'.

> I don't see the issue? Someone could still overwrite the base autodetect in the child class. So they can still do a completely different implementation if they want?

The issue I see here is that I am calling each autodetect static method for each device_type. If we use the following definition for each method, we ask everybody to follow this convention.

```python
def _autodetect(session, matches)
    """
    Parameters
    -----------
    session : netmiko.TerminalServer
        The netmiko session that we can work with to try autodetection.
    matches : list
        List of string might matches after sending some commands
    """
```

As this static method is called from **SSHAutoDetect** class, I think that the matches parameter is useless as it would mean that my SSHAutoDetect method needs to have a list of all matches per device_type. The only thing that might be useful to me is the session. This session can be used and manipulated by each device_type to try to guess the device_type in their own _autodetect method.

The number returned by the _autodetect method corresponds to the confidence of the guessing. If for some reason the best match is actually not the good one (detected during a conversation with the device afterward), the user still can initate a new connection with the 2nd best match.

> I don't see the issue? Someone could still overwrite the base autodetect in the child class. So they can still do a completely different implementation if they want?

It make sense to me and would be more logical. Could you handle the modification in *ssh_dispatch.py* module ?

Hope you will understand my thought 😃

ktbyers:
@maximumG I will see if I can review this tomorrow. I quickly looked at it, but need to look at it more closely.

ktbyers:
@maximumG How about this?

In the SSHDetect class, autodetect() method, I think we should make our call to each driver as follows:

```python
        for k, v in SSH_MAPPER_BASE.items():
            try:
                accuracy = v._autodetect(self.connection)
                self.potential_matches[k] = accuracy
            except Exception:
                pass
```

In base_connection.py, BaseConnect class, we should define `_autodetect()` similar to the following:

```python
    @staticmethod
    def _autodetect(session, cmd="", search_patterns=None, re_flags=re.I, default_priority=99):
        """Default _autodetect method.

        search_list is a regular expression pattern.
        """ 
        if not cmd or not search_list:
            return 0
        try:
            response = session.send_command(cmd)
            for pattern in search_patterns:
                match = re.search(pattern, response, flags=re.I)
                if match:
                    return default_priority
        except Exception:
            return 0
        return 0
```

Then in an individual driver, for example, cisco_ios we do:

```
     @staticmethod
     def _autodetect(self, session, *args, **kwargs):
         cmd = "show version | inc Cisco"
         search_patterns = ["Cisco IOS Software", "Cisco Internetwork Operating System Software"]
         return super(CiscoIosSSH, self)._autodetect(session, 
                                                                                cmd=cmd, 
                                                                                search_patterns=search_patterns)
```

So in the SSHDetect class call of _autodetect (per device_type), the only argument is session.

A specific device driver, could overwrite/change what `_autodetect` did (i.e. not call the parent method) if it so chose. With the only requirement being it accept the `session` argument and return the priority.

The device type specific commands would still be in the driver for that code, but we would minimize the duplicate code (i.e. most of the code would be in base).

maximumG:
@ktbyers : thanks for your review. Your explanation make sense and I agree on the code. Do you want me to work on this or do your prefer to implement the feature yourself ?

ktbyers:
Either way is fine. I can either accept your PR and make modifications to it. 

Or you can modify and we can work from there.

ktbyers:
I am going to merge and then modify this. We can use that as the new base for working on this.

ktbyers:
@maximumG Okay, I converted this to a structure that is similiar to what we did with the snmp_autodetect:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/ssh_autodetect.py

I kept running into problems with trying to use an unbound static method and calling super() on it. I thought it was getting too complicated (even if I would have been able to make it work). And I didn't want a lot of code duplication.

Note, I made a data structure like this:

SSH_MAPPER_BASE = {
    'cisco_ios': {
        "cmd": "show version | inc Cisco",
        "search_patterns": [
           "Cisco IOS Software",
           "Cisco Internetwork Operating System Software",
        ],
        "priority": 99,
    },
    'cisco_nxos': {
        "cmd": "show version | inc Cisco",
        "search_patterns": ["Cisco Nexus Operating System", "NX-OS"],
        "priority": 99,
    },
    'huawei': {
        "cmd": "display version | inc Huawei",
        "search_patterns": ["Huawei"],
        "priority": 99,
    },
}

The 'dispatch' is a function that gets called in the SSHDetect method...so someone can come up with a different method of detection if they wanted.

We need to still to write unit tests for this and expand the platforms.

Let me know if you see issues with this solution.

maximumG:
Thank for the code. If I understand correctly you are bringing back all _autodetect method inside the SSHAutoDetect class. In practice:

- if someone wants to add a new device_type support, **he can use the _autodetect_std method** by adding a new entry in the *SSH_MAPPER_BASE* dict.
- if someone wants to add a ew new device_type support with a all new detection method, **he has to extend the SSHAutoDetect class and add a new method to it specific to its needs**.

Am I right on this ?

I would have preffered the *_autodetect* static method inside each device_type class so that the all methods are well ordered inside their own class. But as you point out this could lead with duplicate code as we are not able to call *super* inside an unbound method... 
**Anyway, your code is doing this same thing, just another way and that's fine for me** 😄 

> We need to still to write unit tests for this and expand the platforms.

I will add to this that we should add some documentation to encourage people to extend this auto-detection. I'm pretty sure that people will warmly welcome this.
Do you have any documentation on how you are making the unit tests currently ? From the code I do not really understand how it is working 😞

ktbyers:
Yes, your understanding is correct. 

I tried to make it work the other way, but it just got too problematic (given the unbound method problems).

I think I probably prefer it the way I did it here (i.e. having the auto-detect code in one place and using bound methods of the SSHDetect class). 

Testing document is here:

Let me know if you have questions on it.

Thanks for your help / work on this.

ktbyers:
FYI, I added some additional platforms. I also added a _results_cache (so the same exact CLI command would only be sent once).

https://github.com/ktbyers/netmiko/commit/c1d8e0e74c75c94db3d5b30eb4759c3d0c967ed6

ktbyers:
Note, I think we also can't have any situation where the search_pattern is exactly in the cli_command that is executed.

For example, 

display version | inc Huawei

and then just search for Huawei

I think we are going to get false positives here as platforms will echo / error the command sent to the screen.

ktbyers:
We probably also need to re-write this to not use the send_command() method and instead use read() and write() methods directly (basically re-create a custom send_command for auto detect purposes).

Wyko:
Cisco WLC:

```
(Cisco Controller) >grep include "Product Name" "show sysinfo"
Press any key to continue..
Product Name..................................... Cisco Controller

There are 1 lines matching the pattern Product Name

(Cisco Controller) >
```

itdependsnetworks:
##### Palo alto 

As far as I know all model's start with PA- something. Only panorama and loggers are different, (I believe)

```
    'paloalto_panos': {
        "cmd":  show system info | match 'model: PA'",
        "search_patterns": [
           "Cmodel: PA",
        ],
        "priority": 99,
```

```
ntc@PAN1> show system info | match 'model: PA'
model: PA-VM
```

Raw 

```
admin@PA-VM> show system info 

hostname: PA-VM
ip-address: 192.168.201.234
netmask: 255.255.252.0
default-gateway: 
ipv6-address: unknown
ipv6-link-local-address: fe80::20c:29ff:fea2:e590/64
ipv6-default-gateway: 
mac-address: 00:0c:29:a2:e5:90
time: Thu Mar 10 21:23:39 2016
uptime: 0 days, 4:34:12
family: vm
model: PA-VM
serial: unknown
vm-mac-base: BA:DB:EE:FB:AD:00
vm-mac-count: 255
vm-uuid: 564D9DE1-5D8A-B67D-D031-5C5B33A2E590
vm-cpuid: 120F6000FFFB8B17
vm-license: none
vm-mode: 
sw-version: 6.1.0
global-protect-client-package-version: 0.0.0
app-version: 451-2337
app-release-date: unknown
av-version: 0
av-release-date: unknown
threat-version: 0
threat-release-date: unknown
wf-private-version: 0
wf-private-release-date: unknown
url-db: paloaltonetworks
wildfire-version: 0
wildfire-release-date: unknown
url-filtering-version: 0000.00.00.000
global-protect-datafile-version: 0
global-protect-datafile-release-date: unknown
logdb-version: 6.1.3
platform-family: vm
vpn-disable-mode: off
multi-vsys: off
operational-mode: normal

```

maximumG:
@ktbyers 

> Note, I think we also can't have any situation where the search_pattern is exactly in the cli_command that is executed.

I agree on that, its a good point. I will rewrite the search pattern tomorrow when I get access to a Huawei device.

> We probably also need to re-write this to not use the send_command() method and instead use read() and write() methods directly (basically re-create a custom send_command for auto detect purposes).

Why can't we use directly the send_command() ?

@Wyko : regarding Cisco WLC, be careful that these devices are using application authentication asking your username and password even when doing a SSH authentication

````
(Cisco Controller)

Username: my-username
Password: ******

(Cisco Controller) >
```

Are you sure that using the SSHAutodetect class connection will work ?

itdependsnetworks:
##### F5

###### 11.x and newer

I beleive sometimes you enter the tmsh shell, and the tmsh is not needed/won't work. Being that command is specfic, anything matching should be fine. 

```
    'bigip_f5': {
        "cmd": "tmsh show sys hardware | grep Name",
        "search_patterns": [
           "Name",
        ],
        "priority": 99,
```

```
[root@BIGIP-F5:Active:Standalone] config # tmsh show sys hardware | grep Name
  Chassis Name
  Name        HD1
  Name        cpus
  Name  BIG-IP Virtual Edition
[root@BIGIP-F5:Active:Standalone]
```

Raw 

 tmsh show sys hardware

```
Sys::Hardware
Chassis Information
  Chassis Name
  Chassis Type
  Maximum MAC Count  1
  Registration Key   -

Hardware Version Information
  Name        HD1
  Type        physical-disk
  Model       VBOX HARDDISK
  Parameters  --                --
              Manufacturer      unavailable
              SerialNumber      VBxxxxxxxx-xxxxxxxx
              Size              100.00G
              Firmware Version  1.0
              Media Type        HDD

  Name        cpus
  Type        base-board
  Model       Intel(R) Core(TM) i5-6260U CPU @ 1.80GHz
  Parameters  --                --
              cache size        6144 KB
              cores             2  (physical:2)
              cpu MHz           1806.966
              cpu sockets       1
              cpu stepping      3


Platform
  Name  BIG-IP Virtual Edition
  BIOS Revision
  Base MAC       08:00:27:42:ff:ff

System Information
  Type                       Z100
  Chassis Serial             xxxxxxxx-xxxx-xxxx-xxxxxxxxxxxx
  Level 200/400 Part
  Switchboard Serial
  Switchboard Part Revision
  Host Board Serial
  Host Board Part Revision



```


###### 10.x and older

Command is very specific to vendor, any match should be fine. Don't have any to test from, took from documentation. 

```
    'bigip_f5': {
        "cmd": "bigpipe platform",
        "search_patterns": [
           "Marketing",
        ],
        "priority": 99,
```

```
PLATFORM INFORMATION --
 Marketing Name: C36
 BIOS Rev: TYAN Tomcat I7210 S5112 12152004 OBJ-0084-00 REV. I
 base MAC: 00:01:D7:2C:B4:00
 Type: C36
 Chassis serial: bip067748s Level 200 part: 200-0138-02 REV E
 Switch board serial: PCA0062MDA33 part: PCA-0062-02 REV G
 Host board serial: TQ45C5000580 part: MOB-0017-02 REV A
 Annunciator board serial: 055673 part: OEM-0029-03 REV A
 Memory: 747.5MB
```

ktbyers:
@maximumG I added some unit tests here:

https://github.com/ktbyers/netmiko/blob/develop/tests/test_netmiko_autodetect.py

and here is an example using it:

https://github.com/ktbyers/netmiko/blob/develop/tests/test_autodetect.sh

maximumG:
@ktbyers  Some comments about the code I tested this morning.

**autodetect() method dispatch key missing**
```python
        for device_type, autodetect_dict in SSH_MAPPER_BASE.items():
            call_method = autodetect_dict.pop("dispatch")
```

If I am running the *autodetect()* method twice, the second time will fail as you are popping the *"dispatch"* key from each the *autodetect_dict*. I know that doing this way is simpler to send the resulting *call_method* dict to the autodetect_method directly. However this might confuse people if they try to autodetect() more than one time. Don't you think we should change this by not popping the key?

**_send_command() method speedup**
```
self.connection.write_channel(cmd + "\n")
time.sleep(1)
```

Is the 1 second sleep required ? The autodetection takes quite a long time (10s) and we could reduce this time 2 ways from my point of view.

- Removing the time.sleep(1) but I'm not sure of the behaviour after that. I add this one for a good reason I guess :smile:
- Changing the way autodetect() works with the accurary. Basically, if the accuracy returned from a specific device_type is 99 we should exit the for loop and stop trying to match other device_type. This way we won't test all the device_type for nothing but this could lead to false positive if some regex are to general

I will create a pull request for the following:

- adding method documentation
- including a new pattern for Huawei
- adding an invalid_response entry
- adding an 'autodetect' device_type in ssh_dispatcher

ktbyers:
@maximumG Why would you call autodetect() twice?

Let's just fix it, by making a copy of the dict, I want to be able to use **kwargs in the method call (which requires the dispatch argument to not be there).

```
114         for device_type, autodetect_dict in SSH_MAPPER_BASE.items():
115             call_method = autodetect_dict.pop("dispatch")
116             autodetect_method = getattr(self, call_method)
117             accuracy = autodetect_method(**autodetect_dict)
```

Yes, we can probably get rid of the time.sleep(1)...note, however, that we need to emphasize reliability over speed so if it unreliable, we might have to add delays back in.



You stated:

```
This way we won't test all the device_type for nothing but this could lead to false positive
if some regex are to general
```

Note, the data structure we are passing in so there is no guarantee that the priority 99 match isn't the last thing done (i.e. we have to try to avoid false positives in the detection code).

That being said, we probably should just return when we get a `priority 99` match (as that should mean we have a definitive match.

maximumG:
> Why would you call autodetect() twice?

I was just testing twice because of an issue with autodetection on Huawei device type. If you think it's no use, just considered it as a feature and not a bug.

> That being said, we probably should just return when we get a priority 99 match (as that should mean we have a definitive match.

That was exactly my point. If the match returns 99 we can be pretty confident that it's the best driver possible and can directly return. We just need to ensure that someone does not add a too global match pattern. 
For example, if someone is adding "Cisco" as a match for device_type 'cisco_ios' with confidence of 99 it will be used even if we are on a 'cisco_asa' and that a false positive...

PR #393, opened 2017-03-06
#387 pull request related code

maximumG:
- Adding some extra documentation
- Adding a new matching regular expression for Huawei
- Adding the autodetect device_type entry based on the TerminalServerSSH class

### DIFF
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -51,7 +51,7 @@ SSH_MAPPER_BASE = {
     },
     'huawei': {
         "cmd": "display version | inc Huawei",
-        "search_patterns": ["Huawei Technologies"],
+        "search_patterns": ["Huawei Technologies", "Huawei Versatile Routing Platform Software"],
         "priority": 99,
         "dispatch": "_autodetect_std",
     },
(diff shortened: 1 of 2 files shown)


ktbyers:
Note, I rolled back the 'Error' pattern as I thought it was too generic (i.e. I though we might have false positives on that).

ktbyers:
@maximumG I also fixed the bug with calling autodetect() twice.

maximumG:
@ktbyers 

> Note, I rolled back the 'Error' pattern as I thought it was too generic (i.e. I though we might have false positives on that).

On Huawei device, I have this pattern to detect Error on command

```
<huawei>foo
             ^
Error: Unrecognized command found at '^' position.
```

On ZTE device, I have this pattern, to detect Error on command

```
zte#foo
foo
  ^
%Error 140303: Invalid input detected at '^' marker.
```

I guess that we could had the following in the error detection, if not already done:

- **'Unrecognized command found'**
- **'Invalid input detected'**

>  I also fixed the bug with calling autodetect() twice.

Thanks for the job :smile:

Do you see anything to add on the code except some unit testing code ?

ktbyers:
@maximumG Okay, let's add something like:

Error: Unrecognized command

maximumG:
> ZTE device? I don't think we support that platform?

Currently working on it on my side...

> Yes, sounds good on submitting a PR to return whenever the first priority 99 is detected.

I will submit a PR as of tomorrow morning (European time).

**I was thinking about a new feature concerning output error** : each device_type class could handle error output and possibly throwing a custom Netmiko Exception when an output error is detected. 

Basically each device type could have a list of predefined regex matching an error output. The *send_command* could have an additional parameter, let say raise_error=True/False (default False) so that error output could raise an error automatically without end user handling it by its own.

This list would be defined as a class variable (and not a class attribute) so that this list could be used by SSHDetect class to match an output error for each device_type without having to instantiate a new class to get access to the list.

Would you be interested in this feature ?

PR #417, opened 2017-03-19
Master

liorfranko:
Add autodetect for hp_comware


ktbyers:
@liorfranko You need to rebase your code to get rid of all these change discrepencies.

See the answer here on how to do this:

http://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository

PR #425, opened 2017-03-26
add extreme_ssh driver specifics

fooelisa:
this fixes a couple of issues with Extreme EXOS 16.2
tested against a VM

### DIFF
--- a/netmiko/extreme/extreme_ssh.py
+++ b/netmiko/extreme/extreme_ssh.py
@@ -1,13 +1,123 @@
 """Extreme support."""
 from __future__ import unicode_literals
+import re
+import time
 from netmiko.cisco_base_connection import CiscoSSHConnection
+
+
+QUESTIONS = {
+    'save': 'Do you want to save configuration to',
+}
 
 
 class ExtremeSSH(CiscoSSHConnection):
     """Extreme support."""
+
     def session_preparation(self):
-        """Extreme requires enable mode to disable paging."""
+        """
+        Extreme requires enable mode to disable paging.
+        """
         self._test_channel_read()
-        self.enable()
         self.set_base_prompt()
         self.disable_paging(command="disable clipaging\n")
+
+    def answer_questions(self, *args, **kwargs):
+        """
+        Send questions/answers with send_command_timing.
+        """
+        if len(args) >= 1:
+            command_string = args[0]
+
+        # handle commands which return a question
+        output = self.send_command_timing(command_string)
+        if QUESTIONS[command_string] in output:
+            output += self.send_command_timing('yes')
+
+        return output
+
+    def send_command(self, *args, **kwargs):
+        """
+        Catch questions from the CLI.
+        Handle prompt changes during command executions.
+        """
+        if len(args) >= 1:
+            command_string = args[0]
+
+        # command that prompt a question
+        if command_string in QUESTIONS.keys():
+            output = self.answer_questions(command_string)
+
+        # all other commands
+        else:
+            # Turn off 'auto_find_prompt' which loses the striped base prompt
+            kwargs.update({'auto_find_prompt': False})
+            # refresh prompt, which could have been changed on previous command
+            self.set_base_prompt()
+            output = super(ExtremeSSH, self).send_command(*args, **kwargs)
+
+        return output
+
+    def send_config_set(self, config_commands=None, exit_config_mode=True, delay_factor=1,
+                        max_loops=150, strip_prompt=False, strip_command=False):
+        """
+        Overwrite send_config_set to acomodate for commands which return a question
+        and expect a reply.
+        """
+        delay_factor = self.select_delay_factor(delay_factor)
+        if config_commands is None:
+            return ''
+        if not hasattr(config_commands, '__iter__'):
+            raise ValueError("Invalid argument passed into send_config_set")
+
+        # Send config commands
+        output = ''
+        for command_string in config_commands:
+
+            # command that prompt a question
+            if command_string in QUESTIONS.keys():
+                output += self.answer_questions(command_string)
+
+            # all other commands
+            else:
+                self.write_channel(self.normalize_cmd(command_string))
+                time.sleep(delay_factor * .5)
+
+        # Gather output
+        output += self._read_channel_timing(delay_factor=delay_factor, max_loops=max_loops)
+
+        output = self._sanitize_output(output)
+        return output
+
+    def set_base_prompt(self, *args, **kwargs):
+        """
+        Extreme attaches an id to the prompt. The id increases with every command.
+        It needs to br stripped off to match the prompt. Eg.
+
+            testhost.1 #
+            testhost.2 #
+            testhost.3 #
+
+        If new config is loaded and not saved yet, a '* ' prefix appears before the
+        prompt, eg.
+
+            * testhost.4 #
+            * testhost.5 #
+        """
+        cur_base_prompt = super(ExtremeSSH, self).set_base_prompt(*args, **kwargs)
+        match = re.search(r'[\*\s]{0,2}(.*)\.\d+', cur_base_prompt)
+        if match:
+            # strip off .\d from base_prompt
+            self.base_prompt = match.group(1)
+            return self.base_prompt
+
+    def config_mode(self, config_command=''):
+        """No configuration mode on Extreme."""
+        return ''
+
+    def check_config_mode(self, check_string='#'):
+        """Checks whether in configuration mode. Returns a boolean."""
+        return super(ExtremeSSH, self).check_config_mode(check_string=check_string)
+
+    def exit_config_mode(self, exit_config=''):
+        """No configuration mode on Extreme."""
+        return ''


netmiko/extreme/extreme_ssh.py
```diff
@@ -1,13 +1,123 @@
...
+
     def session_preparation(self):
-        """Extreme requires enable mode to disable paging."""
+        """
+        Extreme requires enable mode to disable paging.
+        """
         self._test_channel_read()
-        self.enable()
         self.set_base_prompt()
         self.disable_paging(command="disable clipaging\n")
```
ktbyers: I have an earlier note that `enable` was required to disable paging for Extreme. Is this false? Leaving the `enable` call shouldn't hurt anything (if we are unsure).

netmiko/extreme/extreme_ssh.py
```diff
@@ -1,13 +1,123 @@
...
+        """
+        Catch questions from the CLI.
+        Handle prompt changes during command executions.
+        """
+        if len(args) >= 1:
+            command_string = args[0]
+
+        # command that prompt a question
+        if command_string in QUESTIONS.keys():
+            output = self.answer_questions(command_string)
```
ktbyers: I need to make the behavior of `send_command` uniform across a bunch of platforms. So I don't think we should do this.

Instead, I think we might want to just make a `save_config` method (as that is something that is probably worth expanding out to other platforms).

netmiko/extreme/extreme_ssh.py
```diff
@@ -1,13 +1,123 @@
...
+            raise ValueError("Invalid argument passed into send_config_set")
+
+        # Send config commands
+        output = ''
+        for command_string in config_commands:
+
+            # command that prompt a question
+            if command_string in QUESTIONS.keys():
+                output += self.answer_questions(command_string)
+
```
ktbyers: Same issue as above, Netmiko already has solutions for these, and this gets too hard for me to maintain for all of the various device types).

netmiko/extreme/extreme_ssh.py
```diff
@@ -1,13 +1,123 @@
...
+            self.base_prompt = match.group(1)
+            return self.base_prompt
+
+    def config_mode(self, config_command=''):
+        """No configuration mode on Extreme."""
+        return ''
+
+    def check_config_mode(self, check_string='#'):
+        """Checks whether in configuration mode. Returns a boolean."""
+        return super(ExtremeSSH, self).check_config_mode(check_string=check_string)
```
ktbyers: Are we just checking if in `enable_mode` here? If so probably should just call `self.check_enable_mode()`. This will make the code clearer.

netmiko/extreme/extreme_ssh.py
fooelisa: I was wondering where that came from, they don't have any enable, nor configure mode. You drop right into the configuration with an 'admin' account (prefixed with #) or a user account without configuration privileges (prefixed with >). The enable here seems to be sending an empty default anyways, which just results in an extra empty write - I vote for taking it out.

This is with enable:
```
2017-03-27 22:26:36,230 - netmiko - DEBUG - read_channel: ExtremeXOS
Copyright (C) 1996-2016 Extreme Networks. All rights reserved.
This product is protected by one or more US patents listed at http://www.extremenetworks.com/patents along with their foreign counterparts.
==============================================================================

Press the <tab> or '?' key at any time for completions.
Remember to save your configuration changes.


exos-hostname.1 #
2017-03-27 22:26:36,333 - netmiko - DEBUG - read_channel:
2017-03-27 22:26:38,338 - netmiko - DEBUG - read_channel:
2017-03-27 22:26:38,338 - netmiko - DEBUG - read_channel:
2017-03-27 22:26:38,339 - netmiko - DEBUG - write_channel:

2017-03-27 22:26:38,441 - netmiko - DEBUG - read_channel:
exos-hostname.1 #
2017-03-27 22:26:38,546 - netmiko - DEBUG - read_channel:
2017-03-27 22:26:38,547 - netmiko - DEBUG - write_channel:

2017-03-27 22:26:38,548 - netmiko - DEBUG - _read_channel_expect read_data:
exos-hostname.1 #
2017-03-27 22:26:38,652 - netmiko - DEBUG - read_channel:
2017-03-27 22:26:38,653 - netmiko - DEBUG - write_channel: disable clipaging

2017-03-27 22:26:38,654 - netmiko - DEBUG - _read_channel_expect read_data: disable clipaging
exos-hostname.2 #
2017-03-27 22:26:38,956 - netmiko - DEBUG - read_channel:
```

This is the same excerpt without enable (note one less write before the clipaging command): 

```
2017-03-27 22:25:01,218 - netmiko - DEBUG - read_channel: ExtremeXOS
Copyright (C) 1996-2016 Extreme Networks. All rights reserved.
This product is protected by one or more US patents listed at http://www.extremenetworks.com/patents along with their foreign counterparts.
==============================================================================

Press the <tab> or '?' key at any time for completions.
Remember to save your configuration changes.


exos-hostname.1 #
2017-03-27 22:25:01,322 - netmiko - DEBUG - read_channel:
2017-03-27 22:25:03,327 - netmiko - DEBUG - read_channel:
2017-03-27 22:25:03,328 - netmiko - DEBUG - read_channel:
2017-03-27 22:25:03,328 - netmiko - DEBUG - write_channel:

2017-03-27 22:25:03,432 - netmiko - DEBUG - read_channel:
exos-hostname.1 #
2017-03-27 22:25:03,537 - netmiko - DEBUG - read_channel:
2017-03-27 22:25:03,642 - netmiko - DEBUG - read_channel:
2017-03-27 22:25:03,642 - netmiko - DEBUG - write_channel: disable clipaging

2017-03-27 22:25:03,644 - netmiko - DEBUG - _read_channel_expect read_data: disable clipaging
exos-hostname.2 #
2017-03-27 22:25:03,948 - netmiko - DEBUG - read_channel:
```

netmiko/extreme/extreme_ssh.py
fooelisa: I believe there are more cases that will require answers, at least on this platform, so I am not sure if only adding something for the 'save' will be enough. I've seen things like:
```
Warning: Entry exists. Are you sure you want to modify it? (y/N) Yes
```
That's why I initially thought about catching the `(y/N)` part, which could work on other platforms also (with some variation on the exact string).

Let me get back to you on this after some more testing, maybe I can come up with something more generic.

netmiko/extreme/extreme_ssh.py
apepelis: there is no config mode in extreme. I changed my extreme_ssh.py to this and it worked.

"""Extreme support."""
from __future__ import unicode_literals
import string
from netmiko.cisco_base_connection import CiscoSSHConnection


class ExtremeSSH(CiscoSSHConnection):
    """Extreme support."""
    def session_preparation(self):
        """Extreme requires enable mode to disable paging."""
        self._test_channel_read()
        self.set_base_prompt()
        self.disable_paging(command="disable clipaging\n")

    def set_base_prompt(self, pri_prompt_terminator='#', alt_prompt_terminator='>', delay_factor=1):
        super(ExtremeSSH, self).set_base_prompt(pri_prompt_terminator,alt_prompt_terminator, delay_factor)
        if self.base_prompt:
            self.base_prompt = self.base_prompt.strip().rstrip(string.digits)
        return self.base_prompt

apepelis:
the enable check isn't what is my issue. my issue is the wait for prompt. The prompt that is set for the base prompt isn't right since extreme prompts end in an ever increasing number:
* X460G2-48t-10G4.1 # disable clipaging
* X460G2-48t-10G4.2 #
set_base_prompt saves:
*\\ X460G2-48t-10G4.1

but line: 600 in base_connection.py is "output = self.read_until_prompt()"
it sees this: *\\ X460G2-48t-10G4.2
after sending the disable clipaging

the read times out and raises a time out.

ktbyers:
@apepelis Yes, this PR has fix for the increasing number in the prompt.

tknv:
@fooelisa I think pylama error at travis ci that would be fixed by same my way;
https://github.com/ktbyers/netmiko/pull/435/commits/741a8c39e048ea3802672c48c9d5c6a175ab79bd
 i.e. using fix typo or some.
After delete trailing whitespace at 76, 92 then,
Maybe netmiko/base_connection.py:656 `else:   # nobreak` -> `else:   # no break`

ktbyers:
Yes, the pylama error is not important. There are still some questions regarding different Extreme platform (i.e. the behavior of different Extreme devices that we need to resolve).

ktbyers:
Adding some notes about Extreme Software dates EXOS (to try give me some context here).

ExtremeXOS 15.1.1               February 2012
ExtremeXOS 15.3.5               April 2015
ExtremeXOS 16.1                  October 2015
ExtremeXOS 21.1                  March 2016
ExtremeXOS 22.1                  November 2016

ktbyers:
Yes, it doesn't look like there is a `config mode`.

@fooelisa You mentioned the following:

```
You drop right into the configuration with an 'admin' account (prefixed with #) or a user account without configuration privileges (prefixed with >)
```

Is there a way to go from `user account` (i.e. prompt ends in >) to the `admin account`?

Does Extreme even have a command for `enable`? I didn't see it looking through the command reference for EXOS?

ktbyers:
I am going to merge this and then reduce it down to a simpler solution (we can then add upon that, if necessary).

I think the key thing to address (right now is the increasing numbers in the prompt i.e.)

```
X460G2-48t-10G4.1 # disable clipaging
X460G2-48t-10G4.2 #
```

ktbyers:
See updated state:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/extreme/extreme_ssh.py


I removed the questions/answers code as I don't think I want this.

I am open to more discussions on this, but it is a common problem across a lot of platforms and there are existing solutions for it in Netmiko.

I didn't remove the enable() in `session_preparation` as I still had questions about whether you can disable paging when in a `user account` i.e. when prompt ends in `>`. Also if you already have a prompt of `#` the enable method won't send any commands (so it shouldn't hurt anything).

That being said...I am not sure the `enable` command actually exists on EXOS?

ktbyers:
It would also be good if someone could run the unit tests against an Extreme Device.

fooelisa:
Thanks for merging in the first part, that was fast! Let me do some more testing and see if I can come up with a better solution for my other changes.

> Is there a way to go from user account (i.e. prompt ends in >) to the admin account?

Nope, there is no enable command on Extreme.

ktbyers:
@fooelisa Thanks for the comment on `enable`. So we just need to comment that call out in session_preparation method.

fooelisa:
Hey, it's been a while. I have need testing this against real devices for some time now and all seems well, except for one issue...

Whenever I apply config that changes the hostname of the device, it won't recognize the prompt after. I believe you mentioned there are existing functions to deal with something like that @ktbyers - could you point me to them, as I would like to add that as well.

Thanks!

ktbyers:
@fooelisa Can you show me the exception you get after you make this change? I want to make sure I understand where it is failing at.

fooelisa:
Sure, I am applying a config that only merges in a new hostname, on extreme eg.

```
--- running_config.conf
+++ candidate_config.conf
@@ -100,0 +101 @@
+configure snmp sysName test-merge-file-2
```

This results in:

```
  File "./test_pyexos.py", line 40, in <module>
    device.commit_config()
  File "build/bdist.linux-x86_64/egg/pyEXOS/exos.py", line 131, in commit_config
  File "build/bdist.linux-x86_64/egg/netmiko/extreme/extreme_ssh.py", line 52, in send_command
  File "build/bdist.linux-x86_64/egg/netmiko/base_connection.py", line 811, in send_command
IOError: Search pattern never detected in send_command_expect: test\-merge\-file\-2
```

If I log into the device after the timeout, I do see the hostname changed, so it did apply the send command, however never returned from it. I was intuitively attributing it to the new name, when I apply a config that performs other changes on the device it all works as expected.

Thanks!

ktbyers:
@fooelisa You can use `send_command_timing` instead of `send_command`. `send_command_timing` will be entirely time based and not pattern based (so it won't care that the hostname changed).

You can see an example of this here:

https://github.com/ktbyers/netmiko/blob/develop/COMMON_ISSUES.md#show-commands-that-prompt-for-more-information

fooelisa:
Ok, great, this seems to work!
Is there any drawback to this method? How long will it wait for the command to finish?

Issue #430, opened 2017-04-01
read_until_prompt_or_pattern does not apply timeout or sleep


markgoddard:
The read_until_prompt_or_pattern method in base_connection.BaseConnection reads data from the connection in a tight loop reading data from the device, waiting for the prompt or another pattern to be received. There are a couple of issues with this.

1. There is no way to break out of the loop - no timeout mechanism is applied, so we can get stuck here indefinitely if the device does not send the expected pattern or prompt.
2. There is no delay between successive calls to the read_channel method. This can lead to excessive CPU use as the process spins. It also has a negative impact when used in tandem with cooperative multitasking libraries such as eventlet (as used by OpenStack, in which the networking-generic-switch package uses netmiko and where I observed this issue).

The _read_channel_expect method used by read_until_prompt and read_until_pattern avoids these issues by applying a timeout to the read and sleeping between each read.

Just to fill in the details, I saw this in the following environment:
- CentOS 7.3 host.
- OpenStack Neutron Ocata release, running in a CentOS 7.3 Docker container.
- OpenStack Neutron driver: networking-generic-switch-0.3.0
- netmiko-1.3.0
- Dell S6010-ON switches, Dell Network OS 9.10(0.1)

markgoddard:
Related issue in OpenStack networking-generic-switch: https://bugs.launchpad.net/networking-generic-switch/+bug/1678461.

ktbyers:
@markgoddard I will try to respond to this...probably on Wednesday.

ktbyers:
This was fixed here:

https://github.com/ktbyers/netmiko/pull/431

PR #431, opened 2017-04-01
Issue #430: read_until_prompt_or_pattern does not apply timeout or sleep

markgoddard:
The read_until_prompt_or_pattern method in base_connection.BaseConnection reads
data from the connection in a tight loop reading data from the device, waiting
for the prompt or another pattern to be received. There are a couple of issues
with this.

1. There is no way to break out of the loop - no timeout mechanism is applied,
so we can get stuck here indefinitely if the device does not send the expected
pattern or prompt.

2. There is no delay between successive calls to the read_channel method. This
can lead to excessive CPU use as the process spins.  It also has a negative
impact when used in tandem with cooperative multitasking libraries such as
eventlet (as used by OpenStack, in which the networking-generic-switch package
uses netmiko and where I observed this issue).

The _read_channel_expect method used by read_until_prompt and
read_until_pattern avoids these issues by applying a timeout to the read and
sleeping between each read.

This change resolves the issue by combining the prompt and pattern into a
combined regular expression and waiting for it to appear using
_read_channel_expect.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -320,18 +320,10 @@ class BaseConnection(object):
 
     def read_until_prompt_or_pattern(self, pattern='', re_flags=0):
         """Read until either self.base_prompt or pattern is detected. Return ALL data available."""
-        output = ''
-        if not pattern:
-            pattern = re.escape(self.base_prompt)
-        base_prompt_pattern = re.escape(self.base_prompt)
-        while True:
-            try:
-                output += self.read_channel()
-                if re.search(pattern, output, flags=re_flags) or re.search(base_prompt_pattern,
-                                                                           output, flags=re_flags):
-                    return output
-            except socket.timeout:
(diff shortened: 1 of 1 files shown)


ktbyers:
Why isn't the socket.timeout happening (it should happen)?

What is the read_channel() call in this function returning? We might need to log this to see the data (i.e. turn on Netmiko logging)?

markgoddard:
There is no socket.timeout because read_channel checks to see if any data is available on the socket and does not do a read if there is none: https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L216.

This is inconsistent with read_channel_expect which does a blocking read (but should arguably use read_channel).

ktbyers:
I reviewed use of this code and this change looked good and corrected the issue pointed out.

markgoddard:
Thanks Kirk!

PR #434, opened 2017-04-04
conditinal check for fortinet ssh key exchange

itdependsnetworks:
Fortinet has some odd ssh behavior, described: http://napalm.readthedocs.io/en/latest/support/fortios.html

This should conditionally address the issue.


ktbyers:
@itdependsnetworks I incorporated this change, but moved it into the fortinet specific code. 

See https://github.com/ktbyers/netmiko/commit/a815a63b1d3e589f0383b98cebe07aeebb62b320

PR #440, opened 2017-04-07
Compatibility with Huawei USG devices running software version 5

Skrevin:
This code adds support for new devices: the Huawei USG series (tested on version 5). The code in huawei_usgv5_ssh.py is the same as in huawei_ssh.py, with the exception of an if/else statement that will check in which mode the prompt is situated (high availability or not).

I tested it with following code:

`from netmiko import ConnectHandler

net_connect = ConnectHandler(device_type='huawei_usgv5', ip='ip address', username='user', password='password')
net_connect.set_base_prompt()
config_commands=["interface GigabitEthernet 0/0/4", "ip add 10.0.0.60 24", "interface GigabitEthernet 0/0/5", "ip add 10.0.1.61 24"]
output = net_connect.send_config_set(config_commands)
`
=> this works (when device is in high availability, as well as when it isn't)
output:
system-view
Enter system view, return user view with Ctrl+Z.
HRP_M[FW1]interface GigabitEthernet 0/0/4 (+B)
HRP_M[FW1-GigabitEthernet0/0/4]ip add 10.0.0.60 24
HRP_M[FW1-GigabitEthernet0/0/4]interface GigabitEthernet 0/0/5 (+B)
HRP_M[FW1-GigabitEthernet0/0/5]ip add 10.0.1.61 24
HRP_M[FW1-GigabitEthernet0/0/5]return
HRP_M[FW1]

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -73,7 +73,7 @@ class BaseConnection(object):
         :type ssh_strict: bool
         :param system_host_keys: Load host keys from the user's 'known_hosts' file.
         :type system_host_keys: bool
-        :param alt_host_keys: If `True` host keys will be loaded from the file specified in 
+        :param alt_host_keys: If `True` host keys will be loaded from the file specified in
                 'alt_key_file'.
         :type alt_host_keys: bool
         :param alt_key_file: SSH host key file to use (if alt_host_keys=True).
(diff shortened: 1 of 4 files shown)


ktbyers:
@Skrevin I made this change here:

https://github.com/ktbyers/netmiko/commit/ee23c32e75bc04284ec71652579e09b2516bdfb8#diff-351c60dfea17cc8cdb1eafc438c33323R69

Which will strip off any leading HRP_X where X can be any character from the prompt.

That looked like it would work to me (let me know if you see a reason it won't work).

PR #446, opened 2017-04-10
Cisco XR code changes

nkaliape:
1) Cisco IOS XR - OO refactoring
2) Tested for telnet (vty), ssh and console operations
3) Added verbose for concurrent logging
4) Debug flag handling 
5) timer fix - simplified with single keyword "max_timeout" which is backward compatible with max_loops and loop_delay
6) Small failure differentiation with different error msg
7) Handling Standby prompt while connection two console session


nkaliape:
Will fix the issues and get back

PR #447, opened 2017-04-13
Add Check Point Gaia support

alexanderhofmann:
This commit adds support for Check Point Gaia Firewalls.


ktbyers:
Nice...

Any possibility you can run this against the standard unit tests (see https://github.com/ktbyers/netmiko/blob/develop/TESTING.md).

Let me know if you need any assistance on this.

PR #449, opened 2017-04-19
retry of pull #446 after pylama changes 

nkaliape:


nkaliape:
closing again since pylama fails

PR #476, opened 2017-05-24
Adding keepalive option

mirceaulinic:
Mentioned under https://github.com/napalm-automation/napalm-ios/pull/149.
These changes wouldn't affect the existing behaviour, and it enables the keepalive only when explicitly set by the user.
Please let me know what you think.


ktbyers:
This looks fine...only issue I see is you didn't default it to a value of 0. I will update that.

ktbyers:
Here is the update of the default value:

https://github.com/ktbyers/netmiko/commit/42e32092d46ce772d7baf8e0a1b8db5382f89a21

mirceaulinic:
Oops, it was supposed to be 0. While testing I set it to 60 and forgot to revert. Thanks!

PR #478, opened 2017-05-25
added checks for EOF

### LINKED ISSUE #477: No EOF checks after calling `self.remote_conn.recv()`
Both `_read_channel()` and `_read_channel_expect()` do not check for EOF after calling `self.remote_conn.recv()`.  This causes a connection which has closed to raise

`NetMikoTimeoutException("Timed-out reading channel, pattern not found in output: {}".format(pattern))`

This isn't exactly desired behavior as it makes it hard to discern whether the command just took too long or the connection is dead.  Most of the time, a subsequent call to `self.send_command()` will cause paramiko to raise `socket.error("Socket is closed")`, but I have ran into a case where the socket only half closed, so the `socket.error` never got raised.

I can fix this, but I'd like guidance on the preferred way to signal this error to the user.  `EOFError`?  A netmiko exception?  Something else?

kbirkeland:
Fixes #477

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -226,7 +226,10 @@ class BaseConnection(object):
             output = ""
             while True:
                 if self.remote_conn.recv_ready():
-                    output += self.remote_conn.recv(MAX_BUFFER).decode('utf-8', 'ignore')
+                    outbuf = self.remote_conn.recv(MAX_BUFFER)
+                    if len(outbuf) == 0:
+                        raise EOFError
+                    output += outbuf.decode('utf-8', 'ignore')
                 else:
                     break
         elif self.protocol == 'telnet':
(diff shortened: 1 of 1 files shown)


ktbyers:
In this context the SSH socket would be still open (i.e. there is still a TCP session open between the two devices), but the SSH channel has closed at least one way from network device to server running Netmiko.

We will raise a specific error to notify things like Salt that they need to restart the SSH session (probably implies restarting a completely new SSH connection as Netmiko doesn't have any way of restarting a new channel and good chance most network devices would fail if we tried this anyways).

So then raises the question if this happens of whether we should gracefully close the SSH session (i.e. disconnect() ). Which I would think would be yes.

kbirkeland:
Yeah, it just signals that the read direction of the channel is closed.  You're correct for napalm/salt - I just called self.open() again (see napalm-automation/napalm-ios#152).

I don't think a notification needs to be made on a graceful shutdown, since that should be intended.  It's only 'unexpected' that the channel is closed when calling `read_channel` or `write_channel`.  

Even if an automatic recovery worked for most devices, it may silently put an application in a state that it did not expect (e.g. was in config mode but reconnects into exec).  Raising an exception should allow the application to do whatever is necessary.

PR #485, opened 2017-06-03
fixed problem which was produced by last symbol of prompt line

nikolaybykovsky:
for palo alto device command usual prompt ends with '>' while in configuration mode with '#'.
This differences produces the problem when in exit_config_mode method because it looking for a prompt which ends with '#' while real prompt ends with '>'

### DIFF
--- a/netmiko/cisco/__init__.py
+++ b/netmiko/cisco/__init__.py
@@ -6,6 +6,7 @@ from netmiko.cisco.cisco_xr_ssh import CiscoXrSSH
 from netmiko.cisco.cisco_wlc_ssh import CiscoWlcSSH
 from netmiko.cisco.cisco_s300 import CiscoS300SSH
 from netmiko.cisco.cisco_tp_tcce import CiscoTpTcCeSSH
+from netmiko.cisco.cisco_fmc_ssh import CiscoFmcSSH
 
 __all__ = ['CiscoIosSSH', 'CiscoIosTelnet', 'CiscoAsaSSH', 'CiscoNxosSSH', 'CiscoXrSSH',
-           'CiscoWlcSSH', 'CiscoS300SSH', 'CiscoTpTcCeSSH', 'CiscoIosBase']
+           'CiscoWlcSSH', 'CiscoS300SSH', 'CiscoTpTcCeSSH', 'CiscoIosBase', 'CiscoFmcSSH']
--- a/netmiko/cisco/cisco_fmc_ssh.py
+++ b/netmiko/cisco/cisco_fmc_ssh.py
@@ -1,0 +1,14 @@
+"""Subclass specific to Cisco ASA."""
+
+from __future__ import unicode_literals
+import re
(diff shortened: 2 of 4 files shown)


ktbyers:
@nikolaybykovsky 

Can you show me the original error that you received (prior to making changes)?

Looking at the current code:

```
    def exit_config_mode(self, exit_config='exit'):
        """Exit configuration mode."""
        output = ""
        if self.check_config_mode():
            output = self.send_command(exit_config, strip_prompt=False, strip_command=False)
            if self.check_config_mode():
                raise ValueError("Failed to exit configuration mode")
        return output
```

and then check_config_mode is:

```
    def check_config_mode(self, check_string=']'):
        """Checks if the device is in configuration mode or not."""
        return super(PaloAltoPanosSSH, self).check_config_mode(check_string=check_string)
```

So check_config_mode will send an enter. This is in the parent class. It should then return True (as it will see the ']' in config mode). The 'exit' command will then be sent which should cause it to exit from config mode.

It looks like 'quit' might be the more canonical way to exit, however. Here is an example PanOs from the PanOs docs:

```
username@hostname>  configure
Entering configuration mode
[edit]
username@hostname#


username@hostname# quit
Exiting configuration mode
username@hostname>
```

nikolaybykovsky:
@ktbyers 
The original error is the following:
`Traceback (most recent call last):
  File "cli_api_test.py", line 27, in <module>
    print result.result()
  File "/home/proj/venv/local/lib/python2.7/site-packages/concurrent/futures/_base.py", line 422, in result
    return self.__get_result()
  File "/home/proj/venv/local/lib/python2.7/site-packages/concurrent/futures/thread.py", line 62, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/proj/fwaccess/palo_alto_cli.py", line 120, in fw_set_mgtconfig_users
    self.exit_config_mode()
  File "/home/netmiko/netmiko/paloalto/paloalto_panos_ssh.py", line 55, in exit_config_mode
    output = self.send_command(exit_config, strip_prompt=False, strip_command=False)
  File "/home/netmiko/netmiko/paloalto/paloalto_panos_ssh.py", line 161, in send_command
    return super(PaloAltoPanosSSH, self).send_command(*args, **kwargs)
  File "/home/netmiko/netmiko/base_connection.py", line 815, in send_command
    search_pattern))
IOError: Search pattern never detected in send_command_expect: admin\@w1\_paloalto\_267nb\#
`
It happened because it was trying to find "admin@w1_paloalto_267nb#" pattern, pattern "admin@w1_paloalto_267nb>" was presented

ktbyers:
Okay, thanks for the output. On quick glance, I think there is probably an easier fix. I will try to submit this and update you when it is done.

@nikolaybykovsky

ktbyers:
@nikolaybykovsky Sorry for delay on responding on this:

I fixed the exit_config_mode issue here:

PR #487, opened 2017-06-07
Added argument to send_command() method to prevent normalising the command_string

mjuenema:
I added an optional  `normalize_command` argument to the `send_command()` method. The motivation was to be able to send "commands" to the device that are not NEWLINE terminated. For example, with Cisco IOS pressing the ? key will invoke the help feature.

Example in Python
```python
net_connect = ConnectHandler(...)
net_connect.enable()
net_connect.send_command("conf t", expect_string="\(config\)#")

# Send "no cdp ?"
output = net_connect.send_command("no cdp ?", expect_string="\(config\)# no cdp",
                                   normalize_command=False)
```

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -742,7 +742,7 @@ class BaseConnection(object):
 
     def send_command(self, command_string, expect_string=None,
                      delay_factor=1, max_loops=500, auto_find_prompt=True,
-                     strip_prompt=True, strip_command=True):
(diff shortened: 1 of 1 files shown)


ktbyers:
FYI, I am going to change this to `normalize` for an argument name.

ktbyers:
I added this to `send_command_timing` also:

https://github.com/ktbyers/netmiko/commit/0f7a669e3ea1d82e0ebd28b13e4a4f9e7979967f

PR #497, opened 2017-06-16
Multi process queue deadlock

bacada105:
With large switch counts the mp_queue fills up resulting in a deadlock.


ktbyers:
@bacada105 Why did you close this...do you want this re-opened?

We should fix this...just was traveling a lot the last couple of weeks.

bacada105:
@ktbyers  Oops i'm new to GitHub and still trying to figure out the work flow of everything. I'll see if I can correct this.

PR #506, opened 2017-06-23
add support for accedian

bangert:


ktbyers:
@bangert Can you run the show and config unit tests against a test device and post the results here?

Thanks.

bangert:
```
[bangert@stuart tests]$ py.test -v test_netmiko_show.py --test_device accedian
=================================== test session starts ===================================
platform linux2 -- Python 2.7.13, pytest-2.9.2, py-1.4.31, pluggy-0.3.1 -- /usr/bin/python
cachedir: ../.cache
rootdir: /home/bangert/src/gc/cli-tools/accedian/edu5/netmiko, inifile: 
plugins: pylama-7.3.3
collected 12 items 

test_netmiko_show.py::test_disable_paging PASSED
test_netmiko_show.py::test_ssh_connect PASSED
test_netmiko_show.py::test_ssh_connect_cm PASSED
test_netmiko_show.py::test_send_command_timing PASSED
test_netmiko_show.py::test_send_command_expect PASSED
test_netmiko_show.py::test_base_prompt PASSED
test_netmiko_show.py::test_strip_prompt PASSED
test_netmiko_show.py::test_strip_command PASSED
test_netmiko_show.py::test_normalize_linefeeds PASSED
test_netmiko_show.py::test_clear_buffer PASSED
test_netmiko_show.py::test_enable_mode PASSED
test_netmiko_show.py::test_disconnect PASSED

=============================== 12 passed in 38.42 seconds ================================
```

the config test fails due to missing support for enable and config mode...

ktbyers:
Do the `send_config_set` and `send_config_from_file` methods work? If not, we should make these two methods work as they are the standard way in Netmiko to execute config changes.

bangert:
`send_config_set()` requires config mode, which this device does not support, so it does not work.

i would propose to make send_config_set() not require config mode.

ktbyers:
@bangert Looking at the current code:

```
    def send_config_set(self, config_commands=None, exit_config_mode=True, delay_factor=1,
                        max_loops=150, strip_prompt=False, strip_command=False):
        """  
        Send configuration commands down the SSH channel.

        config_commands is an iterable containing all of the configuration commands.
        The commands will be executed one after the other.

        Automatically exits/enters configuration mode.
        """
        debug = False
        delay_factor = self.select_delay_factor(delay_factor)
        if config_commands is None:
            return ''
        elif isinstance(config_commands, string_types):
            config_commands = (config_commands,)

        if not hasattr(config_commands, '__iter__'):
            raise ValueError("Invalid argument passed into send_config_set")

        # Send config commands
        output = self.config_mode()
        for cmd in config_commands:
            self.write_channel(self.normalize_cmd(cmd))
            time.sleep(delay_factor * .5)

        # Gather output
        output += self._read_channel_timing(delay_factor=delay_factor, max_loops=max_loops)
        if exit_config_mode:
            output += self.exit_config_mode()
        output = self._sanitize_output(output)
        if debug:
            print(output)
        return output
```

This would work as long as we had `config_mode`, `check_config_mode`, and `exit_config_mode` do nothing (i.e. return ''...and probably return False for check_config_mode since it generally returns a boolean).

We could then call using super() the `send_config_set` method with the argument `exit_config_mode=False`.

This would allow us to avoid duplicating the `send_config_set` code.

ktbyers:
@bangert Okay, I made changes to fix the send_config_set and send_config_from_file methods here.

https://github.com/ktbyers/netmiko/commit/e8b9cdcf861f311dc2c315546c53b071b1bb35df

I think it should work now on the current develop branch. It would be great if you could test `send_config_set` and `send_config_from_file` methods.

Also what Accedian product(s) is this for (i.e. which product family does this support)

bangert:
great - i'll check this once i am back from holiday and have access to the device (~3 weeks) :-)

ktbyers:
@bangert Sounds good. No worries on the delay...test it when it works for your schedule.

bangert:
this works great!

i only tested the Accedian TE series - which is also sold as Transmode/Infinera TE.

ktbyers:
@bangert Sounds good. Thanks for the follow up testing.

PR #516, opened 2017-07-06
Brocade Fastiron and Netiron Telnet Support

nmusser:
PR #235 only added support for SSH. I've gone ahead and added Telnet support for Fastiron and Netiron. 

Tested against: 
Brocade FCX648S-HPOE, IronWare Version 07.4.00dT7f1
Brocade ICX6610-48-HPOE, IronWare Version 07.4.00dT7f1
Brocade CER, IronWare Version V5.6.0jT183
Brocade MLXe, IronWare Version V5.7.0eT163

I had to override several methods because when connected via Telnet both Fastiron and Netiron expect a combination of '\r\n' to be sent at the end of a line. So every method that calls write_channel('\n') had to be overridden to send '\r\n'. Might patch base_connection in the future to store the newline character in a variable to allow overriding without the duplication of all the functions. 

Also had to override the 3 config related methods (check_config_mode, config_mode, and exit_config_mode) to set the expected pattern to the equivalent of the prompt (I was inheriting from Cisco Base with truncates this because IOS truncates it).


ktbyers:
Yes, let's patch base_connection.py insteado of duplicating all of those methods (i.e. add an argument to the method for `newline_format`).

Probably `def normalize_cmd` method and `def normalize_linefeeds` methods (and maybe a few other minor places).

Does it matter if IOS truncates the trailing prompt that it is searching for? Or asked another way why is this behavior different from the Brocade FastIron/NetIron SSH driver?

nmusser:
As you requested I patched base_connection.py to add an argument for `newline_format` for all the methods requiring it and then overrode those methods for the fastiron and netiron telnet drivers. The methods that I added `newline_format` to  are:
```
    def _test_channel_read(self, count=40, pattern="", newline_format="\r"):
    def find_prompt(self, delay_factor=1, newline_format="\r"):
    def normalize_cmd(command, newline_format="\r"):
    def check_enable_mode(self, check_string='', newline_format="\r"):
    def check_config_mode(self, check_string=')#', pattern='', newline_format="\r"):
```
My code before was really hacked together and sloppy. I took more time testing this time and doing it properly (as you can see I didn't make a million commits to be PEP8 compliant like before). I discovered that you don't have to send a `\r\n` to the Brocade devices. You only have to send a `\r` so that's all I did. Also, I **swear** I had a problem with the IOS truncating behavior but I could not reproduce it so I don't know what I was thinking. I've removed that and everything seems to work fine.

ktbyers:
@nmusser This was merged in here:

https://github.com/ktbyers/netmiko/tree/netmiko_dev_1_5

Note, in Netmiko 1.5 I abstracted away the 'enter' key so I restructured your code to reflect this. It would be helpful if you could test the updated code.

Thanks.

Issue #496, opened 2017-06-16
device type = cisco_ios_telnet - reverse telnet


jdimstrnate:
I am newer to python and have read the documentation but can't identify how to make netmiko enter credentials more than once when prompts are seen.  

trying to telnet to a Terminal Server to validate console ports via script.  
ranging from telnet ports 2051-2063   Only trying one port at the moment.
Error shows **actively refused** however I tested connection from same box using telnet and it works.  

====================================================

import netmiko
import json
device = input('Please input device IP:  ')
device_type = 'cisco_ios'
user = 'uuuuuuuu'
pw = 'xxxxxxxx!'

SSH = netmiko.ConnectHandler(ip='10.232.7.194', device_type='cisco_ios_telnet', username='uuuuuuuu',password='xxxxxxxx!')
SSH.read_channel()

print('########################################################################################')
print('Connecting to port: ', ports2051 , 'on device: ', device)
print('########################################################################################')
SSH.send_command_expect('username:',user)
SSH.send_command_expect('password:',pw)
SSH.send_command('\n')
SSH.send_command_expect('username:',user)
SSH.send_command_expect('password:',pw)
print(SSH.send_command('show ip int br'))

========================================================
ERROR SHOWN BELOW
========================================================
Traceback (most recent call last):
  File "C:/Users/v-nademe.REDMOND/Netmiko_Training/__init__.py", line 40, in <module>
    SSH = netmiko.ConnectHandler(ip='10.232.7.194', device_type='cisco_ios_telnet', username='UUUUUUUU',password='XXXXXXX!')
  File "C:\Python36-32\lib\site-packages\netmiko\ssh_dispatcher.py", line 124, in ConnectHandler
    return ConnectionClass(*args, **kwargs)
  File "C:\Python36-32\lib\site-packages\netmiko\base_connection.py", line 128, in __init__
    self.establish_connection()
  File "C:\Python36-32\lib\site-packages\netmiko\base_connection.py", line 494, in establish_connection
    self.remote_conn = telnetlib.Telnet(self.host, port=self.port, timeout=self.timeout)
  File "C:\Python36-32\lib\telnetlib.py", line 218, in __init__
    self.open(host, port, timeout)
  File "C:\Python36-32\lib\telnetlib.py", line 234, in open
    self.sock = socket.create_connection((host, port), timeout)
  File "C:\Python36-32\lib\socket.py", line 722, in create_connection
    raise err
[...]

PR #524, opened 2017-07-18
Expanding telnet_login username_pattern to add "ser:" and "ogin"

laurentlgm:
As discussed here: https://github.com/ktbyers/netmiko/issues/496
I've added "ser:" and "ogin" to the existing "sername" under Telnet_login so it works with reverse telnet and different prompts.
I thought that "ser" would be too short and generic so added the ":" just like it pops up on the Quanta LB9 running Broadcom ICOS/Fastpath.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -345,7 +345,7 @@ class BaseConnection(object):
         return self._read_channel_expect(combined_pattern, re_flags=re_flags)
 
     def telnet_login(self, pri_prompt_terminator='#', alt_prompt_terminator='>',
-                     username_pattern=r"sername", pwd_pattern=r"assword",
+                     username_pattern=r"ser:|sername|ogin", pwd_pattern=r"assword",
                      delay_factor=1, max_loops=60):
         """Telnet login. Can be username/password or just password."""
         TELNET_RETURN = '\r\n'


ktbyers:
Okay, interesting let me look at it.

Can you just write a one paragraph description of the problem you are solving?

laurentlgm:
Whenever going through a terminal server with reverse telnet to log into the console port of an end device which, has also been configured to ask for authentication on that port, we will have to go through 2 authentication interactions:
The first user/pass prompt comes from the terminal server and the second user/pass prompt comes from the end device we are trying to get access to.

As it is today in 1.4.2, netmiko with “generic_termserver_telnet” will call the “telnet_login” which inserts the same username and password every time “read_channel” comes back with text that matches “username_pattern” and “pwd_pattern”.
The login method will not return a success response until a device prompt terminated by either “>” or “#” is received. 

The issue is that not every OS/vendor asks for [uU]sername by default.
If, for example, the terminal server is an ios device but the target device is a junos, then netmiko will go through the first authentication prompt (because cisco will ask for “Username:”) but will fail to login to the Juniper device since it will send back a “login:” a prompt. The “telnet_login” method will be stuck in a loop for the default “max_loops=60” and finally return a “Telnet login failed:” message even though it had successfully passed the terminal server authentication.

Example:
Laptop --GigEth--> cisco 2901 acting as terminal server --Serial RS232--> Juniper QFX5100 target device
Let’s say the QFX5100 is connected to the first Async port on the 2901: port 2002
To get access to the QFX5100 we use the 2901 IP address visible to the laptop: 10.0.0.1
So, to get access to the QFX5100 console port we issue: telnet 10.0.0.1 2002
First prompt will be the cisco with “Username:/Password:” netmiko will pass this and continue to look for >/#
Second prompt will be the Juniper with “login:/password:” netmiko will fail this because login does not match the username_pattern =r“sername”

With this pull request I’m proposing we extend the possibilities of the username_pattern to  include “[uU]ser:” and “[lL]ogin” which should catch a most scenarios.
Tested successfully in a couple scenarios since yesterday with different devices: Quanta Lb9 Fastpath, QFX5100 junos, c2901 ios, Accton with cumulus

Where I see this breaking: 
Some terminal servers like the Avocent models get you a message like this after the first authentication:
“Type the hot key to suspend the connection: <CTRL>Z”
Because the of the “>” in the phrase above netmiko thinks it’s a router prompt and start sending the show commands before it actually logs into the end device.
To fix this we, would have to change the pri_prompt_terminator='#', alt_prompt_terminator='>' to a regex search for r’>$|#$’.

ktbyers:
Merged here with slight modifications:

https://github.com/ktbyers/netmiko/commit/785c647a8a1546fd2ab70fac97df62f96c959132

PR #526, opened 2017-07-19
Changing prompt_pattern on Telnet_login to avoid false positives and max_loops to 10

laurentlgm:
To fix the problem with terminal servers using ">" outside the real prompt as discussed here:
https://github.com/ktbyers/netmiko/pull/524
Also reducing max_loops from 60 to 10 on the login just as a suggestion to speed-up failures and to avoid blocking ports that record multiple failed login attempts in a short period of time.
Tested with Avocent, cisco, and juniper.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -344,9 +344,9 @@ class BaseConnection(object):
             combined_pattern += "|{}".format(pattern)
         return self._read_channel_expect(combined_pattern, re_flags=re_flags)
 
-    def telnet_login(self, pri_prompt_terminator='#', alt_prompt_terminator='>',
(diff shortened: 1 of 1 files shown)


ktbyers:
@laurentlgm This was incorporated here, but I didn't merge just made similar changes.

https://github.com/ktbyers/netmiko/commit/95244286056fe93aed06173fbb3cf70625f5bfc4

I retained the separate distinct password prompts as it is more consistent to other Netmiko calls and I didn't want to make a backwards incompatible change. 

Also I added `flags=re.M` to the re.search which increases the likelihood of false positives in certain situations, but lowers failures in cases where this output data which follows the prompt (which does happen although probably rarely).

In the terminal server case, you could completely overwrite the call so you could change the search patterns to be what you want or completely create your own login handling.

PR #567, opened 2017-08-31
Shreyash netmiko

shreyashah:


ktbyers:
Thanks for submitting the PR.

Can you provide a high-level overview of what you are trying to fix in the PR? What issues you are trying to resolve?

Note, I am very unlikely to accept this pull request due to the high number of changes in base_connection.py, high number of changes in cisco_base_connection.py, and all the linting failures (including a syntax error).

When submitting a PR it would greatly increase the chance of me accepting it...if you limit the PR to fixing one specific issue: for example adding IOS-XR telnet support or fixing some aspect of IOS-XR SSH support (these should not be coupled into the same PR). 

Additionally, minimize changes in the base_connection.py and cisco_base_connection.py (minimize changes here as changes here have a much higher likelihood of breaking other platforms).

I am happy to discuss with you the things that are not working on IOS-XR, however, as I know some changed need to be made here.

ktbyers:
Closing, no context for change.

PR #570, opened 2017-09-05
add coriant driver

bangert:
this adds support for coriant groove g30 hardware


ktbyers:
@bangert Thanks for PR. Probably will be next week before I get to this (hopefully).

bangert:
NP - thanks for netmiko!

ktbyers:
@bangert Is there any paging of output with this device?

i.e. does it prompt you for

```
--more--
```

on output.

PR #573, opened 2017-09-12
F5 disable paging cli command depends on version

zhecho:
Hello all,

I think that session_preparation() should send various "disable paging CLI commands" depends on the F5 firmware version 
I couldn't find command "set length 0" as previous cli command was set... but as we spoke I decided to leave it as well , so for version 11,12,13 it'll execute "modify cli preference pager disabled". I think that this command will work on ver.10.x.x as well but I couldn't find working v.10 for test it 

Best Regards!

### DIFF
--- a/netmiko/f5/f5_ltm_ssh.py
+++ b/netmiko/f5/f5_ltm_ssh.py
@@ -13,14 +13,10 @@ class F5LtmSSH(BaseConnection):
         delay_factor = self.select_delay_factor(delay_factor=0)
         self._test_channel_read()
         self.set_base_prompt()
-        command = self.dis_paging_cmd()
-        self.disable_paging(command=command)
+        # Send disable paging cmd depends on F5 cli version
+        cmd = self.dis_paging_cmd()
+        self.disable_paging(command=cmd)
         time.sleep(1 * delay_factor)
-        self.tmsh_mode()
-        self.set_base_prompt()
-        # Clear the read buffer
-        time.sleep(.3 * self.global_delay_factor)
-        self.clear_buffer()
 
     def dis_paging_cmd(self):
         """ return F5 disable_paging cli cmd """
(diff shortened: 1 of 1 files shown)


zhecho:
I'll rewrite it with PEP8 requirements

johnarnold:
Related problem -- TMSH can give you really long prompts that the find_prompt function butchers:

h.connection.find_prompt()
read_channel: 
write_channel: b'\n'
read_channel: 
mos)# 
read_channel: 
'mos)#' 

you can fix it by setting the tty width:

h.connection.send_command("run /util bash -c \"stty cols 255\"")
...
h.connection.find_prompt()
read_channel: 
write_channel: b'\n'
read_channel: 
netscript1_rw@(cys01-f72-ucgmc-1a)(cfg-sync Changes Pending)(Standby)(/Common)(tmos)# 
read_channel: 
'netscript1_rw@(cys01-f72-ucgmc-1a)(cfg-sync Changes Pending)(Standby)(/Common)(tmos)#' 

Do you want to add "run /util bash -c \"stty cols 255\""  to session_preparation, so that nobody else hits this?

ktbyers:
@zhecho Why did you remove the entering into `tmsh` mode (line 18)?

ktbyers:
@zhecho @johnarnold I made this change here for the two issues (output paging and terminal width):

https://github.com/ktbyers/netmiko/commit/ac4d676c66f037a850efc452bfd8d6562ae64947

Can one of you test it?

It would be good if we could run the set of Netmiko unit tests against this: 

@zhecho I didn't really want to do the F5 version detection...if you think we really need to do this, let me know. I am a bit out of the loop on F5 LTM, but it sounds like that paging command should go back quite a ways (as far as F5 OS version).

Also, @zhecho just submit a separate PR for your HP change.

Issue #568, opened 2017-09-02
how to detect connection closed by remote host


andreas268:
Kirk,

first of all: Many thanks for this great lib!!!!!!

I'm using it to manage IOS versions on a bunch of routers and switches and it works pretty fine.
But there are some devices with very slow bandwidth and my copy (copy http/ftp://ip/image flash) jobs fail because of exec-timeout terminating the connection before the file is copied (netmiko connections are still active by using delay_factor and  max_loops).
Is there any possibility to detect the reason (ssh client prints "Connection to 1.1.1.1 closed by remote host")? Logging shows "Search pattern never detected in send_command_expect: bytes copied in|Error opening".

Python and netmiko versions:
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import netmiko
>>> netmiko.__version__
u'1.4.2'

Thanks,
Andreas

ktbyers:
@andreas268 I would probably implement something like the following:

```
    def is_alive(net_connect):
        null = chr(0)

        try:
            # Try sending ASCII null byte to maintain the connection alive
            net_connect.write_channel(null)
            return True
        except (socket.error, EOFError):
            # If unable to send, we can tell for sure that the connection is unusable
            return False
        return False
```

This is based on the is_alive method in napalm-ios.

Basically, I would catch the exception you get with `send_command()` and then check `is_alive`.

Let me know if that works.

andreas268:
Kirk,

thanks for this advice, I'm currently testing it. I'll keep you updated...

In my logs I've seen that (paramiko) transport "EOF" is recognized but netmiko still reads the channel. 
Maybe it's a good idea to handle this event in netmiko!?

andreas268:
Sorry, forgot the paramiko logs:


2017/09/11 10:40:29,159 paramiko.transport DEBUG [chan 0] Sesch channel 0 request ok
2017/09/11 10:40:29,256 paramiko.transport DEBUG [chan 0] Sesch channel 0 request ok


2017/09/11 10:50:47,96 paramiko.transport DEBUG EOF in transport thread

ktbyers:
@andreas268 Handle it where? 

In general, I would say if the SSH connection just closes, Netmiko should just raise that error. The person writing code using Netmiko can then decide what to do about it (do nothing, log a message, re-connect, etc).

[...]

PR #576, opened 2017-09-15
Define is_alive class method

mirceaulinic:
To avoid redundancy across SSH-based NAPALM drivers, and also for other
eventual Netmiko needs.

@ktbyers I basically just copied this from napalm-ios and adapted here. I opted to return only the boolean flag, as in opposite to returning a dictionary. Happy to change if you disagree.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -224,6 +224,34 @@ class BaseConnection(object):
             # Always unlock the SSH channel, even on exception.
             self._unlock_netmiko_session()
 
+    def is_alive(self):
+        """Returns a boolean flag with the state of the connection."""
+        null = chr(0)
+        if self.remote_conn is None:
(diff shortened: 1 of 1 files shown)


ktbyers:
I set self.remote_conn to None here (i.e. cleaned this up to be more consistent with is_alive expectations):

https://github.com/ktbyers/netmiko/commit/27c8fd7b101788ef381450ac5291c71c801b066e

ktbyers:
Note, this issue:

https://github.com/ktbyers/netmiko/issues/568

Indicates that ASCII 0 (Null) didn't work and that he had to use ASCII 6 (ACK) for is_alive. This needs investigated more.

mirceaulinic:
Hrm, very interesting... I will be following that topic. Thanks for info @ktbyers!

PR #591, opened 2017-10-02
Serial connection support

tuxinger:
Hi ktbyers,
first of all, thanks for netmiko, it makes my live much easier. As I had the requirement for offline configuration and analytics of cisco IOS devices, I implemented serial support into netmiko. I mostly reused telnet functions because at least for cisco devices serial and telnet is more or less the same. Usage is straight forward:
serialhandle = {
        'device_type':'cisco_ios_serial',
        'port': 'USB Serial', # can be COM<number> or any line you can get from
                              # serial.tools.list_ports.comports() 
        'username':'<username>',
        'password':'<password>',
        'secret':'<secret>',
        'serial_settings':{ # this are the default values
                'baudrate': 9600,
                'bytesize': serial.EIGHTBITS,
                'parity': serial.PARITY_NONE,
                'stopbits': serial.STOPBITS_ONE
            }
        }

I would love seeing this merged into upstream. 
BR
Wilhelm Putz


netmiko/base_connection.py
```diff
@@ -36,7 +37,8 @@ def __init__(self, ip='', host='', username='', password='', secret='', port=Non
                  device_type='', verbose=False, global_delay_factor=1, use_keys=False,
                  key_file=None, allow_agent=False, ssh_strict=False, system_host_keys=False,
                  alt_host_keys=False, alt_key_file='', ssh_config_file=None, timeout=8,
-                 session_timeout=60, keepalive=0, default_enter=None, response_return=None):
+                 session_timeout=60, keepalive=0, default_enter=None, response_return=None,
+                 serial_settings={}):
```
ktbyers: You should make serial_settings default to None and then do:

```
if serial_settings is None:
    serial_settings = {}
```

Otherwise you run into an issue with Python mutables (i.e. we would get some unwanted side-effects).

ktbyers:
@tuxinger Good idea! I added a single comment. I will need to review more thoroughly and test this.

ktbyers:
I will also have to figure out how to test this :-)

tuxinger:
@ktbyers It should run with any serial or serial USB adapter. I mainly tested on python for windows. For usage have a look to tests/test_cisco_ios_serial.py.

netmiko/base_connection.py
```diff
@@ -105,15 +107,19 @@ def __init__(self, ip='', host='', username='', password='', secret='', port=Non
...
             raise ValueError("Either ip or host must be set")
         if port is None:
             if 'telnet' in device_type:
                 self.port = 23
             else:
                 self.port = 22
         else:
-            self.port = int(port)
+            if 'serial' in device_type:
+                self.port = check_serial_port(port)
```
ktbyers: It looks like here you are overloading the meaning of port? In other words, we are normally using this for TCP port 22 and TCP port 23, but here you are switching it to mean 'com1', 'com2', et cetera?

Is that right?

netmiko/base_connection.py
```diff
@@ -272,6 +295,10 @@ def _read_channel(self):
                     break
         elif self.protocol == 'telnet':
             output = self.remote_conn.read_very_eager().decode('utf-8', 'ignore')
+        elif self.protocol == 'serial':
+            output = ""
+            while (self.remote_conn.in_waiting > 0):
+                output += self.remote_conn.read(self.remote_conn.inWaiting())
```
ktbyers: Why call inWaiting() here versus using `in_waiting` that was used in while loop?

http://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.in_waiting

netmiko/base_connection.py
```diff
@@ -379,6 +406,12 @@ def read_until_prompt_or_pattern(self, pattern='', re_flags=0):
             combined_pattern = r"({}|{})".format(combined_pattern, pattern)
         return self._read_channel_expect(combined_pattern, re_flags=re_flags)
 
+    def serial_login(self, pri_prompt_terminator=r'#\s*$', alt_prompt_terminator=r'>\s*$',
+                     username_pattern=r"(?:[Uu]ser:|sername|ogin)", pwd_pattern=r"assword",
+                     delay_factor=1, max_loops=20):
+        self.telnet_login(pri_prompt_terminator, alt_prompt_terminator, username_pattern,
+                          pwd_pattern, delay_factor, max_loops)
+
```
ktbyers: This is fine...but I probably will abstract out the common login code into a new method (i.e. the username/password) handling and then call it from both places.

netmiko/base_connection.py
```diff
@@ -123,6 +129,15 @@ def __init__(self, ip='', host='', username='', password='', secret='', port=Non
...
         self.keepalive = keepalive
+        if serial_settings is None:
+            serial_settings = {}
+        self.serial_settings = serial_settings
+        self.serial_defaults = {
+                'baudrate': 9600,
+                'bytesize': serial.EIGHTBITS,
+                'parity': serial.PARITY_NONE,
+                'stopbits': serial.STOPBITS_ONE
+            }
```
ktbyers: Can we just do the following?

```
if serial_settings is not None:
    self.serial_settings = {
              'baudrate': 9600,
              'bytesize': serial.EIGHTBITS,
              'parity': serial.PARITY_NONE,
              'stopbits': serial.STOPBITS_ONE
            }
    self.serial_settings.update(serial_settings)
```

In other words, set the defaults and then update any defaults that were over-ridden in `__init__`.

netmiko/base_connection.py
```diff
@@ -534,6 +567,13 @@ def establish_connection(self, width=None, height=None):
         if self.protocol == 'telnet':
             self.remote_conn = telnetlib.Telnet(self.host, port=self.port, timeout=self.timeout)
             self.telnet_login()
+        elif self.protocol == 'serial':
+            self.serial_settings["port"] = str(self.port).split(" ")[0]
```
ktbyers: See earlier comment on overloading of port (i.e. I think we should make it a distinct attribute like `comm_port` (or something like that).

netmiko/base_connection.py
```diff
@@ -534,6 +567,13 @@ def establish_connection(self, width=None, height=None):
...
+            for k in self.serial_settings:
+                self.serial_defaults[k] = self.serial_settings[k]
+            self.serial_settings = self.serial_defaults
```
ktbyers: I think we can just get rid of these lines and just use the `update` method in `__init__`.

netmiko/cisco_base_connection.py
```diff
@@ -133,6 +145,8 @@ def cleanup(self):
             # Always try to send 'exit' regardless of whether exit_config_mode works or not.
             pass
         self.write_channel("exit" + self.RETURN)
+        if self.device_type in 'serial':
+            self.remote_conn.close()
```
ktbyers: It should do this in the disconnect method so we shouldn't need it here.

ktbyers:
@tuxinger Can you test the current `develop` branch?

I merged your PR in. I then made a set of modifications (in line with my comments above)...see here:

https://github.com/ktbyers/netmiko/commit/f02d3bac1856952e987723e07dd248f63c9a4655

tuxinger:
Sure, I'll test it tomorrow. Thanks for merging.

netmiko/base_connection.py
tuxinger: Yes.

ktbyers: Okay, I separated that out such that you would pass a 'port' key inside the `serial_settings`.

ktbyers:
Sounds good @tuxinger....thanks for the contribution. Just let me know if I broke anything :-)

ktbyers:
...thinking about this a little bit more, we should probably just make a serial driver for all the platforms (and try to figure out a way to minimize the work to accomplish this).

The driver should just work as is (mostly) once you get passed the initial Username/Password login.

PR #593, opened 2017-10-04
Added the ability to pass a config_command to config_mode when callin…

autonomouse:
…g send_config_set

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -949,7 +949,8 @@ class BaseConnection(object):
             return self.send_config_set(cfg_file, **kwargs)
 
     def send_config_set(self, config_commands=None, exit_config_mode=True, delay_factor=1,
-                        max_loops=150, strip_prompt=False, strip_command=False):
+                        max_loops=150, strip_prompt=False, strip_command=False,
+                        config_mode_cmd=None):
         """
         Send configuration commands down the SSH channel.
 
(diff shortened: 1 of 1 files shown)


autonomouse:
Should have left a comment, sorry. So we're basically tying to use "configure private" for juniper, but I've noticed that the `output = self.config_mode()` in `send_config_set` will overwrite any external use of `netmiko_instance.config_mode(config_command='configure private')`.

netmiko/base_connection.py
```diff
@@ -949,7 +949,8 @@ def send_config_from_file(self, config_file=None, **kwargs):
             return self.send_config_set(cfg_file, **kwargs)
 
     def send_config_set(self, config_commands=None, exit_config_mode=True, delay_factor=1,
-                        max_loops=150, strip_prompt=False, strip_command=False):
+                        max_loops=150, strip_prompt=False, strip_command=False,
+                        config_mode_cmd=None):
```
markgoddard: Perhaps use config_command or config_mode_command for consistency?

netmiko/base_connection.py
```diff
@@ -968,7 +969,7 @@ def send_config_set(self, config_commands=None, exit_config_mode=True, delay_fac
             raise ValueError("Invalid argument passed into send_config_set")
 
         # Send config commands
-        output = self.config_mode()
+        output = self.config_mode(config_mode_cmd) if config_mode_cmd else self.config_mode()
```
markgoddard: Might be better to do this with a kwargs pattern:
```
kwargs = {}
if config_mode_cmd:
    kwargs['config_command'] = config_mode_cmd
output = self.config_mode(**kwargs)
```
Something similar could be achieved with positional args.

ktbyers:
I will modify this slightly after merged.

Issue #601, opened 2017-10-13
Context manager exception handling issues


markgoddard:
Create a connection using a `with` context manager, and raise an exception that takes multiple arguments within the context:

```
import netmiko

class MyException(Exception):
    def __init__(self, a, b):
        pass

with netmiko.ConnectHandler(...) as c:
    raise MyException(1, 2)
```
We see the following exception:

```
Traceback (most recent call last):
  File "t.py", line 8, in <module>
    raise MyException(1, 2)
  File "/home/stack/ngs-stress/venv/lib/python2.7/site-packages/netmiko/base_connection.py", line 166, in __exit__
    raise exc_type(exc_value)
TypeError: __init__() takes exactly 3 arguments (2 given)
```

Two problems:
1. The `__exit__()` method of `netmiko.BaseConnection` should just `raise exc_value` in order to reraise.
2. The `__exit__()` method of `netmiko.BaseConnection` should return, and python will reraise the exception automatically.

ktbyers:
Feel free to create a PR on it...or I will fix.

markgoddard:
I'll create a PR in the next day or two.

PR #602, opened 2017-10-15
Fix context manager exception handling

markgoddard:
When using a connection object as a context manager (via a with
statement), and the context raises an exception, the __exit__ method
should simply return a value other than True in order for the exception
to be raised. The current handling manually re-raises the exception, and
also creates a new exception object from the raised one in a way that
does not always work (e.g. if there are more than one arguments to the
exception's constructor). To resolve this we allow the context manager
to raise the exception.

Fixes: #601


markgoddard:
Thanks Kirk.

ktbyers:
Thanks for the fix.

PR #615, opened 2017-11-01
Break if the output is repated & Fix send_command_w_enter method

isidroamv:
We should break the loop if the output of the channel is not increasing, we can use `delay_factor` parameter if the commad could take time to start returning output.

For `send_command_w_enter` we expect the default prompt, _Cisco Controller_

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -806,6 +806,8 @@ class BaseConnection(object):
         # Keep reading data until search_pattern is found (or max_loops)
         i = 1
         output = ''
+        count_repeated_output = 0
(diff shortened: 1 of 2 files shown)


ktbyers:
@isidroamv I fixed the WLC error here:

https://github.com/ktbyers/netmiko/commit/4836b52a1b1fe95f5cd88cdd9e2d5b766d8bd1b2

This was a side effect of a change I made a while back (Netmiko 1.0 IIRC), but I didn't realize that I hadn't update the WLC.

What were you trying to do in the `break if output is repeated section` i.e. what issue are you trying to fix here?

PR #617, opened 2017-11-03
Add send_command_set method that send_config_set wraps

Boomerkuwanger:
As a network engineer I would like the ability to send a bulk of show commands and worry about the output after all the responses have returned. Also I would like to be able to send the following example:
```
show interface Ethernet1/1
configure terminal
interface Ethernet1/1
description foo
end
show interface Ethernet1/1
```
This allows an automated system to assert the state of the interface's description before and after the change.

To maintain backwards compatibility and keep the code clean I chose to have `send_config_set` wrap the new method `send_command_set` that is almost identical to the previous function.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -946,38 +946,47 @@ class BaseConnection(object):
         with io.open(config_file, "rt", encoding='utf-8') as cfg_file:
             return self.send_config_set(cfg_file, **kwargs)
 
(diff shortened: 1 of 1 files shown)


ktbyers:
I won't couple the config commands and show commands as keeping those decoupled is a pretty fundamental aspect of Netmiko and has a lot of implications across a lot of diverse platforms.

I would like to add the other part of this, however, which is send a list of show commands and get a list of responses from those show commands.

If you want to reimplement with the second part I would definitely be open to that.

PR #620, opened 2017-11-10
Add send_command_set and strip_commands methods

Boomerkuwanger:
This PR adds the following functionality:

- `send_command_set` function added that allows a list of commands to be sent and output collected after
- `strip_commands` method added that tries to find all commands in some output and remove them
- `_sanitize_output` method updated to call `strip_commands` when a command_set is passed in

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -877,7 +877,7 @@ class BaseConnection(object):
             command_set = [self.normalize_cmd(command) for command in command_set]
 
         for cmd in command_set:
-            self.write_channel(self.normalize_cmd(cmd))
+            self.write_channel(cmd)
             time.sleep(delay_factor * .5)
 
         # Gather output
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -519,7 +519,7 @@ class BaseConnection(object):
             output = self.strip_command(command_string, output)
         elif strip_command and command_set:
             command_set = [self.normalize_linefeeds(command) for command in command_set]
-            output = self.strip_commands(command_string, output)
+            output = self.strip_commands(command_set, output)
         if strip_prompt:
             output = self.strip_prompt(output)
         return output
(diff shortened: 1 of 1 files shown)


netmiko/base_connection.py
```diff
@@ -840,6 +843,50 @@ def send_command_expect(self, *args, **kwargs):
...
+            command_set = (command_set,)
+
+        if not hasattr(command_set, '__iter__'):
+            raise ValueError("Invalid argument passed into send_command_set")
+
+        if normalize:
+            command_set = [self.normalize_cmd(command) for command in command_set]
+
+        for cmd in command_set:
+            self.write_channel(self.normalize_cmd(cmd))
```
jjmanzer: This is a duplicate use of normalize_cmd().

netmiko/base_connection.py
```diff
@@ -509,14 +509,17 @@ def _connect_params_dict(self):
...
         """Sanitize the output."""
         if self.ansi_escape_codes:
             output = self.strip_ansi_escape_codes(output)
         output = self.normalize_linefeeds(output)
         if strip_command and command_string:
             command_string = self.normalize_linefeeds(command_string)
             output = self.strip_command(command_string, output)
+        elif strip_command and command_set:
+            command_set = [self.normalize_linefeeds(command) for command in command_set]
+            output = self.strip_commands(command_string, output)
```
jjmanzer: Should be command_set.

ktbyers:
I think we should return a list of results one for each command executed. 

This implies that I will change the behavior of this a bit.

ktbyers:
We could do something like the following:

```
    def send_command_set(self, command_set, use_timing=False, **kwargs):
        """Generator to send commands down the SSH channel."""
        if isinstance(command_set, string_types):
            command_set = (command_set,)
        if not hasattr(command_set, '__iter__'):
            raise ValueError("Invalid argument passed into send_command_set")
        for cmd in command_set:
            if use_timing:
                yield self.send_command_timing(cmd, **kwargs)
            else:
                yield self.send_command(cmd, **kwargs)
```

And then using this would be as follows:

```
for output in net_connect.send_command_set(command_set):
    print(output)
```

Or:

```
output = list(net_connect.send_command_set(command_set))
```

ktbyers:
I am going to close this...I am open to a slightly different implementation.

PR #624, opened 2017-11-14
  Privilege escalation ( "enable" mode) on HP Comware v5.x/v7.x before any other cli commands

zhecho:
Hello All,

Let's suppose we have RADIUS/AAA configuration and we don't have permission to use some cli commands, but we have user with full rights for the device. ... so we need to escalate our privileges before other commands.

For that purpose : 
   - Add priv_escalation() to hp_comware_ssh.py  
        - priv_escalation() should be executed before "screen-length disable"
        - priv_excalation() depends on comware version v5.x or v7.x 
	modified:   netmiko/hp/hp_comware_ssh.py

### DIFF
--- a/netmiko/hp/hp_comware_ssh.py
+++ b/netmiko/hp/hp_comware_ssh.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from netmiko.cisco_base_connection import CiscoSSHConnection
 import re
 
+
 class HPComwareSSH(CiscoSSHConnection):
 
     def session_preparation(self):
@@ -19,22 +20,24 @@ class HPComwareSSH(CiscoSSHConnection):
     def priv_escalation(self):
         """ Privilege Escalation before entering config mode """
         try:
-            pass
             out_version = self.send_command_timing("display version")
-            match_v = re.search(r'Comware\s+Software,\s+Version\s+(.*),\sRelease\s+',out_version, re.I | re.M)
-            # send 'e' in case of "---- More ----" returned by "display version"
-            self.send_command_timing('e') 
+            # regex_v = re.compile(r"""Comware\s+Software,\s+Version\s+(.*),
+            # \sRelease\s+""", out_version, re.I | re.M | re.X)
+            # match_v = regex_v.search(out_version)
+            match_v = re.search(r'Comware\s+Software,\s+Version\s+(.*),\sRelease\s+', out_version, re.I | re.M)
+            self.send_command_timing('e')
+
             # Comware version 5.xx
             if match_v and match_v.group(1).startswith('5'):
                 priv_escalation_cmd = 'super'
-                self.send_command(priv_escalation_cmd, expect_string='Password:', auto_find_prompt=False) 
-                self.send_command(self.secret, expect_string='>', auto_find_prompt=False) 
+                self.send_command(priv_escalation_cmd, expect_string='Password:', auto_find_prompt=False)
+                self.send_command(self.secret, expect_string='>', auto_find_prompt=False)
             # Comware version 7.xx
             elif match_v and match_v.group(1).startswith('7'):
                 priv_escalation_cmd = 'super level-15'
-                self.send_command(priv_escalation_cmd, expect_string='Username:', auto_find_prompt=False) 
-                self.send_command(self.username, expect_string='Password:', auto_find_prompt=False) 
-                self.send_command(self.secret, expect_string='>', auto_find_prompt=False) 
+                self.send_command(priv_escalation_cmd, expect_string='Username:', auto_find_prompt=False)
+                self.send_command(self.username, expect_string='Password:', auto_find_prompt=False)
+                self.send_command(self.secret, expect_string='>', auto_find_prompt=False)
         except Exception as e:
             raise e
 
--- a/netmiko/hp/hp_comware_ssh.py
+++ b/netmiko/hp/hp_comware_ssh.py
@@ -1,8 +1,8 @@
 from __future__ import print_function
 from __future__ import unicode_literals
-import time
-from netmiko.cisco_base_connection import CiscoSSHConnection
 
+from netmiko.cisco_base_connection import CiscoSSHConnection
+import re
 
 class HPComwareSSH(CiscoSSHConnection):
 
@@ -13,11 +13,30 @@ class HPComwareSSH(CiscoSSHConnection):
         """
         self._test_channel_read(pattern=r'[>\]]')
         self.set_base_prompt()
-        command = self.RETURN + "screen-length disable"
-        self.disable_paging(command=command)
-        # Clear the read buffer
-        time.sleep(.3 * self.global_delay_factor)
-        self.clear_buffer()
+        self.priv_escalation()
+        self.disable_paging(command="\nscreen-length disable\n")
+
+    def priv_escalation(self):
+        """ Privilege Escalation before entering config mode """
+        try:
+            pass
+            out_version = self.send_command_timing("display version")
+            match_v = re.search(r'Comware\s+Software,\s+Version\s+(.*),\sRelease\s+',out_version, re.I | re.M)
+            # send 'e' in case of "---- More ----" returned by "display version"
+            self.send_command_timing('e') 
+            # Comware version 5.xx
+            if match_v and match_v.group(1).startswith('5'):
+                priv_escalation_cmd = 'super'
+                self.send_command(priv_escalation_cmd, expect_string='Password:', auto_find_prompt=False) 
+                self.send_command(self.secret, expect_string='>', auto_find_prompt=False) 
+            # Comware version 7.xx
+            elif match_v and match_v.group(1).startswith('7'):
+                priv_escalation_cmd = 'super level-15'
+                self.send_command(priv_escalation_cmd, expect_string='Username:', auto_find_prompt=False) 
+                self.send_command(self.username, expect_string='Password:', auto_find_prompt=False) 
+                self.send_command(self.secret, expect_string='>', auto_find_prompt=False) 
+        except Exception as e:
+            raise e
 
     def config_mode(self, config_command='system-view'):
         """Enter configuration mode."""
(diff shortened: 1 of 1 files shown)


ktbyers:
@zhecho Does "screen-length disable" require the privilege escalation to happen before this? 

I will speak in Cisco terms as I am not sure whether this maps to Comware or not...but does 'screen-length disable' work at the equivalent of privilege-level1?

Does the screen-length disable not work because of the TACACS+ configuration?

The current Netmiko-Comware code also has this:

```
    def enable(self, cmd='system-view'):
        """enable mode on Comware is system-view."""
        return self.config_mode(config_command=cmd)
``` 

Is this wrong? What you implemented really seems like it should map to an enable method?

My other concern is that this code would fail if there is no secret defined. In other words, it would always require everyone to go through this `priv_escalation` even if you were already at the right privilege level (since the method is always called in session_preparation)

zhecho:
Hello Byers,

Yes, privilege escalation should happen before "screen-length disable" if
you have radius/tacacs configured on the device.

"system-view" is like enable mode in Cisco devices, configuration is
possible only after this command

I'm not able to look at the implementation of the Cisco code about "enable
mode" right now but it should be the same as if you have configuration for
password on Cisco enable mode.

How did you decide this problem on the Cisco devices..?

I think that we should expect "]" or "Username:" or "Password:" or ">"  as
output of system-view command . What is your thoughts about this?

We can try this on the hp comware simulator. And make some tests for
version 7... But after my vacation ;) regards from sunny Greece ;)

Merry Christmas!
..and Happy New year..
Best regards!

On Dec 30, 2017 11:21, "Kirk Byers" <notifications@github.com> wrote:

@zhecho <https://github.com/zhecho> Does "screen-length disable" require
the privilege escalation to happen before this?

I will speak in Cisco terms as I am not sure whether this maps to Comware
or not...but does 'screen-length disable' work at the equivalent of
privilege-level1?

Does the screen-length disable not work because of the TACACS+
configuration?

The current Netmiko-Comware code also has this:

    def enable(self, cmd='system-view'):
        """enable mode on Comware is system-view."""
        return self.config_mode(config_command=cmd)

Is this wrong? What you implemented really seems like it should map to an
enable method?

My other concern is that this code would fail if there is no secret
defined. In other words, it would always require everyone to go through
this priv_escalation even if you were already at the right privilege level
(since the method is always called in session_preparation)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/ktbyers/netmiko/pull/624#issuecomment-354536062>, or mute
the thread

ktbyers:
@zhecho Is the below a function of your TACACS/RADIUS config or a general statement for Comware and TACACS/RADIUS (it seems a bit of a strange behavior):

```
Yes, privilege escalation should happen before "screen-length disable" if
you have radius/tacacs configured on the device.
```


```
"system-view" is like enable mode in Cisco devices, configuration is
possible only after this command
```

That sounds more like `config mode` on Cisco devices and not enable (which is privilege level 15)?


```
How did you decide this problem on the Cisco devices..?
```

I didn't understand this question, can you expand on it?


```
I think that we should expect "]" or "Username:" or "Password:" or ">"  as
output of system-view command . What is your thoughts about this?
```

So `system-view` will prompt you for a Username/Password if you don't have privileges to go directly into system-view?


So issues I see:
1. I think this should probably map this privilege escalation to enable. I think it is probably more logically consistent with Netmiko behavior.
2. We need to make sure the code doesn't break for users that already login with the proper privileges.

Adding @netmanchris onto this thread so he can comment also.

netmanchris:
Just a quick chime in - Comware 5 and 7 both have a user-level similar to Cisco. Comware5 uses 0-3 ( 3 ==  15)  and Comware7 uses 0-15 which maps to the Cisco methods like you would expect it to. 

When dealing with RADIUS, I typically have the set the service-level sent automatically. It’s been a long time since I locked at this though... so I’d have to double check.

@ktbyers - You should be able to follow the similar pattern to what you’ve done with the Cisco method.  The “system-view” is actually closer to the “config terminal” in Cisco terms than the “enable” mode. Although it often gets compared to the enable, it’s more accurate to say it’s a combination of enable+config terminal together as a single command. 

It’s possible to login with a lower privilege user-level and then escalate priveleg but I’d have to go back and look at that to figure out the exact sequence.  

Once @zhecho gets back from his sunny vacation, we can see how he wants to handle this and we’l figure it out what makes sense.

ktbyers:
@netmanchris Chris, thanks for chiming in.

It would be nice to automatically configure `screen-length disable`, but we just have to make sure it works for the various use cases.

It is not clear to me whether that normally breaks in the TACACS+/RADIUS situation or whether that is a special configuration for @zhecho (i.e. that they have this configured more tightly than normal).

zhecho:
Hello All,

Let's try to execute  "screen-length disable" after logging to the device
(in session_preparation() ) and if it returns "Permission denied" we'll
trow some log  message and execute priv_escalation() method and again
disable_paging().  This should resolve problems with/without TACACS/RADIUS
configuration permissions.

What do you think about this ?

@ktbyers <https://github.com/ktbyers>

Maybe it's not good idea to compare "system-view" with cisco "enable" mode
at all @netmanchris <https://github.com/netmanchris> - is absolutely right
.. -> "it’s more accurate to say it’s a combination of enable+config
terminal together as a single command"?


in Mordor again ;)
Best Regards!

ktbyers:
@zhecho That sounds reasonable. 

I also think we should change some of the Comware Netmiko behavior with respect to `enable` and config mode so that `system-view` maps to config mode and enable maps to the privilege escalation.

Does that sound reasonable @netmanchris and @zhecho?

zhecho:
Hello @ktByers, @netmanchris, all,

Yes,  it's sounds  reasonable for me too. Take a look at
https://github.com/ktbyers/netmiko/pull/624/commits/8f0bd49c73885a2dc070a49224fca7d59d8256b7
.. it should resolve those problems. I've made some tests with real devices
and it works as expected. Please feel free to comment and give me some
feedback.

Best Regards!

ktbyers:
@zhecho @netmanchris  I don't think we should try to detect the version:

```
+        if self.com_version in '5':
 +            self.enable_cmd = 'super'
 +        elif self.com_version in '7':
 +            self.enable_cmd = 'super level-15'
```

I think we should just execute the command and catch the failure. What happens on Comware5 if you do `super level-15` (i.e. what error message shows)?

I say this as it is problematic to do much before the device is initialized by session_preparation method.

ktbyers:
Also what does the CLI look like when you do `super level-15` on a Comware7? I want to see how the prompts change.

netmanchris:
I could check it out, but Comware5 didn’t have a level super-level 15, the authentication levels were 0-3. With 3 being the equivalent of the 15 in IOS/CW7 syntax. 


Christopher Young
https://kontrolissues.net/ blog
https://homekitgeek.com/ blog
https://goo.gl/apT9cd/ YouTube
https://goo.gl/bPZBQ3/ Udemy
https://github.com/netmanchris/ Github
@netmanchris/ Twitter
+1(514) 244-0868 / Mobile

ktbyers:
@netmanchris Yes, I know it doesn't have it, but I want to see the error message that Comware5 generates so I can catch it and then do the Comware5 command.

In other words, first try the Comware7 command (`super level-1 level-15`), if that fails, then catch the error and try the Comware5 command.

netmanchris:
Cool. Out and about right now. I’ll capture output and post later. 

> 
Christopher Young
https://kontrolissues.net/ blog
https://homekitgeek.com/ blog
https://goo.gl/bPZBQ3/ Udemy
https://github.com/netmanchris/ Github
@netmanchris/ Twitter
+1(514) 244-0868 / Mobile

zhecho:
Hello all,

That's the output of 2 switches with comeware 5 and 7 version. The problem
is that version 7 accept 'super' command ask for credentials and get
authentication error sending proper credentials. So probably  we can do
7th  version command (super level-15) first.


Comware version 5


<switch-12508>super ?

  INTEGER<0-3>  Priority level

  <cr>


<switch-12508>super level-15

                                ^

% Wrong parameter found at '^' position.

<switch-12508>display version | include Comware

HP Comware Platform Software

Comware Software, Version 5.20.101, Release 1829
<switch-12508>super

Password:

Password:

User privilege level is 3, and only those commands can be used

whose level is equal or less than this.

Privilege note: 0-VISIT, 1-MONITOR, 2-SYSTEM, 3-MANAGE

<switch-12508>



Comware version 7

switch -> 5900AF-48XG-4QSFP+

version -> 7.1.045, Release 2422P02





<switch-5900AF>super ?

  STRING<1-63>      User role name

  network-admin

  network-operator

  network-operator

  level-0

  level-1

  level-2

  level-3

  level-4

  level-5

  level-6

  level-7

  level-8

  level-9

  level-10

  level-11

  level-12

  level-13

  level-14

  level-15

  security-audit

  <cr>

<switch-5900AF>super le

<switch-5900AF>super

Username: username

Password:

Username: username

Password:

Username:

Authentication failed.

<switch-5900AF>super lev

<switch-5900AF>super level-15

<switch-5900AF>super level-15

Username: username

Password:

User privilege role is level-15, and only those commands that authorized to
the role can be used

<switch-5900AF>


Best regards!

ktbyers:
So if I do a `super level-15` and it prompts me for a username, is this almost always going to be a different 'username' than the one I logged in with?

This is on Comware7?

ktbyers:
So per communication with @netmanchris, I think we are probably going to do the following:

1. Handle the error that comes back if the disable paging call fails. If this failure happens, then call enable() method to elevate privileges.

2. Create an enable method around `super level-15`.

Once again this is all around Comware7. @netmanchris mentioned that there might be similar issues for Provision/ProCurve products.

Outstanding issues:
- What is the error string that comes back on a failed `screen-length disable` call (due to insufficient privileges)?
- Do we need two separate usernames for `super level-15` or is a single username sufficient here?

ktbyers:
I am going to close this...it doesn't seem like we are making any progress on it and I don't have sufficient information/knowledge to complete it myself.

PR #638, opened 2017-11-26
Set ansi_escape_codes=True to support Alpine Linux

ogenstad:
To support tests from Brigade.


ktbyers:
@ogenstad I pushed this change in this commit. I added a call to super so it does all of the standard session preparation items.

Also I added the one particular ansi escape code into the strip ansi section.

PR #645, opened 2017-12-01
Add JUNOS SCP Checksum

frankiexyz:
Hi, I would like to implement JUNOS SCP Checksum like what Cisco does.


ktbyers:
I will review, but note that SCP transfer mechanism for Juniper is only partially done (maybe a 1/3 done). I started it, but it definitely is not complete.

PR #654, opened 2017-12-11
Support huawei_vrpv8

gaozhangfei:
huawei_vrpv8 switch has been supported in networking-generic-switch
Support huawei_vrpv8 in netmiko as well

Otherwise, error reported like:
networking_generic_switch.devices.netmiko_devices.huawei_vrpv8:Huawei: Netmiko does not support device type huawei_vrpv8

### DIFF
--- a/netmiko/ssh_dispatcher.py
+++ b/netmiko/ssh_dispatcher.py
@@ -90,6 +90,7 @@ CLASS_MAPPER_BASE = {
     'hp_comware': HPComwareSSH,
     'hp_procurve': HPProcurveSSH,
     'huawei': HuaweiSSH,
+    'huawei_vrpv8': HuaweiSSH,
     'juniper': JuniperSSH,
     'juniper_junos': JuniperSSH,
     'linux': LinuxSSH,


gaozhangfei:
Huawei VRPV8 switch is added in https://review.openstack.org/#/c/473800/
See bug: https://bugs.launchpad.net/networking-generic-switch/+bug/1697676

However, netmiko does not support Huawei VRPV8 switch, 
see bug: https://bugs.launchpad.net/networking-generic-switch/+bug/1736341

This patch solve this issue.
Help take a look

Thanks

ktbyers:
@gaozhangfei Given the report description, it sounds like you should make a separate class in Netmiko and add a commit method (i.e. inherit from `HuaweiSSH` class):

Something like:

```
class HuaweiVRPV8(HuaweiSSH):
    def commit(self, confirm=False, confirm_delay=None, comment='', label='', delay_factor=1):
```

You can look at the IOS-XR or Juniper code in Netmiko for an example of commit.

gaozhangfei:
@ktbyers thanks for the reply.

Not sure I am understand correctly, not quite familiar.

Huawei VRPV8 switch had already been merged https://review.openstack.org/#/c/473800/ for networking-generic-switch, which simply add "commit" to each command, since it is the only one difference.
With this change, Huawei VRPV8 behaves same as HuaweiSSH.
If we now add "commit" by def commit method, it will send two "commit" for each command, which maybe wrong.

So we can simply reuse HuaweiSSH with additional type device_types.

What do you think?

Thanks.

ktbyers:
@gaozhangfei The proper way to do this in Netmiko is to implement the `commit()` method. 

For the same reason that OpenStack needed to differentiate between standard Huawei devices and VRRP8, Netmiko will need to do the same thing. 

In other words, you are asking Netmiko to be the same as the standard Huawei device when it isn't the same (i.e. it needs a commit method).

Note, this wouldn't break the openstack code, you would just not be calling the `commit` method (i.e. you would be doing the commit manually and not via the commit method).

Does that make sense?

gaozhangfei:
@ktbyers thanks for the suggestion
Have resubmitted https://github.com/ktbyers/netmiko/pull/671, 
Could you help take a look.

PR #664, opened 2018-01-02
Add Enable mode support to SROS.

ddfrathb:
Add the functionality to enter/exit enable mode using the SSH connection.

### DIFF
--- a/netmiko/alcatel/alcatel_sros_ssh.py
+++ b/netmiko/alcatel/alcatel_sros_ssh.py
@@ -25,8 +25,13 @@ class AlcatelSrosSSH(CiscoSSHConnection):
             self.base_prompt = match.group(1)
             return self.base_prompt
 
-    def enable(self, *args, **kwargs):
-        pass
+    def enable(self, cmd='enable', pattern='password', re_flags=re.IGNORECASE):
+        """Enter enable mode."""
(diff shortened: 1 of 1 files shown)


ktbyers:
@ddfrathb Can you show me in the CLI what using `enable` and `disable` looks like?

Also is `enable` the `enable-admin` command? I was having trouble finding just `enable` in the SROS CLI guide?

ddfrathb:
Yes, the enable-admin command is enable - sorry I actually forgot about this alias. it auto completes the command.  I realized later that the check_enable will actually fail on this and I was actually being authenticated as an escalated user - so I wasn't seeing the "Already enabled" error message. The disable also was my mistake.. apologies for the confusion as again it was checking admin state and "succeeding" as prompts are the same. I believe you would have to run "enable" and look for "Already enabled" in order to check, rather than relying on a prompt.

Fail:
*A:HOSTNAME# enable
Password:
MINOR: CLI Invalid password.
*A:HOSTNAME# 

Success:
*A:HOSTNAME# enable
Password:
*A:HOSTNAME# 

Already enabled:
*A:HOSTNAME# enable
MINOR: CLI Already in admin mode.

ktbyers:
@ddfrathb Okay, do you want to update this PR, to reflect the correct behavior (for the `enable` method)?

Is there anything like a `show privilege` command that would tell us if we were in `enable` without having to check the `CLI Already in admin mode`.

ktbyers:
Re-implemented in PR #681

Issue #659, opened 2017-12-16
Netmiko fails to install on windows due to textfsm/terminal/fcntl import


ktbyers:

ktbyers:
https://github.com/google/textfsm/issues/20

ryanbowen:
Any progress on this?

ktbyers:
You can test this branch:

https://github.com/ktbyers/netmiko/tree/clitable_fix

There is a fix in there. I haven't released it yet, but let me know if it works.

ryanbowen:
I'll give it a test run tomorrow morning when I get back into work. Thanks for the update.

mreams13:
It works for me (Windows 10).

``` mreams@DESKTOP   play  C\..\Lib  python
Python 3.6.4 (v3.6.4:d48eceb, Dec 19 2017, 06:54:40) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import napalm
>>> driver = napalm.get_network_driver('ios')
>>> device = driver(hostname='192.168.1.1', username='mreams', password='password')
>>> device.open()
>>> device.load_replace_candidate(filename='super_candidate_config.txt')
```

ryanbowen:
@ktbyers It works for me also. Thanks for the update.

ktbyers:
Okay, I will see if I can clean it up slightly and get it released.

ktbyers:
This is fixed in Netmiko release 2.0.1

PR #666, opened 2018-01-04
fix netmiko

afourmy:
my apps break on windows py3.6 because of this dependency


ktbyers:
@afourmy It is fixed in a separate branch...see:

https://github.com/ktbyers/netmiko/issues/659

I will try to clean-up and release to pypi before too long.

PR #668, opened 2018-01-07
Clitable fix

ktbyers:
Fix textfsm windows install issue.


dtantsur:
Neat, I was just about to request the same :) One question @ktbyers: textfsm is still in requirements.txt, should it be removed?

dtantsur:
Also you may want to update LICENSE, as Google's code is APL-2.0

ktbyers:
I will update that _textfsm section to be licensed as APL-2.0.

PR #671, opened 2018-01-08
Support huawei_vrpv8 switch

gaozhangfei:
Huawei VRPV8 switch need to execute the command 'commit' to complete the command
Add new class HuaweiVrpv8SSH to support this switch

### DIFF
--- a/netmiko/huawei/huawei_ssh.py
+++ b/netmiko/huawei/huawei_ssh.py
@@ -80,3 +80,34 @@ class HuaweiSSH(CiscoSSHConnection):
         log.debug("prompt: {0}".format(self.base_prompt))
 
         return self.base_prompt
+
+
+class HuaweiVrpv8SSH(HuaweiSSH):
+
+    def commit(self, comment='', delay_factor=1):
+        """
+        Commit the candidate configuration.
+
+        Commit the entered configuration. Raise an error and return the failure
+        if the commit fails.
+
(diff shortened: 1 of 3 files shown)


netmiko/huawei/huawei_ssh.py
```diff
@@ -80,3 +80,34 @@ def set_base_prompt(self, pri_prompt_terminator='>', alt_prompt_terminator=']',
...
+        delay_factor = self.select_delay_factor(delay_factor)
+        error_marker = 'Failed to generate committed config'
+        command_string = 'commit'
+
+        if comment:
+            command_string += ' comment "{}"'.format(comment)
+
+        output = self.config_mode()
+        output += self.send_command_expect(command_string, strip_prompt=False,
+                                           strip_command=False, delay_factor=delay_factor)
```
ktbyers: You probably should exit config mode after you do this `self.exit_config_mode()`.

ktbyers:
Everything else looks good on this besides the `exit_config_mode`. Thanks for updating your original PR.

netmiko/huawei/huawei_ssh.py
gaozhangfei: Thanks  @ktbyers 
have updated with self.exit_config_mode(), the patch is verified with ironic baremetal with huawei switch ce5800

Issue #441, opened 2017-04-07
Implement save_config() method for all devices.


jumpojoy:
However there is commit() which already implemented but supports only limited hardware. It should be good to have common method for example save_config() which allow to save configuration for devices that doesn't support commit()

ktbyers:
And this method just saves running-config to start-config for platforms that use this model?

jumpojoy:
Exactly, or is a generic method that calls commit() if it is supported and if not do "copy running-config startup-config"

ktbyers:
@jumpojoy It is a lot of work to properly implement this on 25 or so drivers including unit tests. Are you signing up for doing that?

jacrotts:
I could help with this if needed, at least on cisco platforms.

jumpojoy:
@ktbyers I have only Cisco switches, but can try it with virtualized appliances for Juniper/Arista.

ktbyers:
I still think we need someone to sign-up for implementing on the other 20-25 platforms (i.e. working out the commands for performing a save_config for that platform). Or at least > half of them.

Even if we don't have a device to test on.

If we can get closer to 50-75% coverage on the platforms, then I would be more interested in doing this.

jacrotts:
I would be open to implementing it on the other platforms as well, but I just wouldn't have gear to test it on.

ktbyers:
Yes, if you want to take a go at implementing it that would be great. It is okay to not have the gear to test it on.

You just need to determine for each platform what the commands are to do this.

jacrotts:
I will start compiling a list of the commands for each platform.

jacrotts:
Here is my first pass at the supported/and somewhat tested platforms. 
I can go into more detail as needed!

[netmiko_save_config_commands.txt](https://github.com/ktbyers/netmiko/files/1037370/netmiko_save_config_commands.txt)

gmcms:
You can add (no enable nor config mode):

Alcatel AOS
write memory flash-synchro

Thanks

jacrotts:
added 
[netmiko_save_config_commands.txt](https://github.com/ktbyers/netmiko/files/1056381/netmiko_save_config_commands.txt)

jcrotts:
I am going to start working on this. I plan to do Cisco devices first, then I'll move on to the others.

PR #686, opened 2018-01-18
Implement Save Config for all Devices issue 441

jcrotts:
Still to do:
1. need to write tests for cisco devices
2. Implement save_config and tests for non cisco devices 
https://github.com/ktbyers/netmiko/issues/441


jcrotts:
What should I be targeting specifically with the tests, and is there a system for checking the tests on live or emulated devices?

ktbyers:
@jcrotts I will handle the tests. Is it ready to go besides that?

On the testing I have a set of live devices to test on (but only a subset of platforms).

jcrotts:
It should be! 
I implemented save_config() on the following. Happy to work through any issues.  I have access to all the Cisco OS's if you need any additional testing.

Cisco
Arista
Juniper 
HP 
Brocade
Huawei
Paloalto
Vyos
Alcatel
Avaya
Dell

solotronics:
just an FYI juniper is a different syntax

```agaddy@dar01.dal10-re0> configure private
warning: uncommitted changes will be discarded on exit
Entering configuration mode

{master}[edit]
agaddy@dar01.dal10-re0# commit and-quit
re0:
commit complete
Exiting configuration mode

{master}```

**Comment by ktbyers** [comment]
@solotronics good point...I will need to review the solution for Juniper as Netmiko already has a separate `commit` method for Juniper, IOS-XR, and one other platform.

**Comment by ktbyers** [comment]
@jcrotts I will try to see if I can review this, this weekend.

**Comment by jcrotts** [comment]
@solotronics Cool, thanks for the callout!
I was going based of the following, maybe that is just for a specific model of router?
https://www.juniper.net/documentation/en_US/junose16.1/topics/task/configuration/e-series-routers-current-configuration-save.html

@ktbyers Sounds good. For most of the OS's with a commit method I just had the save_config() method call the commit method.

**Comment by ktbyers** [comment]
@jcrotts I did some work on this here (using your work as the starting point):

https://github.com/ktbyers/netmiko/tree/jcrotts-develop

Main item was restructuring to allow for more code sharing. I also eliminated Cisco WLC as I didn't think it was right, but was from memory (and didn't have a device to test against). I also removed the call to commit for IOS-XR (i.e. I didn't think it was a good idea to mix commit and save_config). I am open to more discussion on this though.

There were some other minor changes.

T

**Comment by jcrotts** [comment]
CooI! will check it out thanks, for looking at it. I actually have a Cisco 5508 WLC, I'll post a screenshot of the command sequence to make sure It is correct. 

As far as mixing save_config and commit, I agree with you. I think I included it because the issue creator originally asked for it. I think I did a similar thing with the other devices that had commit commands, if so I'll remove those.

**Comment by ktbyers** [comment]
@jcrotts Sounds good on the WLC, I probably remembered wrong. We can re-add it.

Don't worry about the commit() I will remove the others. I think it doesn't make sense to group together because it is two different operations:

```
save_config == save already running config to startup config
commit == apply configuration items in candidate config to running config
```

**Comment by jcrotts** [comment]
Yeah I would agree, those are definitely distinct. Is there anything else I can do for this issue?

**Comment by ktbyers** [comment]
@jcrotts Let me see if I can work some more on integrating it tonight...and then see where we are at.

**Comment by jcrotts** [comment]
@ktbyers Okay happy to do whatever is needed. Any other issues I can work on in the meantime ?

**Comment by octupszhang** [comment]
cool, do you know networking-generic-switch?  It's an openstack project which use netmiko to configure switch, it also implement some save configuration method in it(may be because netmiko doesn't support to save configure before), but doesn't work in cisco. Are you working for that? If you are not, I will try to fix that after this patch merged in to netmiko. If you are also working for networking-generic-switch, I'm happy to wait for your patch in it. Thanks you.

**Comment by ktbyers** [comment]
@octupszhang No, we aren't doing any direct fixes to the `networking-generic-switch` in OpenStack so that would need fixed/patched on the OpenStack side.

File: netmiko/cisco/cisco_s300.py [hunk 1]
```diff
@@ -21,4 +21,9 @@ def session_preparation(self):
         self.set_terminal_width(command='terminal width 511')
         # Clear the read buffer
         time.sleep(.3 * self.global_delay_factor)
-        self.clear_buffer()
+
+    def save_config(self):
+        """ Saves config """
+        self.enable()
+        self.send_command(command_string='write', expect_string='[Yes')
```
octupszhang: hi, is this tested in physical switch? in my cisco s300 switch, the expect_string is (Y/N)
switch797d1f#write
Overwrite file [startup-config].... (Y/N)[N] ?N

ktbyers: Yes, I have an updated branch for this work and I had previously changed that to execute a 'write memory' and to confirm with a 'Y').

```
 25     def save_config(self, cmd='write memory', confirm=True, confirm_response='Y'):
 26         return super(CiscoS300SSH, self).save_config(cmd=cmd, confirm=confirm)
```

ktbyers:
Updated branch is here:

https://github.com/ktbyers/netmiko/tree/jcrotts-develop

ktbyers:
Updated PR at https://github.com/ktbyers/netmiko/pull/697

PR #693, opened 2018-01-26
Fix telnet connection to DELL PowerConnect switches

vinzBad:
My fixes for our PowerConnect switches.

### DIFF
--- a/netmiko/dell/dell_powerconnect.py
+++ b/netmiko/dell/dell_powerconnect.py
@@ -27,7 +27,7 @@ class DellPowerConnectBase(CiscoSSHConnection):
     def set_base_prompt(self, pri_prompt_terminator='>', alt_prompt_terminator='#',
                         delay_factor=1):
         """Sets self.base_prompt: used as delimiter for stripping of trailing prompt in output."""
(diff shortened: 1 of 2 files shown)


netmiko/dell/dell_powerconnect.py
```diff
@@ -27,7 +27,7 @@ def session_preparation(self):
     def set_base_prompt(self, pri_prompt_terminator='>', alt_prompt_terminator='#',
                         delay_factor=1):
         """Sets self.base_prompt: used as delimiter for stripping of trailing prompt in output."""
-        prompt = super(DellPowerConnectSSH, self).set_base_prompt(
+        prompt = super(CiscoSSHConnection, self).set_base_prompt(
```
ktbyers: This should be `DellPowerConnectBase` i.e. you call super using the class name of the class you are currently in.

ktbyers:
FYI, I corrected some of the super references here:

https://github.com/ktbyers/netmiko/commit/48f16e1a06049ad14d91f7f4ea9ab5470e47f7d1

Let me know if you have questions on this.

vinzBad:
Nice, thank you. I find the super keyword a bit confusing :D I think I get it now.

PR #697, opened 2018-02-03
Implement save_config method

ktbyers:


ktbyers:
@jacrotts I think this is good except that I need to re-add the Cisco WLC. We also need to add unit tests.

Let me know if you see issues.

jcrotts:
I'll look over everything, and I can start on the unit tests I will have a few uninterrupted hours on a flight later today. 

Should I assume, we have real network devices of each type, and then execute the save_config(method) and look for the expected response?

ktbyers:
@jacrotts See this document for testing:

I will test on real/virtual devices for all the platforms listed on the README.md as regularly tested. I won't test on anything else (as I don't have the device to test against).

jcrotts:
Would it make sense to check if the device is in config mode? IOS devices won't let you save the config unless you prepend 'do' to the command, and the enable command won't work from config.

`  """Saves Config."""
        if self.check_config_mode():
            self.exit_config_mode()):
        self.enable()
        if confirm:
            output = self.send_command_timing(command_string=cmd)
            if confirm_response:
                output += self.send_command_timing(confirm_response)
            else:
                # Send enter by default
                output += self.send_command_timing(self.RETURN)
        else:
            # Some devices are slow so match on trailing-prompt if you can
            output = self.send_command(command_string=cmd)
        return output`

ktbyers:
@jcrotts Not understanding why you would want to do that?

jcrotts:
The enable command and copy run start wouldn't work in config mode (see the screenshot). Not sure if that is worth checking though ?
screen shot 2018-02-10 at 6 09 04 pm

ktbyers:
I think what we currently have is fine. In general, people shouldn't be in config mode when they execute this and `send_config_set()` will automatically exit config mode (i.e. you would have to do something a bit unusual to be sitting in config_mode).

jcrotts:
Yeah you are right,  I can't imagine a situation where it would come up unless something odd was going on.

PR #708, opened 2018-02-10
Fix Python 3 compatibility

xavierhardy:


ktbyers:
I will review. I probably won't get to it this week though.

--- note, for other readers of this PR ---
the current Netmiko code is perfectly fine from a PY3 compatibility perspective. These fixes are all outside the library itself (examples, unit tests, et cetera) or entirely cosmetic (for example, u'prefix').

Netmiko only supports Python 3.5 and Python3.6 (and Python 2.7) so the u'prefix' doesn't cause any meaningful issues. I will probably fix with your fixes though (since it is a better way to have the code given that all string literals are already unicode due to the from `__future__` statement).

ktbyers:
This was done here. 

https://github.com/ktbyers/netmiko/commit/de5ae0bea4aa7c334cad0a3f7b920274a8ffb527

Merge was going to be too hard since I restructured most of those ./examples files.

I also didn't do any of the unicode prefix changes:

i.e.  u'whatever' is fine (since I only support PY2.7 and PY3.5 or greater).

I also didn't make some of the py23compat changes as what I had was fine.

PR #709, opened 2018-02-10
Use % strings in logs to avoid computing the string value outside of debug mode

xavierhardy:
Currently, debug messages are computed even if the log message is not displayed. To fix this, I removed `format` and appended its arguments to the `debug` method. 

Furthermore, % strings are much faster than `format` (though slower than Python 3.6+ f strings) and compatible in older Python versions (older than 2.6).
```py
>>> from logging import getLogger, INFO
>>> from timeit import timeit
>>> logger = getLogger()
>>> logger.setLevel(INFO)
>>> timeit(lambda: logger.debug("{}".format("test")))
1.7110893740027677
>>> timeit(lambda: logger.debug("%s", "test"))
1.4494130310049513
>>> timeit(lambda: "{}".format("test"))
0.3650825089935097
>>> timeit(lambda: "%s" % "test")
0.15864467000210425
```

### DIFF
--- a/netmiko/arista/arista_ssh.py
+++ b/netmiko/arista/arista_ssh.py
@@ -26,13 +26,13 @@ class AristaSSH(CiscoSSHConnection):
 
         Can also be (s2)
         """
-        log.debug("pattern: {0}".format(pattern))
+        log.debug("pattern: %s", pattern)
         self.write_channel(self.RETURN)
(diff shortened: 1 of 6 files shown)


ktbyers:
The performance difference isn't worth worrying about. Using .format() is fine; it is probably the canonical way to format python strings.

PR #710, opened 2018-02-10
Replace some while statements with more performant for each

xavierhardy:
```py
>>> from timeit import timeit
>>> def wh():
...     i = 0
...     while i < 42:
...        i += 1
... 
>>> def fr():
...     for i in range(42):
...        pass
... 
>>> timeit(wh)
2.586169958114624
>>> timeit(fr)
1.5286250114440918
```


ktbyers:
It is probably not worth worrying about. We are usually sleeping a good chunk of time each time through the existing loop.

xavierhardy:
Let's close this then.

PR #713, opened 2018-02-12
Fix ProxyCommand case

xavierhardy:
There's a typo in line 575, it should be `'ProxyCommand'` not `'proxycommand'`. This will fail for sure, as the previous test checked its existence and failed.


ktbyers:
@xavierhardy Good catch...will fix.

PR #724, opened 2018-02-20
Unit tests for utilities.py and base_connection.py

xavierhardy:
Those unit tests cover most of the code that does not use `paramiko`, with basic lock tests included (#414).


ktbyers:
This was merged here:

https://github.com/ktbyers/netmiko/pull/1003

ktbyers:
@xavierhardy Thanks for the work here.

Issue #621, opened 2017-11-12
Login retry attempts for telnet method


zeepi222:
Hello,

I am connecting to a mix of SSH and Telnet enabled devices. 
I noticed that SSH method does not retry failed login attempts, whilst the telnet method will retry failed logins up to three times. I am concerned that multiple telnet failures could potentially lock out my TACACS account. 

Is it possible to control / set the number of login retry attempts used by the telnet method?

Example ...
```
In [29]: try:
    ...:         cli_session = ConnectHandler(ip="10.2.3.5",
    ...:                                      username="user1",
    ...:                                      password="passwd1",
    ...:                                      device_type="cisco_ios_telnet",
    ...:                                      port=23,
    ...:                                      verbose=True)
    ...:         print cli_session.find_prompt()
    ...:     
    ...: except Exception as err:
    ...:     logging.exception(err)
    ...:     
DEBUG:netmiko:read_channel: 

User Access Verification

Username: 
DEBUG:netmiko:write_channel: user1

DEBUG:netmiko:read_channel: user1
Password: 
DEBUG:netmiko:write_channel: passwd1

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
% Login invalid

Username: 
Username: 
Username: 
Username: 
Username: 
DEBUG:netmiko:write_channel: user1

DEBUG:netmiko:read_channel: user1
Password: 
DEBUG:netmiko:write_channel: passwd1

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: 

DEBUG:netmiko:read_channel: 
% Login invalid

Username: 
Username: 
Username: 
Username: 
Username: 
DEBUG:netmiko:write_channel: user1

DEBUG:netmiko:read_channel: user1
Password: 
DEBUG:netmiko:write_channel: passwd1

DEBUG:netmiko:read_channel: 
[...]

PR #729, opened 2018-02-28
Improve error handling, consistency of telnet_login()

ddisisto:
These changes to `BaseConnection.telnet_login()` & `CiscoBaseConnection.telnet_login()`:
1. Allow detection of login/authentication errors reported by remote device, raising an exception if encountered.
2. ~~Ensure that credentials are sent only once per method call.~~ _[no longer the case, see discussion below]_
3. Remove duplication between the two classes by capturing the differences into default parameter values.

No other `*Connection` class currently implement this method, but if they ever need to (e.g. different prompts, error reporting, etc) this should now be easier to do.

There are no tests that cover this code, although I'll keep an eye on https://github.com/ktbyers/netmiko/pull/724 and am happy to contribute tests in future if relevant.

Note that this is related to https://github.com/ktbyers/netmiko/issues/621, in which concerns were raised around login reliability with telnet. The only part of this PR that could potentially impact this is point 2 above (i.e. packet lost in transit, username written to channel but not received by remote host). My preference in this case is to raise an exception and have the end user to decide if/when/how to attempt retries in their own code, although am open to discuss and explore other options here.


ktbyers:
What issue is the current solution causing?

ddisisto:
I guess the main issue for me is really if the remote device replies to our credentials with something clearly indicating that those credentials are invalid, and then re-prompts for username, this IMO should raise an exception rather retrying with the same credentials again. This comes back to a) not locking out TACACS+ accounts.

I've made a new commit that reinstates the behaviour of responding to either username/password prompt as many times as they appear in the output (up to `max_loops`), but only while no explicit error is received.

Here's an example of the login process with this error parsing in place -
```
DEBUG:netmiko:read_channel: C

  [snip:MOTD]

Username:
DEBUG:netmiko:write_channel: ddisisto

DEBUG:netmiko:read_channel: ddisisto
Password:
DEBUG:netmiko:write_channel: incorrect_password

DEBUG:netmiko:read_channel:
DEBUG:netmiko:read_channel:
DEBUG:netmiko:read_channel:

% Authentication failed.


User Access Verification

Username:
---------------------------------------------------------------------------
NetMikoAuthenticationException            Traceback (most recent call last)
 [snip:traceback]
NetMikoAuthenticationException: Telnet login failed: 152.110.11.2; Authentication failed.
```
Does this seem like a more acceptable approach?

ddisisto:
Ahh, I've just come across that there are `% <timeout>` errors that would have been treated as errors, but for which are more appropriate for retrying. I suspect this is more the type thing you have reliability concerns around, and which I'd overlooked until now. I've just updated the `error_pattern` to be more specific to actual authentication failures, and less likely to trigger on communications / latency issues.

```
Username: ddisisto
Password:
% Password:  timeout expired!

User Access Verification

Username:
%
User Access Verification

Username:  timeout expired!
```

ktbyers:
Okay, I will think about it some more. I definitely am worried that only attempting a single login and not multiple logins will reduce the telnet login reliability in automation scripts.

In other words, in cases when you do have the right credentials, I think there are scenarios where it will fail on the first attempt (although it is possible these scenarios are more unlikely than I think they are).

We could possibly go to something that was more deterministic than the current process (i.e. it would only try to login three times and then give up), but I pretty reluctant to make it a single login. I guess another option is create an argument in __init__ that let's you override the default behavior and just login once.

ddisisto:
I've been giving it some more thought, and the only situation that I can come up with that would cause non-timeout login errors to fail initially, then succeed on subsequent retry is if bits were flipped in transit without triggering CRC / checksum failures. That failure mode to my knowledge is pretty uncommon, and should probably be ignored by network software (i.e. imagine if `requests` retried on `HTTP 400` just-in-case that 400 was really a 200 as sent by the server)

That said, regardless of whether the connection is made over telnet or SSH, if it's a very slow / unreliable / error-prone link, exceptions are going to occur sometimes anyway. If the user really needs a particular connection to occur if at all possible, then they'll want to perform exception handling and retries in their own code in either case in my opinion.

ktbyers:
I am going to close this as I think the current solution is probably better than the proposed change. I am happy to discuss some more, however.

Also telnet is very low on my priorities.


Regards, Kirk

PR #743, opened 2018-03-13
Fix cxr console

shengwang2016:


ktbyers:
Did you intend to submit this PR back upstream?

Assuming you intended to submit this back to me, you need to probably re-format it and re-submit it. A PR should really have only one bugfix or one new feature (and should not include other non-essential items like general documentation fixes).

You can do general documentation fixes that is good, but they should be in their own separate PR.

shengwang2016:
Hi Kirk,

Sorry for the inconvenience. Can you just close the PR? I submit it to the
wrong branch. I am intending to submit to a branch that is maintained by
Cisco.

Thanks

Sheng Wang

Issue #585, opened 2017-09-23
Make sure Calix B6 Driver RETURN is right


ktbyers:
This:

```
def __init__(self, *args, **kwargs):
         default_enter = kwargs.get('default_enter')
         kwargs['default_enter'] = '\r\n' if default_enter is None else default_enter
         super(CalixB6SSH, self).__init__(*args, **kwargs)
```

And this:
```
def exit_config_mode(self, exit_config='exit{}{}end'.format(self.RETURN, self.RETURN),
                          pattern=''):
```

carlniger:
Looks like this was addressed in #748 , closing.

PR #745, opened 2018-03-15
Add Calix B6 save_config and telnet support

rptetzloff:
Add Calix B6 save_config and telnet support
correct exit_config_mode (#585) 
add to device templates


ktbyers:
Replaced by #748

PR #749, opened 2018-03-18
Support DNOS6

danimo:
Almost like PowerConnect, but password auth via SSH actually happens on on SSH level.

This also aliases DNOS9, the successor of F10OS, to the F10OS implementation.

### DIFF
--- a/netmiko/dell/__init__.py
+++ b/netmiko/dell/__init__.py
@@ -1,6 +1,9 @@
 from __future__ import unicode_literals
+from netmiko.dell.dell_dnos6 import DellDNOS6SSH
+from netmiko.dell.dell_dnos6 import DellDNOS6Telnet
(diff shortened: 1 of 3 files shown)


ktbyers:
@danimo Thanks for the contribution. Can you run the unit tests against this (both for telnet and SSH)?

See here https://github.com/ktbyers/netmiko/blob/develop/TESTING.md


Let me know if you have questions.

Kirk

skg-net:
@danimo Thanks a lot for providing PR to support this. Can you please rename the aliases to dellos9 and dellos6 instead of dell_dnos6 and dell_dnos9 respectively. Reason being thats the standard name we follow in Dell EMC, also thats the same aliase used for Ansible networks_os_name.

ktbyers:
Replaced by https://github.com/ktbyers/netmiko/pull/952

PR #757, opened 2018-03-23
Add support for "_cmdline-mode on" in HP Procurve 1920 and HP Procurve 1910

gininfra:
Add support for "_cmdline-mode on" in HP Procurve 1920 and HP Procurve 1910


gininfra:
Add support for "_cmdline-mode on" in HP Procurve 1920 and HP Procurve 1910

PR #798, opened 2018-04-25
disable paging does not change fortinet configuration anymore

ylebiannic:
Hi,

On fortinet when the device is already configured with : 

```
config system console
   set output standard
end
```

When we connect/disconnect with netmiko, the device ends up with a different configuration
```
config system console
   set output more
end
```

I propose a simple patch that retreive the output mode in order to set it properly at cleanup without change on this configuration.

Cheers,

Yann

### DIFF
--- a/netmiko/fortinet/fortinet_ssh.py
+++ b/netmiko/fortinet/fortinet_ssh.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 import paramiko
 import time
+import re
 from netmiko.cisco_base_connection import CiscoSSHConnection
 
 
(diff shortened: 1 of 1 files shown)


ktbyers:
Thanks @ylebiannic

ktbyers:
@ylebiannic FYI, I also added the following into your code:

```
-        result_mode = reg_mode.search(output)
-        if result_mode:
-            self._output_mode = result_mode.group('mode')
+        result_mode_re = reg_mode.search(output)
+        if result_mode_re:
+            result_mode = result_mode_re.group('mode').strip()
+            if result_mode in ['more', 'standard']:
+                self._output_mode = result_mode
```

Basically, I put a check that the mode is going to be set to either 'more' or 'standard'. Let me know if you see any issues with this.

PR #800, opened 2018-04-26
Dell powerconnect : fix disable paging for Telnet connection

ylebiannic:
There is no reason to send different command to disable paging between SSH and Telnet connection. seems to be an oversight

### DIFF
--- a/netmiko/dell/dell_powerconnect.py
+++ b/netmiko/dell/dell_powerconnect.py
@@ -106,7 +106,7 @@ class DellPowerConnectTelnet(DellPowerConnectBase):
         self._test_channel_read()
         self.set_base_prompt()
         self.enable()
-        self.disable_paging(command="terminal length 0")
+        self.disable_paging(command="terminal datadump")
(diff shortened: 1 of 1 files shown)


ktbyers:
@rverchere Do you have any comments on this? I think you submitted the original code for the Powerconnect on telnet (that is getting changed here).

I saw two differences between the session_preparation for telnet and SSH.
1. Telnet code go into enable.
2. Telnet code uses 'terminal length 0' and not 'terminal datadump'

Do we need both of those differences?

If we can make them identical, then we should just use single session_preparation method for both.

rverchere:
Hi, I submitted the code for SSH, I did not use Telnet.

I cannot test this code, as I do not have access to a PowerConnect switch anymore, but I guess the `terminal datatumpd` works on both SSH and Telnet sessions.

ylebiannic:
I follow your advice and run some test on my Dell devices (ssh and telnet) and all seems to work fine with this.

PR #803, opened 2018-05-01
Add paloalto_panos to snmp_autodetect.py

liorfranko:


ktbyers:
@liorfranko It looks like you broke the lint checking in a minor way:

```
netmiko/snmp_autodetect.py:40:20: E128 continuation line under-indented for visual indent [pep8]
netmiko/snmp_autodetect.py:41:20: E128 continuation line under-indented for visual indent [pep8]
```

Can you fix these issues?

You can check by running `pylama .` from the netmiko root directory (same directory as the setup.cfg file).

Issue #774, opened 2018-04-05
save_config on Cisco WLC no longer working


rehlers:
Hello, it looks like a couple of months back, an update was done on the save_config function to make a more generic save_config method in the CiscoBaseConnection class and have each subclass just pass in the correct parameters.

The generic save_config was added to CiscoBaseConnection, but the cisco_wlc_ssh.py file with the CiscoWlcSSH class is a subclass of BaseConnection (not CiscoBaseConnection). BaseConnection does not have the generic save_config implemented, so now calling save_config on Cisco WLC devices raises a Not Implemented exception.

I don't know what the right solution here is, as the Cisco WLC interface is really nothing like anything else Cisco makes. 
Moving the generic save_config from CiscoBaseConnection to BaseConnection? 
Adding a CiscoWlcBaseConnection subclass of BaseConnection in cisco_base_connection.py that just has the save_config?
Making CiscoWlcSSH a subclass of CiscoSSHConnection instead?

ktbyers:
Yes, that is a bug...let me look at it some more regarding what solution makes sense for fixing it.

jcrotts:
You could also just override the save_config method in CiscoWlcSsh, since there probably won't be a need to subclass CiscoWlc.

jcrotts:
I can work on a PR for this, any preference on the above stated solutions?

rehlers:
great thanks so much for working on getting that issue resolved!

PR #815, opened 2018-05-14
    Fix for Issue  #774 save_config on Cisco WLC no longer working.

jcrotts:
Create a class in cisco_base_connection CiscoWlcConnection to house Wlc specific behavior that                   inherits from base connection.
    WlcSSH now inherits from this Cisco WlcConnection, so that Cisco
    devices are all inheriting from the cisco_base_connection file.
    Add a save_config function to the CiscoWlcConnection class.

### DIFF
--- a/netmiko/cisco/cisco_wlc_ssh.py
+++ b/netmiko/cisco/cisco_wlc_ssh.py
@@ -159,6 +159,18 @@ class CiscoWlcSSH(BaseConnection):
         log.debug("{}".format(output))
         return output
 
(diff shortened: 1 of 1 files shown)


ktbyers:
@jcrotts Instead of making an intermediate class that is only used by the Cisco WLC, I think it is better just to locate the `save_config` code directly into the `cisco_wlc_ssh.py` file.

jcrotts:
Yeah sounds good I'll do that later tonight.

PR #833, opened 2018-06-05
Added reset-graphics-mode code to strip_ansi_ecsape_codes list. Added…

mreams13:
… fix to strip ANSI escape codes before prompt search in send_command.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1028,6 +1028,8 @@ class BaseConnection(object):
         while i <= max_loops:
             new_data = self.read_channel()
             if new_data:
+                if self.ansi_escape_codes:
+                    new_data = self.strip_ansi_escape_codes(new_data)
                 output += new_data
                 try:
                     lines = output.split(self.RETURN)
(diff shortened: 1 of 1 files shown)


netmiko/base_connection.py
```diff
@@ -1354,7 +1357,7 @@ def strip_ansi_escape_codes(self, string_buffer):
         code_set = [code_position_cursor, code_show_cursor, code_erase_line, code_enable_scroll,
                     code_erase_start_line, code_form_feed, code_carriage_return,
                     code_disable_line_wrapping, code_erase_line_end,
-                    code_reset_mode_screen_options, code_erase_display,
+                    code_reset_mode_screen_options, code_reset_graphics_mode, code_erase_display,
```
ktbyers: Yes, this looks totally fine.

netmiko/base_connection.py
```diff
@@ -1028,6 +1028,8 @@ def send_command(self, command_string, expect_string=None,
         while i <= max_loops:
             new_data = self.read_channel()
             if new_data:
+                if self.ansi_escape_codes:
+                    new_data = self.strip_ansi_escape_codes(new_data)
```
ktbyers: I am a bit worried about how slow this is going to be...I assume this was on some Linux platform (I seem to recall on Slack or on your GitHub issue that you had stated it was)?

mreams13: I have to use Windows 10 for my development machine.

ktbyers: This is probably fine...let me just see how long this takes to execute, but it is probably no issue.

PR #837, opened 2018-06-09
Rely on using the cache for SSH autodetect

ktbyers:


ktbyers:
Yeah, I had forgotten that was even there...

PR #850, opened 2018-06-21
Update calix_b6.py

Neiby:
Moved 

def __init__(self, *args, **kwargs):
        super(CalixB6SSH, self).__init__(*args, **kwargs)

from CalixB6Base to CalixB6SSH


ktbyers:
Unfortunately, this is not quite the right solution.

The fix is that line 24 needs to be changed from:

```
-        super(CalixB6SSH, self).__init__(*args, **kwargs)
```

To

```
-        super(CalixB6Base, self).__init__(*args, **kwargs)
```

I think that should fix the problem.

You also have a very minor linter issue that you can see failing on the Travis-CI. Do you think you could update those two items?


Kirk

Neiby:
Okay, I think I have fixed it. This is my first time ever submitted a patch/pull request, so it has been a learning experience!

ktbyers:
Looks good.

Issue #842, opened 2018-06-15
Allow session with a device to be recorded


jobec:
It would be very handy to have netmiko log the session with a device to a specific file. Allowing to keep a log of every device session for logging/auditing/troubleshooting purposes. Like for example keeping all communications of the last month.

Because:

* **Threading/Multiprocessing:** using the logging feature that's implemented now (standard python logging) results in scrambled log files. All threads write at random in the log file. You can't make up what was send to which device.

* **Auditing/review:** For forensics, auditing, logging, troubleshooting... purposes you often need a single file of what happened towards a device. What commands were sent, what was returned. 

* **Current Logging:** With the current logging implementation, it's rather difficult to see at a glance what was sent. You first need to remove a lot of extra information from the logs.

* **Debug log levels:** At the moment you have to enable debugging log levels while I believe this kind of logging should not depend on a debug setting. If you want to do this logging always and in a structured way, you can't do it. Also, by sending this data to the logging subsystem of Python, you risk that it gets send to a remote syslog server. Given the amount of log messages it produces this is not a good idea.

jobec:
Example of a log file (with multithreading, 3 devices at once) with the default netmiko log output:
https://gist.github.com/jobec/ca17b0be5212dfbcfd7b45ff7f8fcd2a

Example of how a session log file could look like without other information, and only for a single device.
https://gist.github.com/jobec/a39b061cf29101cc037cfe9a9189d380

jobec:
A work around could be to write a wrapper around netmiko and implement all methods that send "something" to a device. Capture the command on the way in and out and write it to a file. But that is much more complex then doing it in the `_write_channel` and `_read_channel` methods. And it also doesn't allow you to capture things like disabling the screen paging and such.

Also, having this inside netmiko itself allows it to bubble up towards packages like napalm and ansible, where it's now very cumbersome to enable this kind of logging without drowning is log messages or breaking things.

ktbyers:
```
Threading/Multiprocessing: using the logging feature that's implemented now (standard python logging) results in scrambled log files. All threads write at random in the log file. You can't make up what was send to which device.
```

That is on the code writer i.e. you need to have a thread dedicated to this and some sort of an ID. I don't think the library is the proper place to handle this.

```
Auditing/review: For forensics, auditing, logging, troubleshooting... purposes you often need a single file of what happened towards a device. What commands were sent, what was returned.
```

Can't logging handle this (assuming you handle the concurrency issue?

```
[...]

Issue #851, opened 2018-06-25
Expose Full Output


jonesbra:
I am interested in seeing the full device output from beginning to end of the SSH session. This could be as simple as creating a property of the Base Connection class containing the full output, and anytime data is received from the stream appending to that property. I looked into the paramiko module and seen that they have a "makefile" method that can be used on channels such as Netmiko's remote_conn property of the base class to create "file-like objects". Would it be possible to use this to redirect output to a variable or file?

ktbyers:
@jonesbra The closest thing I have is probably this:

https://github.com/ktbyers/netmiko/blob/develop/COMMON_ISSUES.md#enable-netmiko-logging-of-all-reads-and-writes-of-the-communications-channel

ktbyers:
Yes, @jobec proposal in PR #855 should be very close to what you are requesting.

It is probably better to do what @jobec proposed in his PR as Netmiko also uses telnetlib and can use a serial channel (i.e. it would be good to make the session log independent of Paramiko.

Let us know if what is proposed in PR #855 solves your issue.

jonesbra:
That will work fine, thank you.

PR #855, opened 2018-06-29
Implement session log file

jobec:
A possible implementation for #842


ktbyers:
@jobec This looks reasonable to me. 

The only issue I see is that the `_read_channel_expect` method has an independent read of the SSH channel:

```
        while i < max_loops:
            if self.protocol == 'ssh':
                try: 
                    # If no data available will wait timeout seconds trying to read
                    self._lock_netmiko_session()
                    new_data = self.remote_conn.recv(MAX_BUFFER)
                    if len(new_data) == 0:
                        raise EOFError("Channel stream closed by remote device.")
                    new_data = new_data.decode('utf-8', 'ignore')
                    log.debug("_read_channel_expect read_data: {}".format(new_data))
                    output += new_data
```

You missed that so you would miss receiving this data in your session log.

I really need to consolidate this so that it is inside of _read_channel(), but until I do that you would need to write this to the session log separately.

I assume you have manually tested this?

I will write unit tests for it.

jobec:
I've added logging to the `_read_channel_expect` method too.

Yes, this was manually tested. But if you want, I'll add some unit tests for it. You'll have to give me some pointers as to how the tests suite in netmiko is set up though 😇 

concerning issue #851
Shall I adapt the code so it also accepts a stream/buffer object instead of a file path? Or leave it like it is.

ktbyers:
Yes, I think the stream-buffer idea would probably be good.

ktbyers:
Testing docs are here:

I will probably write the unit tests for it; it would be a good case for mock, but I don't have that setup.

jobec:
I've changed to code to also allow a subclass of `io.BufferedIOBase` to be passed in also.

Testing code:

### File
```python 

settings = {
    "device_type": 'juniper',
    "host": "test-device-1",
    "username": 'user',
    "password": 'password',
    'session_log': '/path/to/logfile.log'
}
with ConnectHandler(**settings) as device_conn:
    device_conn.send_command("show version")
```

### Buffer
```python 
import io
session_log = io.BytesIO()

settings = {
    "device_type": 'juniper',
    "host": "test-device-1",
    "username": 'user',
    "password": 'password',
    'session_log': session_log
}
with ConnectHandler(**settings) as device_conn:
    device_conn.send_command("show version")
    print(session_log.getvalue().decode())

# Outside the context manager we can still read the session
print(session_log.getvalue().decode())
```

jobec:
I did not test it with python 2.7 (yet)

jobec:
Rebased on ktbyers:develop and added python 2.7 support

jobec:
@ktbyers this should be good to go.

One thing maybe: it's writing bytes instead of strings. This isn't really a problem for my initial use case of writing to a file, but with the added functionality of writing to a buffer, user might expect that buffer to be string-based instead of bytes (like the return value of a `send_command()` call).

What do you think? Keep the bytes based approach? Or convert to strings? Usability-wise I think converting is better.

PR #857, opened 2018-06-30
Implement fast_mode for improved performance

ktbyers:
Netmiko is slow especially send_config_set. Try to create a way to improve performance (albeit with lower reliability).

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -38,7 +38,7 @@ class BaseConnection(object):
                  key_file=None, allow_agent=False, ssh_strict=False, system_host_keys=False,
                  alt_host_keys=False, alt_key_file='', ssh_config_file=None, timeout=90,
                  session_timeout=60, blocking_timeout=8, keepalive=0, default_enter=None,
-                 response_return=None, serial_settings=None):
+                 response_return=None, serial_settings=None, fast_cli=False):
         """
         Initialize attributes for establishing connection to target device.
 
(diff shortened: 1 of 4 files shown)


ktbyers:
Test cases
1. ACL 1-line
2. ACL 10-line
3. ACL 100-line
4. 1 x show command
5. 2 x show command
6. 10 x show command
----


Testing results:
-----

Cisco IOS

Baseline results (develop branch 2018-06-30)
1. 17.5 seconds
2. 22.0 seconds
3. 107.0 seconds 
4. 5.8 seconds
5. 6.5 seconds 
6. 11.2 seconds 

Fix to reduce send_config_set delay by factor of 10 (applies without any changes to end-user code)
1. 16.5 seconds
2. 17.0 seconds
3. 21.6 seconds

fast_cli enabled
1. 2.7 seconds
2. 2.8 seconds
3. 4.5 seconds
4. 1.6 seconds
5. 1.6 seconds
6. 2.3 seconds

Arista with fix to reduce send_config_set delay by factor of 10
1. 13.5 seconds
2. 13.4 seconds
3. Fail (exit_config_mode in send_config_set)

Arista fast_cli enabled
1. Fail
2. Fail
3. Fail
4. Fails in set base prompt
5. Fails in set base prompt
6. Fails in set base prompt

Note, this Arista virtual device is very slow.

ktbyers:
How to test, pass in `fast_cli = True` argument to ConnectHandler.

ktbyers:
The implications of this change are:

1. Default send_config_set() will reduce the delay between config command writes by a factor of 10 (to 50 milliseconds per command). Which I should have done earlier independent of this as it was way too slow...

2. If you pass the fast_cli=True argument to ConnectHandler then.

    a. global_delay_factor changes to the lesser of global_delay_factor and delay_factor. This implies global_delay_factor less than 1 should cause things to be faster.

    b. The default global_delay_factor goes to .1 which implies sleep times will be reduced by a factor of 10 (by default).

    c. Default timeout for send_command will also be reduced by a factor of 10 (from about 100 seconds to about 10 seconds).

    d. send_config_set() will have no delay between config commands on a write action i.e. it will write the config commands to the device as fast at it can.

PR #864, opened 2018-07-09
Session log updates

ktbyers:


ktbyers:
@jobec I made some minor changes to session_log here:

1. Added a mode option that defaulted to write so you could control whether was an append to the file or a write to the file.
2. Eliminated the file close() on disconnect() and added separate method to close the session_log. Basically, I didn't like that we were closing the file when it was a file, but not when it was a file handle.

I still need to add some unit tests. I did some manual testing on it.

ktbyers:
Let me think about your points some more.

ktbyers:
@jobec Okay, I agree with you (i.e. that the principle should be if Netmiko opens the file that we close the file).

Will add this back in.

ktbyers:
@jobec FYI, I think I am going to switch to a default mode where the session_log only contains the channel reads otherwise you get a duplication of output data in the session log (due to echo of commands).

For example here is the current output:

```

pynet-rtr1#

pynet-rtr1#terminal length 0
terminal length 0
pynet-rtr1#terminal width 511
terminal width 511
pynet-rtr1#

pynet-rtr1#show ip interface brief
show ip interface brief
Interface                  IP-Address      OK? Method Status                Protocol
FastEthernet0              unassigned      YES unset  down                  down    
FastEthernet1              unassigned      YES unset  down                  down    
FastEthernet2              unassigned      YES unset  down                  down    
FastEthernet3              unassigned      YES unset  down                  down    
FastEthernet4              10.220.88.20    YES NVRAM  up                    up      
Vlan1                      unassigned      YES unset  down                  down    
pynet-rtr1#

pynet-rtr1#exit
```

Here is what the read-only session log would look like:

```

pynet-rtr1#
pynet-rtr1#terminal length 0
pynet-rtr1#terminal width 511
pynet-rtr1#
pynet-rtr1#show ip interface brief
Interface                  IP-Address      OK? Method Status                Protocol
FastEthernet0              unassigned      YES unset  down                  down    
FastEthernet1              unassigned      YES unset  down                  down    
FastEthernet2              unassigned      YES unset  down                  down    
FastEthernet3              unassigned      YES unset  down                  down    
FastEthernet4              10.220.88.20    YES NVRAM  up                    up      
Vlan1                      unassigned      YES unset  down                  down    
pynet-rtr1#
pynet-rtr1#
```

I will probably have some sort of a `mode` flag for people that want both the reads and the writes.

ktbyers:
Unit tests should be done also.

Issue #890, opened 2018-07-25
change default_enter behavior for Aruba driver


BigPat:
In reference to "how to hit enter after a send_command #330" you offered a solution to respond to a router prompt.  I am having an issue in changing a password, when manually performing this it prompts for a Password: and for you to re-enter the password as seen below.  The solution that you offered did not work for me and after looking at the logs it looks to be sending the first password successfully but it it not reading the prompt for the second password.  

Any code examples or solutions that you can provide would be amazing.

**Manual Example:**
```
(exampledevice1) (config) #mgmt-user admin root              
Password:********
Re-Type password:********
```
**My Python Code:**

```python
net_connect.config_mode()

command = 'mgmt-user admin root'
aruba_pass = 'examplepassword'

configure4 = net_connect.send_command_timing(command)
        if 'Password:' in configure4:
            configure4 += net_connect.send_command_timing(aruba_pass)
        if 'password:' in configure4:
            configure4 += net_connect.send_command_timing(aruba_pass)

net_connect.exit_config_mode()

output2 = net_connect.save_config()
```

I have tried several different ways and it still wont work, any help would be greatly appreciated.

ktbyers:
How about this:

```python
net_connect.config_mode()

command = 'mgmt-user admin root'
aruba_pass = 'examplepassword'

output = net_connect.send_command_timing(command)
if 'Password:' in output:
    new_output = net_connect.send_command_timing(aruba_pass)

if 'Re-Type password' in new_output:
    net_connect.send_command_timing(aruba_pass)

net_connect.exit_config_mode()

output2 = net_connect.save_config()
```

ktbyers:
I didn't see an issue with what you originally posted besides indentation.

BigPat:
In my script the indentation is actually good, my copy and paste must have messed it up.  Oddly it doesnt work, I will post my whole script below.  Note I have a lot of commented out lines from other things I tried.

BigPat:
```
import netmiko
from datetime import datetime
from shutil import copy
import os
from os import listdir
from os.path import isfile, join
import logging
import time
import threading

run_date = datetime.today().date()
run_date = str(run_date)
run_time = datetime.now().time()
run_time = str(run_time)
run_time = run_time[:-7]
run_time = run_time.replace(':', '.')

logging.basicConfig(filename='Aruba_password_' + run_time + '.log', filemode='w', level=logging.DEBUG)

creds = []
[...]

PR #903, opened 2018-08-06
Patch for enter issue on IP Infusion

ktbyers:


ktbyers:
@mvsiva Does it work if we do this?

ktbyers:
Okay, bummer...

I will switch it back more like your original solution, I sort of wanted to avoid that as I didn't want to duplicate an almost exact replica of the `enable()` method...but it looks like that is what we probably need to do.

Although maybe I will code that as an optional argument directly into the parent method.

ktbyers:
@mvsiva Okay, I made updates to this. Can you test and see if this new fix works?

This also relates to this Aruba issue which has a very similar issue:

https://github.com/ktbyers/netmiko/issues/890

mvsiva:
@ktbyers Yes.. All tests for ssh and telnet passing now with both enable password being set and not set.. Thanks.

PR #880, opened 2018-07-19
Add support for ASA terminal width

ChristopherJHart:
Added code in ASA subclass that fixes #281 using the steps described within the issue.

Tested using a Cisco ASA 5508 running 9.4(4)20 with `terminal width 55` set in running-config. Below code was used to test:

```
from netmiko import ConnectHandler

asa_device = {"device_type": "cisco_asa", "ip": "10.122.160.155", "username": "admin", "password": "cisco!123"}
asa_conn = ConnectHandler(**asa_device)
print(asa_conn.send_command("show run | i terminal width"))
asa_conn.cleanup()
```
Output from the script:
```
C:\Users\chart2\Documents\Python\netmiko>python asa_test.py
terminal width 511
```

Output from the device after test script was run:
```
asa# show version | i Software|Hardware
Cisco Adaptive Security Appliance Software Version 9.4(4)20 
Hardware:   ASA5508, 8192 MB RAM, CPU Atom C2000 series 2000 MHz, 1 CPU (8 cores)
asa# show run | i terminal width
terminal width 55
```

The only possible bug I can see with the changes made in this PR is if configuration commands sent during the SSH session involve changing the terminal width. For example, let's say that an ASA starts with an initial terminal width of 80. A user uses netmiko to SSH into the ASA and change the terminal width to 450. Netmiko will change the terminal width to 511, store the previously-configured value of 80 as an attribute of the session, and then configure the terminal width to be 450 as the user requested. However, *if* the user calls the `cleanup()` method, the previously-configured value of 80 will be restored, thus overwriting the user's changes.

The obvious workaround to this issue would be to not call the `cleanup()` method at the end of a script, which I think we can all agree is less than desirable. Perhaps we can change the `send_command()` method (and all necessary related methods) to check to see if `terminal width XXX` is part of the commands sent, and if so, change `self.prev_term_width` to the user's defined value?

## Unit Test Results

```
C:\Users\chart2\Documents\Python\netmiko\tests>py.test -v test_netmiko_show.py --test_device cisco_asa_21
============================= test session starts =============================
platform win32 -- Python 3.6.4, pytest-3.6.3, py-1.5.4, pluggy-0.6.0 -- c:\users\chart2\appdata\local\programs\python\python36-32\python.exe
cachedir: ..\.pytest_cache
rootdir: C:\Users\chart2\Documents\Python\netmiko, inifile:
collected 12 items

test_netmiko_show.py::test_disable_paging PASSED                         [  8%]
[...]

PR #906, opened 2018-08-08
FIX: ASA terminal width issue

ktbyers:


ktbyers:
@ChristopherJHart FYI, I superseded your PR #880 with this one.

I didn't want to track state of the configuration change and I didn't want to automatically change the configuration.

So I added an argument you can pass that will allow this to happen.

```
self.allow_auto_change
```

The argument will default to false, but you can override it.

Netmiko won't track state of these terminal settings, however. 

So it will be the responsibility of the Netmiko user to restore state, if they need to (i.e. if you use this argument on the ASA and then perform a `wr mem`, you would need to set the terminal width to what you want it to be).

PR #908, opened 2018-08-09
Added support for jumphost directive in SSH config file

muffizone:


muffizone:
As the ProxyJump method is not available directly in paramiko , I had to convert the proxyjump directive into proxycommand

enzzzy:
The ProxyJump directive supports multiple proxies (comma separated), which it is going to use sequentially. I don't think this implementation would support that?

muffizone:
@enzzzy, that is correct.

I can modify the code to support multiple proxies

enzzzy:
You can also look at how Fabric implements support for ProxyJump 
https://github.com/fabric/fabric/blob/2.0/fabric/connection.py

muffizone:
Thanks for the link. I was planning something similar but this makes it easy

muffizone:
Hey @enzzzy,

I have the below code which works fine with multiple jumphosts. As you can see I am manually constructing the proxycommand instead of using the paramiko's connect method. Do you see any issues with manually constructing the command.

I have also removed the dependencey on netcat.

```
        elif 'proxyjump' in source:
            hops = list(reversed(source['proxyjump'].split(',')))
            source['proxycommand'] = 'ssh -W ' + source['hostname'] + ':' + source.get('port', '22') + ' ' + hops[0]
            proxy = paramiko.ProxyCommand(source['proxycommand'])
```

ktbyers:
Fixed merge conflicts in updated PR.

PR #911, opened 2018-08-09
Support directly using paramiko PKey objects for SSH key authentication

merlinthp:
This PR adds support for using a Paramiko PKey subclass (RSAKey etc) for key authentication, as an alternative to Paramiko extracting the key itself from a file.


ktbyers:
Replaced with:

https://github.com/ktbyers/netmiko/pull/929

Note, I renamed attribute to be consistently `pkey`

PR #924, opened 2018-08-20
Adding encoding for writing output data

ktbyers:


ktbyers:
@labulaka521 FYI, I am going to probably update your PR to be this way.

Basically, I am concerned whether writing UTF-8 is generally supported to network devices. I don't really have much data one way or another so am defaulting to doing what I am currently doing.

Yes, ASCII is a subset of UTF-8, but the question is what happens when you start sending non-ascii characters.

This will still give you a workaround for your problem, but you will have to specify the `write_encoding="utf-8"`.

Note, I am open to more data that network devices will generally support UTF-8 encoded bytes.

ktbyers:
I was writing this PR on airplane so I need to review it some more.

labulaka521:
thanks you very much @ktbyers

PR #931, opened 2018-08-27
Add blurb about TextFSM to README

tyler-8:
The main README doesn't mention the TextFSM integration so I just copy/pasted a bit of text from https://pynet.twb-tech.com/blog/automation/netmiko-textfsm.html and put it in the README as a means to get started quickly, with a link to the article for more detail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,76 +6,76 @@ Netmiko
 
 Multi-vendor library to simplify Paramiko SSH connections to network devices
 
-Python 2.7, 3.5, 3.6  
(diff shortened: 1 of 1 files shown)


tyler-8:
Seems my editor didn't like the trailing whitespace in the README. I can revert that part if necessary.

ktbyers:
@tyler-8 Yes, if you could fix the trailing whitespace...that would be great.

We need the trailing whitespace so that the Markdown renders it as a newline (otherwise it just wraps it all on the same line).

PR #932, opened 2018-08-27
Raising exception for insanely long passwords

whitej6:
Cisco WLCs support a maximum password length of 25 characters.  This change will prevent the script from continuing into the self.find_prompt method and crashing when the prompt cannot be found.

### DIFF
--- a/netmiko/cisco/cisco_wlc_ssh.py
+++ b/netmiko/cisco/cisco_wlc_ssh.py
@@ -24,6 +24,8 @@ class CiscoWlcSSH(BaseConnection):
         Password:****
         """
         password_length = len(self.password)
+        if password_length > 25:
+            raise ValueError('The supplied password of [{}] is greater than the maximum supported length of [25] for Cisco WLC.'.format(str(lenPassword)))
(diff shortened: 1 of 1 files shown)


whitej6:
Here is the link that confirmed suspicions with password limitations http://www.routereflector.com/2016/08/maximum-username-password-length-on-cisco-wlc-and-radius/

ktbyers:
Can you explain why this would be on the Netmiko side? I would have thought this should be on the end-user side (i.e. the person that is writing the code using Netmiko).

Also I am worried that Cisco will change this in the future and then Netmiko code will break (i.e. this is the type of fix that in the long-run causes problems).

whitej6:
I had initially submitted this with a bug and closed this one to fix and resubmit.  I am going to closed the newly submitted pull request and redo this to raise an authentication error.  Currently the cisco_wlc driver does not account for authentication errors in a user readable error.  The current error when an incorrect password is received is ValueError: Router prompt not found: 'User:', this isn't the first time I have dealt with this error to later find a locked out account or as in today the account password is too long for a WLC to pass on the authentication request to the AAA server.

PR #934, opened 2018-08-27
Adding cisco_wlc ability to return auth exception

whitej6:
Previously cisco_wlc did not have a built in method of returning an authentication exception if incorrect credentials were provided.  It would raise ValueError: Router prompt not found: 'User:' is misleading if the issue is due to an incorrect login credentials.  This will now return an exception only if the password prompt occurs but then returns back to the user prompt.


OzNetNerd:
@ktbyers , this PR can be closed due to https://github.com/ktbyers/netmiko/pull/1107 being merged.

ktbyers:
Agreed--I don't think is really necessary given #1107

PR #939, opened 2018-08-30
Create send_command_multiline and add list support to send_command_timing

ktbyers:


ktbyers:
Issues:
1. Only gathers output until first match of the prompt.
2. Wouldn't handle in any backspace characters in command echo.
3. TextFSM won't work

carlniger:
@ktbyers do you still want to have this functionality? I feel like I maybe heard you say that it was causing more trouble than it's worth?

ktbyers:
I am just going to close...probably best to revisit it at a later date.

Issue #922, opened 2018-08-16
Option to skip verify space for SCP function


TomCos:
Was wondering the LoE to add an option to skip the verify space function being automatically run when the file_transfer function is called.  Reason being is, a lot of devices (mostly terminal servers I'm seeing) that are backend unix/linux, they mount their temporary filesystems straight from memory, not from a disk.  The verify space is getting false data because the filesystem it's checking is returning all 0's for space available, used, blocks, etc.
Thoughts?

ktbyers:
@TomCos Can you show an example? What does 'df -k' show for that file system. I wouldn't think it would matter if it a memory file-system or a disk file-system?

The general process would be to subclass and then implement a new remote_space_available method similar to what was done here:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/juniper/juniper.py#L209-L238


Also here is what is called:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/juniper/juniper.py#L209-L238

ktbyers:
Strange...is that actually a proper implementation of a tmpfs (I would have thought it still needed to report its size and space available).

Not that it helps you...

```
df -k
Filesystem     1K-blocks    Used Available Use% Mounted on
/dev/xvda1      16382888 4455636  11827008  28% /
devtmpfs          769336      16    769320   1% /dev
tmpfs             786164       0    786164   0% /dev/shm
```

We could probably add some sort of a force option or skip_space_check. Someone has to do the PR on it though.

TomCos:
I've opened a ticket with the vendor about how their tmpfs is implemented, I agree it's not correct.  For now, I'll just call the functions directly to avoid the check.  Thanks!

ktbyers:
Okay, sounds good. I will just close this as I am not seeing a further action on my part. But re-open if there is more to do.

BenjaminHoegh:
what is the status on this?

BenjaminHoegh:
what's the status on this? has the option to skip verify space for SCP been implemented yet?

ktbyers:
@BenjaminHoegh I don't think this was ever done. The SCP primitives are all there so you could do it yourself.

What is the underlying issue you ran into?

BenjaminHoegh:
> @BenjaminHoegh I don't think this was ever done. The SCP primitives are all there so you could do it yourself.
> 
> What is the underlying issue you ran into?

For me the issue is that I do not have permission to use `df` or `/bin/df` on the target server

PR #947, opened 2018-09-08
Add option for progress display during SCP

TomCos:
Just pushing the progress callback at SCP object establish.

### DIFF
--- a/netmiko/arista/arista.py
+++ b/netmiko/arista/arista.py
@@ -56,7 +56,7 @@ class AristaTelnet(AristaBase):
 
 class AristaFileTransfer(CiscoFileTransfer):
     """Arista SCP File Transfer driver."""
-    def __init__(self, ssh_conn, source_file, dest_file, file_system="/mnt/flash", direction='put'):
+    def __init__(self, ssh_conn, source_file, dest_file, file_system="/mnt/flash", direction='put', progress=None):
         return super(AristaFileTransfer, self).__init__(ssh_conn=ssh_conn,
                                                         source_file=source_file,
                                                         dest_file=dest_file,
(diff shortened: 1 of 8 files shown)


ktbyers:
It looks like you would also need to add the `progress` function somewhere (i.e. that function/method needs to exist).

https://stackoverflow.com/questions/49657776/python-scpclient-copy-progress-check

You can fix or leave this PR open and I can potentially fix.


Regards, Kirk

TomCos:
I don't think the function needs to exist if the variable defaults to None, as it also does in the SCPClient code (https://github.com/jbardin/scp.py/blob/master/scp.py) and my merge is based off his examples on the readme page (https://github.com/jbardin/scp.py).

My thought process was, there are probably 1000 different ways people are going to want the data displayed, from print, to logging, writing to files, etc, all different formats.  I figured leave it as None and let the users make their own progress function displays.

Saw the errors from the build, just fixed them.

TomCos:
It's also worth discussing, my implementation just into the establish_scp_conn function is because of the issues I was facing here (https://github.com/ktbyers/netmiko/issues/922), so I was skipping the file_transfer class and just calling the function in FileTransfer directly.  Thinking in the context of a normal implementation, the progress variable is going to need to be carried through BaseFileTransfer as well, correct?

ktbyers:
You still have some minor lint errors that you need to fix here.

bradh11:
In reviewing the paramiko implementation it seems to have a function called progress_tracker which is a member of the SCPClient class.  This seems to be related to what you are trying to do. 

‘’’
def _progress_tracker(self, method, basename, size, file_pos, peername):
        #count number of arguments
        count = method.__code__.co_argcount
        if count == 3:
            method(basename, size, file_pos)
        elif count == 4:
            method(basename, size, file_pos, peername)
        else:
            #do not break, just skip
            pass
‘’’

ktbyers:
This PR looks like it was never fully implemented so I am going to close. Feel free to re-open/re-create if you want to re-implement it.

assadniang:
Was wondering if this has been implemented in netmiko scp function, and how to use it.

ktbyers:
No, there never was a working implementation of this built.

PR #949, opened 2018-09-11
add support to detect an F5 and Linux system

TomCos:
add support to detect an F5 system when the the user logs into the tmsh shell directly


TomCos:
After some research, 'uname -a' is implemented on almost every Unix based system.  It's safe to say that if "Linux" is detected in the output, you're in a Linux shell, and that driver can be used.  The only caveat I can think of is the OVS Linux implementation, which after opening its class file, it just imports the Linux one, so we should be good 👍

PR #953, opened 2018-09-15
Telnet is_alive() improvements

ktbyers:


ktbyers:
I think this is better than what we previously had, but it still doesn't really work very well.

closed connection by sending "exit\n" from Netmiko. Would get into a close_wait state at the TCP level is_alive() would still return True.

closed connection by clearing the line on the router. Once again close_wait state at the TCP level and is_alive() for telnet would report true.

I am inclined to deprecate this is_alive both for SSH and telnet, as it is very hard to make it work in a reasonable way.

PR #963, opened 2018-09-27
Close connection on prompt read failure

roman-verchikov:
Resolves the issue observed with Dell OS10 switches.

In our environment we're using code, which can be simplified to this:

```

while 1:
    try:
        dev = ConnectHandler(
            device_type='dell_os10',
            ip=<ip>,
            username=<username>,
            password=<password>)
        dev.send_command(<some command>)
     except:
         LOG.critical("something bad happened, retrying")
         time.sleep(5)
```

For some OS10 devices, prompt gets shown with a significant delay, in
our particular case, more than 3 seconds. This delay results in empty
prompt returned by `BaseConnection.find_prompt()`, which in turn
results in `ValueError` raised. This error is not handled by
`BaseConnection` constructor and happens right after
`BaseConnection.establish_connection()` is called, which for the code
above results in endlessly growing number of SSH connections to an OS10
device, which at some point exhausts allowed number of SSH connections,
and noone can access device over SSH unless script above is stopped
which cleans up all the connections.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -239,12 +239,12 @@ class BaseConnection(object):
             self.protocol = 'telnet'
             self._modify_connection_params()
             self.establish_connection()
-            self.session_preparation()
+            self._safe_session_preparation()
         elif '_serial' in device_type:
             self.protocol = 'serial'
             self._modify_connection_params()
             self.establish_connection()
-            self.session_preparation()
+            self._safe_session_preparation()
         else:
             self.protocol = 'ssh'
 
@@ -268,7 +268,7 @@ class BaseConnection(object):
 
             self._modify_connection_params()
             self.establish_connection()
-            self.session_preparation()
+            self._safe_session_preparation()
 
     def __enter__(self):
         """Establish a session using a Context Manager."""
@@ -638,6 +638,17 @@ class BaseConnection(object):
         msg = "Login failed: {}".format(self.host)
         self.remote_conn.close()
         raise NetMikoAuthenticationException(msg)
+
+    def _safe_session_preparation(self):
+        # `session_preparation()` may fail to retrieve command prompt, and raises
+        # ValueError, which results in connection not being closed in cases
+        # when context manager cannot be used. Make sure connection is always closed
+        # in those cases.
+        try:
+            self.session_preparation()
+        except Exception:
+            self.disconnect()
+            raise
 
     def session_preparation(self):
         """


ktbyers:
Have you tried setting `global_delay_factor=2` (or 4) as a ConnectHandler argument and see if that fixes your find_prompt problem.

The PR won't work as the current behavior is better (i.e. generate the exception associated with the thing that fails). 

If you want to ensure a clean disconnect on failure then use the Netmiko context manager (i.e. use a with statement).


Regards, Kirk

roman-verchikov:
I tried `global_delay_factor=5` and it did solve the issue, but it's a workaround and not a solution to the problem, because it will still fail if prompt takes even longer to load.

Please reopen and consider merging.

The issue with existing code is that it doesn't close established connection, and in the example above it exhausts SSH connection limit, so no one can SSH into device.

roman-verchikov:
Also, this patch doesn't change existing behavior, exception is still raised, but it doesn't leave established connection behind.

ktbyers:
Why does the prompt have so much delay and variation in delay?

Good point on the exception (i.e. that you just re-raise the current exception), but why not just use the context manager? 

```python
with ConnectHandler(**device) as conn:
    my_prompt = conn.find_prompt()
```

Also your fix won't work in general as the session_preparation method is generally overridden by each driver (i.e. the change would be much bigger than what you did). But I don't think this is the right solution.

If it is a problem with the dell_os10 driver and needing more delay in general during session_preparation, then we should put a fix in there (including potentially creating a specific session_preparation method for that driver).

roman-verchikov:
Unfortunately context manager won't help, since we have an abstraction layer on top of Netmiko and other libraries to connect to different devices and run commands there, and based on our specific use-case this abstraction cannot use context manager, instead it opens session on service start, and runs all commands inside that session. The service also has logic which would try to re-open connection if session closed for any reason (simplified in the `while 1:` loop in PR description), and since `ConnectHandler()` class raises exception inside constructor, and keeps established connection open, this loop leads to endless number of connections open, until the service is killed and all the connection objects are destroyed.

> Also your fix won't work in general as the session_preparation method is generally overridden by each driver (i.e. the change would be much bigger than what you did). But I don't think this is the right solution.

Ok, I can change that by introducing a wrapper around `session_preparation()`, something like:

```
def try_open_session():
    try:
         self.session_preparation()
    except ValueError:
         self.disconnect()
         raise
```

... and then also replace all `session_preparation()` calls with `try_open_session()`. This will solve the issue, and will not require any changes in derived classes.

The issue is not specific to Dell OS10, according to the code. Any failed `session_preparation()` would result in the same behavior, the only difference would be in how the transport protocol is implemented. In Dell case it results in SSH connection limit being exhausted on server side (device), not sure what happens for other backends, but most likely something very similar, since many drivers use SSH as transport protocol.

The issue is pretty significant in our use case, and it doesn't seem there is a better solution for Netmiko client code.

Please consider reopening, I'm ready to update the code to make sure it satisfies all the requirements.

roman-verchikov:
> Why does the prompt have so much delay and variation in delay?

I don't know. I suppose there's some long-running scripts that are executed on user login. But for the client it shouldn't matter at all, the library should handle that issue properly, and clean up all the internally-create resources on failure, which does not happen in this particular case.

ktbyers:
Let me know if the above fix works for you?

 I am going to re-open this so I don't forget about the issue.

roman-verchikov:
> Let me know if the above fix works for you?

Yes, it does. The only thing is that you'd also have to update the same logic for `_telnet` and `_serial`, that's why I suggested to isolate it in a separate method `try_open_session()` and replace all calls to `session_preparation()` with `try_open_session()`. This way the logic would be isolated and potentially would be easier to modify, since it's defined in a single place, not many.

ktbyers:
Sounds good. Let's call it `_session_prep_wrapper()` or something similar to that.

roman-verchikov:
@ktbyers updated. Please take a look. Thanks!

PR #966, opened 2018-10-02
F5 tmsh fix

TomCos:
add support for f5 bash and rename ltm to tmsh for proper naming convention.  ltm is one of their products


ktbyers:
I am going to keep the old name in ssh_dispatcher.py for backwards compatibility.

TomCos:
Yeah that's a good idea, we should have f5_ltm map to the same class as f5_tmsh

PR #967, opened 2018-10-02
add support for avocent acs and opengear console servers

TomCos:


TomCos:
The conflict caused a lot of issues...I'll re-commit

PR #970, opened 2018-10-03
Re-commit of the acs and opengear devices

TomCos:


TomCos:
@ktbyers any issues with this one?

TomCos:
@ktbyers just following up again

ktbyers:
Sorry for the delay...so I don't really like making extra additional copies of the "Linux" driver as it seems like you could easily get a bunch of "Linux" drivers that are all identical.

So I would vote we get rid of the extra drivers and keep the autodetect i.e. keep the SSH autodetect for the these platforms, but just have it pick the standard Linux driver (and probably add a comment in the SSH autodetect for which platform it is for).

Does that sound reasonable?

TomCos:
The idea was that if we expand the Avocent or OpenGear drivers to be more than just the import.  What I put into the autodetect was just the bare minimum to allow that.  In the future, the class for the driver might contain more than just to "pass," thinking this solution is more forward thinking.

ktbyers:
We should cross that bridge when we get there (or if we already know there are things required for these drivers implement those things now).

There is also a generic "terminal_server" device type. On second thought, these probably should be grouped under this for auto-detection.

I am pretty reluctant to build/maintain drivers for both Linux variants and for Terminal Server variants. In both cases, there are a lot of them. In the terminal server case there are a lot of idiosyncrasies. Also I don't think this is what Netmiko is primarily for.

So for terminal servers the "terminal_server" device type should give Netmiko users a way to use Netmiko with almost any terminal server, but pushes the terminal server interface management back on the user. In other words, they have to handle the special commands, menus, logins, and then call redispatch once they actually connected to their end-device (i.e. switch to the right Netmiko class for the end-device they are reaching through the terminal server).

TomCos:
So maybe I'm missing something here.  The terminal server driver in your documentation is supposed to be used to navigate a generic terminal server console to get to a specific port that has another OS on it.  In the case of the ACS and Opengear devices, these are the actual network appliances, before we navigate to a port, and potentially remap, so I'm not sure the example lines up.

From a users perspective, shouldn't we be looking to expanding Netmiko to support as many network vendors as possible.  The code of importing the Linux driver came about from the OVS one, which I did not create, so I assume there was a previous approval for that one as well?  The f5_linux one follows the same format.

I don't think there is any maintaining needed either, with all of the OS's being linux and if anything needs to change, just overriding it within the object itself, which is why I still think best practice would be to keep it separate.

ktbyers:
@TomCos 

Yes, I probably shouldn't have created the OVS driver or the F5 Linux driver. 

I don't want a proliferation of either Linux drivers or Terminal Server drivers that are identical. It just causes an unnecessary duplication of classes. In both cases, it is for a purpose that is fairly tangential to netmiko.

Remember that all the code that is added is code that someone has to maintain.

Terminal server support isn't a high-priority for me (and we do have a way of supporting them--it just pushes most of the interfacing to them back on the library user).

Terminal servers strike me as a pretty fragmented market (definitely an assumption on my part as I haven't looked at the data) with idiosyncrasies in interfacing to them. Additionally, you usually have special considerations in handling whatever they are connected (meaning even if you work out interfacing to the terminal server, the end-user would still probably have issues programmatically handling the login to the end device).

Netmiko really isn't intended to support Linux. Linux support is mostly a secondary consideration largely for SSH proxy and for testing. 

Didn't really understand your last comment on why creating separate drivers for these two items would be a best practice?

ktbyers:
@TomCos Thanks for the contribution.

I am going to close this one--I don't think we want to make unnecessary duplicates of the Linux driver.


Regards,

Kirk

PR #979, opened 2018-10-15
Telnet return update

TomCos:
Have the return character assigned to the telnet return if it's a telnet session.  I've tested it with success against a couple telnet devices.  Per the RFC (https://tools.ietf.org/html/rfc318) this should always be the case if we're connecting to a device using telnet, so best practice.

I left the hard-coded one in play for backwards compatibility.


netmiko/base_connection.py
```diff
@@ -188,8 +188,16 @@ def __init__(
         :type encoding: str
         """
         self.remote_conn = None
-        self.RETURN = "\n" if default_enter is None else default_enter
-        self.TELNET_RETURN = "\r\n"
+
+        self.TELNET_RETURN = '\r\n'
+        if default_enter is None:
+            if not 'telnet' in device_type:
```
ktbyers: There is a minor linting issue here in that the linter wants this to be 'if 'telnet' not in device_type (that is why the tests are failing).

TomCos:
Should be good now, let me know.

Issue #673, opened 2018-01-10
Improve Handling of Uncommon Prompt Terminators


tyler-8:
There should be a way to customize or look for additional prompt terminators other than `#` and `$`, particularly for Linux driver as it's possible to customize the terminator on most Linux machines.

Line in question: https://github.com/ktbyers/netmiko/blob/develop/netmiko/linux/linux_ssh.py#L22

ktbyers:
The following environment variables were just added for Linux:

```
LINUX_PROMPT_PRI = os.getenv("NETMIKO_LINUX_PROMPT_PRI", "$")
LINUX_PROMPT_ALT = os.getenv("NETMIKO_LINUX_PROMPT_ALT", "#")
LINUX_PROMPT_ROOT = os.getenv("NETMIKO_LINUX_PROMPT_ROOT", "#")
```

Based on the PR #984 that you submitted.

PR #984, opened 2018-10-22
Add environment variable option to set custom prompt terminators for linux ssh

tyler-8:
This allows you to use environment variables to override the default prompt terminators in the event your Linux server doesn't use the `$` or `#` characters. If no environment variables are set, it defaults to the original terminators.

Addresses https://github.com/ktbyers/netmiko/issues/673


ktbyers:
Minor updates here:

https://github.com/ktbyers/netmiko/pull/1338


Sorry for the delay :-(

PR #985, opened 2018-10-24
Add support for Endace devices

je-clark:
Added support for Endace packet capture devices. This has been tested in both development and production environments.


ktbyers:
Updated PR for Endace here:

https://github.com/ktbyers/netmiko/pull/1266

PR #991, opened 2018-10-26
Ruckus FastIron/ICX telnet truncation fix

wonderbred:
`strip_command()` from `BaseConnection()` class doesn't need to remove the command string from `send_command()` output when connecting to Ruckus FastIron via telnet. Currently strips characters from beginning of the output even though command string is not in output.

### DIFF
--- a/netmiko/ruckus/ruckus_fastiron.py
+++ b/netmiko/ruckus/ruckus_fastiron.py
@@ -57,7 +57,7 @@ class RuckusFastironBase(CiscoSSHConnection):
             )
             raise ValueError(msg)
 
-    def save_config(self, cmd="write mem", confirm=False):
+    def save_config(self, cmd="write mem", confirm=False, confirm_response=""):
         """Saves configuration."""
         return super(RuckusFastironBase, self).save_config(cmd=cmd, confirm=confirm)
 
--- a/netmiko/ruckus/ruckus_fastiron.py
+++ b/netmiko/ruckus/ruckus_fastiron.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 import re
 import time
+from telnetlib import DO, DONT, ECHO, IAC, WILL, WONT
 from netmiko.cisco_base_connection import CiscoSSHConnection
 
 
@@ -68,26 +69,23 @@ class RuckusFastironTelnet(RuckusFastironBase):
         kwargs["default_enter"] = "\r\n" if default_enter is None else default_enter
         super(RuckusFastironTelnet, self).__init__(*args, **kwargs)
 
-    def strip_command(self, command_string, output):
+    def _process_option(self, tsocket, command, option):
         """
-        Modified strip_command for Ruckus FastIron telnet connections
-
-        :param command_string: The command string sent to the device
-        :type command_string: str
-
-        :param output: The returned output as a result of the command string sent to the device
-        :type output: str
+        Ruckus FastIron/ICX does not always echo commands to output by default.
+        If server expresses interest in 'ECHO' option, then reply back with 'DO
+        ECHO'
         """
-        backspace_char = "\x08"
+        if option == ECHO:
+            tsocket.sendall(IAC + DO + ECHO)
+        elif command in (DO, DONT):
+            tsocket.sendall(IAC + WONT + option)
+        elif command in (WILL, WONT):
+            tsocket.sendall(IAC + DONT + option)
 
-        # Check for line wrap (remove backspaces)
-        if backspace_char in output:
-            output = output.replace(backspace_char, "")
-            output_lines = output.split(self.RESPONSE_RETURN)
-            new_output = output_lines[1:]
-            return self.RESPONSE_RETURN.join(new_output)
-        else:
-            return output
+    def telnet_login(self, *args, **kwargs):
+        # set callback function to handle telnet options.
+        self.remote_conn.set_option_negotiation_callback(self._process_option)
+        return super(RuckusFastironTelnet, self).telnet_login(*args, **kwargs)
 
 
 class RuckusFastironSSH(RuckusFastironBase):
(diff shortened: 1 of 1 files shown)


ktbyers:
This doesn't seem right? In other words, it seems like we are just masking the problem for some reason (i.e. why is our character count wrong on the telnet driver for ruckus). For the number of characters to strip?

Can you enable the `session_log` and post that data here (for your original failing send_command operation)?

```
icx_telnet = Netmiko(
  host='172.22.16.50',
  device_type='ruckus_fastiron_telnet',
  session_log='output_file.txt',
)
```

This will create a file in your current working directory named output_file.txt containing details about what happened during the session.

Also why is there no username and password above?



Kirk

wonderbred:
> This doesn't seem right? In other words, it seems like we are just masking the problem for some reason (i.e. why is our character count wrong on the telnet driver for ruckus). For the number of characters to strip?

The character count is off by exactly the length of the command because the ICX doesn't echo the command in the output for telnet connections. 


> Can you enable the `session_log` and post that data here (for your original failing send_command operation)?

[output_file.txt](https://github.com/ktbyers/netmiko/files/2567455/output_file.txt)


```
... icx_telnet = Netmiko(
...   host='172.22.16.50',
...   device_type='ruckus_fastiron_telnet',
...   session_log='output_file.txt',
... )
... 
... output = icx_telnet.send_command('show flash')
... 
... print(output)
... 
1:
  NAND Type: Micron NAND 2GiB (x 1)
  Compressed Pri Code size = 25713184, Version:08.0.70cT211 (SPS08070c.bin)
  Compressed Sec Code size = 29437672, Version:08.0.70cT213 (SPR08070c.bin)
  Compressed Boot-Monitor Image size = 786432, Version:10.1.13T225
  Code Flash Free Space = 1319170048
```

> Also why is there no username and password above? 

Because I'm using this for zero-touch provisioning of brand new ICX switches. They ship from the factory with telnet enabled and no username/password.

ktbyers:
So when you manually telnet into the device and type "show flash" you don't actually see the "show flash" on your telnet terminal?

ktbyers:
I think the solution will probably be turn on the normal echo behavior which will probably be a pattern similar to what we did here:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/ipinfusion/ipinfusion_ocnos.py#L40

Except using options for telling the remote device to do the character echo.

This will help main consistent behavior with other netmiko drivers (including the SSH driver for this platform).

wonderbred:
> So when you manually telnet into the device and type "show flash" you don't actually see the "show flash" on your telnet terminal?

That's the strange thing -- I do see the command when I type it in a telnet session from Terminal or SecureCRT:

```
➜  ~ telnet 172.22.16.50
Trying 172.22.16.50...
Connected to 172.22.16.50.
Escape character is '^]'.
telnet@ICX7150-48P Switch>en
No password has been assigned yet...
telnet@ICX7150-48P Switch#
telnet@ICX7150-48P Switch#show flash
Stack unit 1:
  NAND Type: Micron NAND 2GiB (x 1)
  Compressed Pri Code size = 25713184, Version:08.0.70cT211 (SPS08070c.bin)
  Compressed Sec Code size = 29437672, Version:08.0.70cT213 (SPR08070c.bin)
  Compressed Boot-Monitor Image size = 786432, Version:10.1.13T225
  Code Flash Free Space = 1319170048
telnet@ICX7150-48P Switch#
```

I don't know why it's not echoed in the Netmiko session.

wonderbred:
> I think the solution will probably be turn on the normal echo behavior which will probably be a pattern similar to what we did here:
> 
> https://github.com/ktbyers/netmiko/blob/develop/netmiko/ipinfusion/ipinfusion_ocnos.py#L40
> 
> Except using options for telling the remote device to do the character echo.
> 
> This will help main consistent behavior with other netmiko drivers (including the SSH driver for this platform).

I can take a stab at this today.

ktbyers:
Sounds good...this issue is a bit strange.

wonderbred:
> I think the solution will probably be turn on the normal echo behavior which will probably be a pattern similar to what we did here:
> 
> https://github.com/ktbyers/netmiko/blob/develop/netmiko/ipinfusion/ipinfusion_ocnos.py#L40
> 
> Except using options for telling the remote device to do the character echo.
> 
> This will help main consistent behavior with other netmiko drivers (including the SSH driver for this platform).

Maybe I'm missing something, but I don't see how the above example applies to what needs to be changed with the Ruckus driver. How does overriding `telnet_login()` fix an issue related to commands [not] echoing after we've already successfully logged into the device?

Instead of overriding the `strip_command()` method, I _did_ consider overriding all of the methods that use the `strip_command` keyword argument and setting the default to `strip_command=False`, but that means overriding considerably more code and it seems like that could potentially break in the future.

At its most basic level, the problem is that Netmiko expects telnet commands to be echoed back to the terminal, but they aren't. If the command is 12 characters long, `strip_command()` removes 12 characters from the beginning of the command's output. And since the default behavior for `send_command()` and `send_command_timing()` is `strip_command=True`, every single telnet command's output is getting truncated.

So if `strip_command` will never work as intended for FastIron telnet connections, why not just override it?

ktbyers:
@wonderbred Telnet (generally) always echoes the commands and it echos the commands for all other platforms that Netmiko connects to...so the more general question is why is it not echoing on this platform (and not only this platform, but this platform and Netmiko).

So I don't want to change strip_command() as I think we are just hiding a different problem (i.e. the echoing of characters on telnet for some specific subset of fastiron_ruckus). And I have reason to believe that this works as-is for some other individual(s) on ruckus_fastiron.

So what I linked to had an example of communicating telnet options during telnet connection. Basically you can communicate telnet control options via the telnet channel itself. What I linked to contained an example of sending telnet options down the telnet channel (in the process_options method).

Does that make sense?

```
    def _process_option(self, tsocket, command, option):
        """
        For all telnet options, re-implement the default telnetlib behaviour
        and refuse to handle any options. If the server expresses interest in
        'terminal type' option, then reply back with 'xterm' terminal type.
        """
        if command == DO and option == TTYPE:
            tsocket.sendall(IAC + WILL + TTYPE)
            tsocket.sendall(IAC + SB + TTYPE + b"\0" + b"xterm" + IAC + SE)
        elif command in (DO, DONT):
            tsocket.sendall(IAC + WONT + option)
        elif command in (WILL, WONT):
            tsocket.sendall(IAC + DONT + option)
```

So we would need to figure out how to communicate down the telnet channel to enable echo (which should be the default behavior anyways...which is the strange part).

Does that help?

PR #992, opened 2018-10-27
EltexEsrSSH created

Uburro:
Created a module for working with routers Eltex ESR


Uburro:
@ktbyers this is a router. ESR has its own software that differs from other Eltex devices

ktbyers:
Updated this PR here (fixing the merge conflict):

https://github.com/ktbyers/netmiko/pull/1376

PR #996, opened 2018-10-30
Add generic prompt pattern for check_config_mode

glennmatthews:
Somewhat analogous to changes made in #820.

By providing a simple default prompt pattern in `check_config_mode`, it can typically return much faster than when no pattern was specified. As calling `send_config_set()` results in calling `check_config_mode()` four times, this change correspondingly makes `send_config_set()` that much faster - in my testing, with simple config CLI, as much as 65% faster.

### DIFF
--- a/netmiko/cisco/cisco_ios.py
+++ b/netmiko/cisco/cisco_ios.py
@@ -22,6 +22,16 @@ class CiscoIosBase(CiscoBaseConnection):
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()
 
+    def check_config_mode(self, check_string=")#", pattern="#"):
+        """
+        Checks if the device is in configuration mode or not.
+
+        Cisco IOS devices abbreviate the prompt at 20 chars in config mode
+        """
+        return super(CiscoIosBase, self).check_config_mode(
+            check_string=check_string, pattern=pattern
+        )
+
     def save_config(self, cmd="write mem", confirm=False):
         """Saves Config Using Copy Run Start"""
         return super(CiscoIosBase, self).save_config(cmd=cmd, confirm=confirm)
--- a/netmiko/cisco/cisco_nxos_ssh.py
+++ b/netmiko/cisco/cisco_nxos_ssh.py
@@ -23,6 +23,12 @@ class CiscoNxosSSH(CiscoSSHConnection):
         newline = re.compile(r"(\r\r\n|\r\n)")
         return newline.sub(self.RESPONSE_RETURN, a_string).replace("\r", "")
 
+    def check_config_mode(self, check_string=")#", pattern="#"):
+        """Checks if the device is in configuration mode or not."""
+        return super(CiscoNxosSSH, self).check_config_mode(
+            check_string=check_string, pattern=pattern
+        )
+
 
 class CiscoNxosFileTransfer(CiscoFileTransfer):
     """Cisco NXOS SCP File Transfer driver."""


ktbyers:
This CiscoSSHConnection class is used by a lot of platforms (close to 40) so I am worried that checking on the '#' as opposed to using the current timing mechanism might have unanticipated side-effects (and given the limited number of platforms I have to test it is hard for me to ensure it doesn't break things).

Consequently, I am probably not going to implement this change.

If you only need it for a given Cisco platform then we could look to implement it in that specific platform potentially.

You might want to use fast_cli=True to improve the performance.

jcinma:
It might be worthwhile to move the same change over to /netmiko/cisco/cisco_ios.py. Same performance impact, but it applies to and will be used by fewer devices.

yoonghm:
Just an observation.  When we login to a network device, the default prompt is unprivileged. Once it enters into privileged mode, the prompt change. Once it enters into config mode, the prompt change again. Could we use this behaviour to detect different modes to the response faster?

ktbyers:
@jcinma Yes, we could apply this change to the cisco_ios driver that should work.

The issue is that the change presupposes that the returned prompt contains a '#' which may or may not be true. This Cisco base case is used by all of these devices:

```
$ grep -l CiscoSSHConnection `find . -name '*.py'`
./vyos/vyos_ssh.py
./eltex/eltex_ssh.py
./quanta/quanta_mesh_ssh.py
./fortinet/fortinet_ssh.py
./enterasys/enterasys_ssh.py
./coriant/coriant_ssh.py
./cisco_base_connection.py
./mellanox/mellanox_ssh.py
./ruckus/ruckus_fastiron.py
./linux/linux_ssh.py
./calix/calix_b6.py
./ciena/ciena_saos_ssh.py
./ubiquiti/edge_ssh.py
./hp/hp_procurve.py
./hp/hp_comware.py
./extreme/extreme_wing_ssh.py
./extreme/extreme_ers_ssh.py
./extreme/extreme_exos.py
./extreme/extreme_nos_ssh.py
./extreme/extreme_netiron.py
./extreme/extreme_slx_ssh.py
./extreme/extreme_vsp_ssh.py
./accedian/accedian_ssh.py
./arista/arista.py
./alcatel/alcatel_aos_ssh.py
./alcatel/alcatel_sros_ssh.py
./huawei/huawei_ssh.py
./mrv/mrv_ssh.py
./a10/a10_ssh.py
./dell/dell_os10_ssh.py
./dell/dell_force10_ssh.py
./aruba/aruba_ssh.py
./cisco/cisco_s300.py
./cisco/cisco_tp_tcce.py
./cisco/cisco_asa_ssh.py
./cisco/cisco_nxos_ssh.py
./cisco/cisco_xr_ssh.py
./apresia/apresia_aeos.py
```

We might be able to work it out by looking at the check_string for each type of device and seeing if they all contain a '#' in their check string, but we would have to go through them one-by-one.

I also am leery of this change as I broke one of the drivers (Huawei driver) by making a similar change recently. 

Since we don't have a direct way to test a large set of those platforms, it becomes hard to determine whether the change will break things (for performance improvements which you probably get in other ways i.e. by using fast_cli=True).

glennmatthews:
So, FWIW, I tried using `fast_cli=True`, but found that the change in behavior of send_config_set (sending all lines of config as rapidly as possible) caused undesirable behavior in my testing. Hence my attempt to improve the performance by this method instead. 

Netmiko is already using a prompt `pattern` of '#' in `CiscoBaseConnection.exit_config_mode`, so I'd hoped that using it in `check_config_mode` as well would be a low-risk change. It sounds like that isn't the case?

ktbyers:
@glennmatthews Can you expand on what you saw on fast_cli=True (i.e. what issue you ran into)? So I have some feedback on this.

```
Netmiko is already using a prompt pattern of '#' in CiscoBaseConnection.exit_config_mode, so I'd hoped that using it in check_config_mode as well would be a low-risk change. It sounds like that isn't the case?
```

Yes, that is not really the case as any method can be overridden by the child-class so you would have to review what is being done in each child-class. It might not be an issue, but someone would have to go through the 38 classes/modules and check. And I think that exit_config_mode change possibly broke the Huawei driver (or a similar change did).

When you say it is 65% faster--how many seconds are we talking about?

glennmatthews:
Typical `send_config_set` calls with a one-line config command improved from 11.3 seconds to 2.9 seconds with this change. Doesn't sound like much but when you're doing unit test automation with hundreds of such calls, it adds up fast.

I'm willing to rework this as a change confined to cisco_ios/cisco_xr/cisco_nxos if that would give you more confidence - I'll admit I hadn't realized there were so many non-Cisco devices that inherited from CiscoSSHConnection.

---

With fast_cli, I'd see inconsistent results - for example, here are outputs from the same `send_config_set` call on different invocations:

```
glenn-csr1000v#no ntp server 10.1.1.1
glenn-csr1000v(config)#end
glenn-csr1000v#
```

```
config term
glenn-csr1000v(config)#no ntp server 10.1.1.1
glenn-csr1000v(config)#end
glenn-csr1000v#
```

```
width 511
glenn-csr1000v#
glenn-csr1000v#
glenn-csr1000v(config)#no ntp server 10.1.1.1

glenn-csr1000v(config)#
```

Lines seem to be missing, unexpected lines included, etc. I can open a separate issue on that if you'd like.

ktbyers:
@glennmatthews Yes, if you could open a separate issue on fast_cli that would be helpful.

Yes, I think it would be easier if we apply this fix to only some subset of Cisco device types.

ktbyers:
Can you parallelize your tests at all (i.e. use concurrency to improve the testing time)? In general that is the fix for that type of problem (of course in a test environment that might be less practical as you might be testing against the same device multiple times)>

glennmatthews:
Reworked to only touch cisco_ios/cisco_nxos/cisco_xr device types. I've done some very quick sanity checks against devices of all three types and it appears to work.

And unfortunately, as you thought - this is executing a whole series of tests against a single device. We're looking at some various ways of optimizing the overall test flow, to be sure, but this was looking like an easy fix to gain a lot of ground. :-)

glennmatthews:
@ktbyers Any additional concerns/feedback on this PR?

ktbyers:
No probably not...still need to review it, but as long as it is device specific it shouldn't be an issue.

PR #1037, opened 2018-12-14
Fix Cisco IOS prompting for confirmation on save_config on OS version change

TomCos:
Fixes confirm failures in Napalm for copy run start, as well as fixes it in Netmiko to account for in the function.  I might go back and expand Napalm as well if this is approved.  NXOS uses a copy run start function, Napalm should too and look for an "[OK]"


ktbyers:
The linter is failing. You will need to run 

```
black /home/travis/build/ktbyers/netmiko/netmiko/cisco_base_connection.py
```

You will need to use Python3.6 or 3.7 for this (i.e. black requires PY3.6 or PY3.7 to execute).

TomCos:
Alright maybe I'm being a bit too ambitious (or maybe it just needs to be written a bit more pythonic), but I've re-written and tested a new way of implementing this.  Tests pass, so let me know what you think.

TomCos:
@ktbyers after some testing, it looks like "reload" won't work because we lose the socket instantly, shooting an exception from Paramiko.  How did it work in this example?
https://github.com/ktbyers/netmiko/blob/c8570c4a466a3dace74c22733ed20b1eb86e7dbb/examples/asa_upgrade.py

ktbyers:
I don't think we should have a `reload` method i.e. that this should be outside of Netmiko. 

Remember we would have to work this out for close to every platform which is on the order of 45 platforms.

In other words, Netmiko provides a way for the library user to send a reload pretty easily across a whole bunch of platforms using send_command/send_command_timing, but that implementation should be outside of Netmiko.

ktbyers:
Adding for reference, on Cisco IOS we are trying to fix this problem:

```
R2#write mem
Warning: Attempting to overwrite an NVRAM configuration previously written
by a different version of the system image.
Overwrite the previous NVRAM configuration?[confirm]
```

netmiko/cisco_base_connection.py
```diff
@@ -204,25 +205,51 @@ def _autodetect_fs(self, cmd="dir", pattern=r"Directory of (.*)/"):
...
+        # If confirm prompt detected or confirm=True, progress through action, otherwise return output
+        # Prompts are expected after confirms, so back to send_command
+        if confirm or any(confirm in output for confirm in ["[confirm]", "[y]", "[n]"]):
+            if confirm_response:
+                output += self.send_command(confirm_response, delay_factor=delay_factor)
+            else:
+                # Default RETURN if no confirm_response given
+                output += self.send_command(self.RETURN, delay_factor=delay_factor)
+        return output
+
```
ktbyers: I don't think we should have a separate method for this: I think handling of this should just all be inside of the save_config method using confirm and confirm_response. 

We can potentially convert confirm_response to be regex, but if we do that we need to make the behavior consistent across all of the netmiko platforms (so that is definitely more work).

ktbyers:
@TomCos Let me know if what you are trying to fix is broader than the specific issue I referenced above (excluding reload which I think we should discuss separately from this PR).

Regards, Kirk

TomCos:
I can remove reload.  I thought it fit the scheme well for the confirm command stand alone function.  I have not seen a platform where that reload code would not work, but if you don't think it should be in Netmiko, due to risk or scope, I can remove.  For the save command though, I don't see a good reason to combine that code again.  I think the confirm code does have usefulness outside of the save case.

It's really up to you though in the end.  My goal was to write something dynamic if possible.  If we think the abstraction adds too much complexity or risk, it can be removed.

ktbyers:
@TomCos We can talk about `reload`, but it should definitely be a separate issue or PR. I think you would have to research the proper `reload` procedure for each platform (one-by-one). In other word, what the command is, and what the network device sends back for this command. So it is a lot of work. Parenthetically, that is what we essentially did for `save_config` (though we didn't get them 100% complete we got pretty good coverage on them).

Yes, I didn't want the `confirm_code` precisely because I didn't want a separate use case. In other words, there are already two known patterns in Netmiko for handling additional prompting that are pretty straight forward (and I didn't want to create a third solution).

Regards, Kirk

TomCos:
All good notes, I made the changes per the comments.  Let me know what you think.

TomCos:
Don't do anything on this yet. I have a suspicion this isn't working as expected.  I didn't realize find_prompt() sends a return character (https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L1052).  The response send_command is likely sending a return before the command is even run (https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L1241), so it's an illusion it's actually working the way we expect.  It needs to be re-written using send_command_timing.

TomCos:
Applied changes above.

TomCos:
Just following up with some old code.  Any reason this never went in?

ktbyers:
At a minimum it needs to pass the linter/travis-CI, I am probably not going to look at it much if it doesn't do that.

There are probably other things I would want to change, but if it is not passing Travis-CI, then it isn't really a valid PR.

If you need help getting that done, let us know.


Regards, Kirk

TomCos:
Should be good now.

TomCos:
@ktbyers still want this to go in?

ktbyers:
Will have to check...it has been a while since I looked at this PR :-(

ktbyers:
Adding this reference in:

```
R2#write mem
Warning: Attempting to overwrite an NVRAM configuration previously written
by a different version of the system image.
Overwrite the previous NVRAM configuration?[confirm]
```

ktbyers:
Since this is very much the non-standard case, I think we should over-write this with arguments i.e.

```
net_connect.save_config(cmd="write mem", confirm=True, confirm_response="confirm")
```

Although that doesn't say what to do with NAPALM or napalm-ansible when this happens.

Also at the Netmiko level, it is hiding a potential issue (so I don't think Netmiko should mask).

NAPALM should potentially mask that, but at Netmiko-level you can work around this via arguments.


Sorry if I changed messaging on this across time.

TomCos:
@ktbyers hmmm ok so if we don't want to go with this approach, how are we going to fix it when it comes up?  I originally came up with the solution to solve https://github.com/ktbyers/netmiko/issues/610

ktbyers:
I don't think it necessarily needs fixed. 

It is a pretty non-standard situation...and the user can just fix it by executing a `write mem` and confirming.

And it only happens in OS upgrade which the user should be verifying the configuration changes from old-to-new anyways (post OS upgrade). This applies in both the manual and automated situations.

ktbyers:
FYI, the problem can be worked around in napalm also, by just executing:

`napalm_conn.device` to directly access the Netmiko driver and then call the `save_config` method as per above (with the arguments you need).

You can also work around this in ansible, but it will probably be more challenging.

TomCos:
Would it be unreasonable to go back to the original idea and have it as a feature add?

ktbyers:
Can you expand on the context a bit?

A feature add in NAPALM or napalm-ansible or something else?

And can you expand on what you mean by a feature add?

TomCos:
A feature added into Netmiko for automatically confirming a command, based on an unexpected prompt.  That was the way I originally had it written, but we moved it to be more specific.

ktbyers:
That is already a solved problem in Netmiko?

You can use either send_command with an expect_string or send_command_timing to do that?

TomCos:
Yes, but you'll need to expect the unexpected, which was what this pull was trying to do.  Personally, I've fixed this in my code to basically use this feature.  My direction was to solve it for people at a larger scale than just my code.  If we don't want to do that, I'll leave it at that.

ktbyers:
There are hundreds, if not thousands of these special cases we could bake in like this (where the behavior changes in odd ways). And this one isn't even that odd (since you would have had to upgrade the OS).

Netmiko would really become incredibly hard to maintain if we tried to integrate in handlers for all these special cases.

And we already have general solutions to solve them.

sljwampa:
So I updated my server and now I am running into netmiko timing out on the copy running-config startup-config. I just have a file that I send to the switch as I was doing before the update. I could change it to write memory but that is a depreciated command. In sending the file how do I get it to confirm with copy running-config startup-config?

ktbyers:
@sljwampa Does `write mem` still work? `write mem` has been deprecated on IOS for 15+ years, but until they actually remove the command it isn't actually deprecated (i.e. if they never remove the command, then it is just Cisco saying they don't like it...even though it is better since it doesn't prompt you for confirmation).

But if you really want to use `copy run start` sure go ahead...but then you will need to handle the extra prompting using an `expect_string` or `send_command_timing`

https://github.com/ktbyers/netmiko/blob/develop/EXAMPLES.md#handling-commands-that-prompt-expect_string

sljwampa:
Yes write mem still works and that is what I am doing for now. It of course not on my Nexus switches. I will try the expect string. This was working before I did some updates recently.

ktbyers:
Does NX-OS also do additional prompting upon an OS Upgrade?

Regards, Kirk

sljwampa:
I just look and the NX-OS does not do an additional prompt. So that will work for it. I already do a detection on what OS the device is.

ktbyers:
Yeah, then I would think the standard Netmiko `save_config()` method should work.

PR #1050, opened 2019-01-07
Modified base_connection to check last 3 buffer reads for saved prompt.

### LINKED ISSUE #1023: Cannot detect prompt if it is truncated
Hi,

I'm having prompt detection issues with an Exablaze device and I think I might have found where the issue lies.
Here's the error message:
> IOError: Search pattern never detected in send_command_expect

I added debug in the `send_command`([right here](https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L1262)) method and here's what I got:

```python
while i <= max_loops:
    new_data = self.read_channel()
    if new_data:
        if self.ansi_escape_codes:
            new_data = self.strip_ansi_escape_codes(new_data)

        # Case where we haven't processed the first_line yet (there is a potential issue
        # in the first line (in cases where the line is repainted).
        if not first_line_processed:
            print('\t=========== STARTING HERE ===============')
            output += new_data
            print('\tsub-output:\n{}'.format(output))
            output, first_line_processed = self._first_line_handler(
                output,
                search_pattern
            )
            print('\thandler-output:\n{}\nprocessed: {}\npattern: {}\n'.format(output, first_line_processed, search_pattern))
            # Check if we have already found our pattern
            if re.search(search_pattern, output):
                print('\tFound pattern {}'.format(search_pattern))
                break
            print('\t=========================================')
        else:
            output += new_data
            print('\t************** And then... ****************')
            print('\tOUTPUT: {}\nnew_data: {}\nsearch_pattern: {}'.format(output, new_data, search_pattern))
            # Check if pattern is in the incremental data
            if re.search(search_pattern, new_data):
                 break
```

This gave me an output similar to:

> 	=========== STARTING HERE ===============
> 	sub-output:
> config show running-config
> hostname xpar1
> [...]
> user@xp
> // The output ends here, and we can see that the prompt was about to be written, but a part of it is missing.
> processed: True
> pattern: user\@xpar1\>
> // Another read loop occurs...
> 	************** And then... ****************
> 	OUTPUT: config show running-config
> hostname xpar1
> [...]
> user@xpar1> 
> new_data: ar1> 
> search_pattern: user\@xpar1\>

The issue I see here according to the code is that since the prompt was truncated, it will neither be processed during the first network read (line ~1270) nor during the second (line ~1279, because the prompt is looked for in new_data only, not output), leading to a useless wait for new data that will never come. Is this analysis correct or am I missing something ?
If it isn't, how could I get my script to stop hanging ?
If it is, couldn't the method take "X" preceding characters, concatenate it with new_data and look for a prompt in this ?

Thanks.

### LINKED ISSUE #1049: Output split across two reads so send_command pattern never detected
Hi folks, 
I have a strange problem with the build-in `send_command()` method and I am not able to find a good solution to it.
I have a script which connects to several Cisco IOS-XR routers and issue some commands. In most cases it is working just fine, but sometimes there is this error message: `Search pattern never detected in send_command_expect`.

Here is my simplified code:
```
router_params = {
                    "host": router_hostname,
                    "username": username,
                    "password": password,
                    "device_type": 'cisco_xr',
                    "ssh_config_file": ssh_config_file
                }
                
router = ConnectHandler(**router_params)
router.send_command('show int description | e down | i R2-R2,', delay_factor=2)
router.send_command('show ipv4 int brief | e down | e Loop | e unassigned | i default', delay_factor=2)
```

When I tried to debug it I noticed strange behavior in the `self.read_channel()` method so I put some additional print() statements into the `send_command()` method as follow:
```
....
# Keep reading data until search_pattern is found or until max_loops is reached.
while i <= max_loops:
    print(f'Searching for prompt: {search_pattern}')          <---- MY DEBUG LINE
    new_data = self.read_channel()
    print(f'new_data: {new_data}')                            <---- MY DEBUG LINE
.....
```

And here is the output when the `Search pattern never detected in send_command_expect` is raised:

```
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: 
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: show int description | e down | i R2-R2,
Thu Jan  3 19:15:49.283 CET

Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: 
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: Te0/0/0/0          up          up          R2-R2,abcde
RP/0/RP0/CPU0:fra-xxxx-yyyy#
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: 
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: 
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: show ipv4 int brief | e down | e Loop | e unassigned | i default
Thu Jan  3 19:15:50.564 CET
TenGigE0/0/0/0                 10.1.1.1    Up              Up       default 
TenGigE0/0/0/18                10.1.2.1   Up              Up       default 
TenGigE0/1/0/18                10.1.3.1   Up              Up       default 
RP/0/RP0/C
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: PU0:fra-xxxx-yyyy#
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: 
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: 
.......
Search pattern never detected in send_command_expect: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
```

If you take a closer look especially into this line:
```
RP/0/RP0/C
Searching for prompt: RP/0/RP0/CPU0:fra\-xxxx\-yyyy\#
new_data: PU0:fra-xxxx-yyyy#
```
you will see the router prompt `RP/0/RP0/CPU0:fra-xxxx-yyyy#` is split to 2 outputs from the `self.read_channel()` and that is the reason why the prompt is never found and the error `Search pattern never detected in send_command_expect` is issued.

I suspect some delay problem with this but I also think the behavior of the `self.read_channel()` method could be improved as it reads in the middle of the output which could cause problems like this.

mreams13:
Fixes #1023. Fixes #1049.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -15,6 +15,7 @@ import re
 import socket
 import telnetlib
 import time
+from collections import deque
 from os import path
 from threading import Lock
 
@@ -1256,6 +1257,7 @@ class BaseConnection(object):
 
         i = 1
         output = ""
+        past_three_reads = deque(maxlen=3)
         first_line_processed = False
 
         # Keep reading data until search_pattern is found or until max_loops is reached.
@@ -1265,10 +1267,12 @@ class BaseConnection(object):
                 if self.ansi_escape_codes:
                     new_data = self.strip_ansi_escape_codes(new_data)
 
+                output += new_data
+                past_three_reads.append(new_data)
+
                 # Case where we haven't processed the first_line yet (there is a potential issue
                 # in the first line (in cases where the line is repainted).
                 if not first_line_processed:
-                    output += new_data
                     output, first_line_processed = self._first_line_handler(
                         output, search_pattern
                     )
@@ -1277,9 +1281,8 @@ class BaseConnection(object):
                         break
 
                 else:
-                    output += new_data
-                    # Check if pattern is in the incremental data
-                    if re.search(search_pattern, new_data):
+                    # Check if pattern is in the past three reads
+                    if re.search(search_pattern, "".join(past_three_reads)):
                         break
 
             time.sleep(delay_factor * loop_delay)


ktbyers:
Thanks for this fix...looks good.

PR #1060, opened 2019-01-11
Fix Python 3 serial driver issue.

### LINKED ISSUE #955: Serial driver python3 issues
```
edited line 173 in utilities.py to the following:
       cdc = next(serial.tools.list_ports.grep(name))
       return cdc[0]

and edited base_connection.py to add:
output += self.remote_conn.read(self.remote_conn.in_waiting).decode('utf-8', 'ignore')
on my line 385 (edited)
```

mreams13:
I made the changes that ktbyers specified in #955. I tested this with Python 3.6 and Windows 10. Fixes #955.

### DIFF
--- a/netmiko/utilities.py
+++ b/netmiko/utilities.py
@@ -183,7 +183,7 @@ def check_serial_port(name):
     """returns valid COM Port."""
     try:
         cdc = next(serial.tools.list_ports.grep(name))
-        return cdc.split()[0]
+        return cdc[0]
     except StopIteration:
         msg = "device {} not found. ".format(name)
         msg += "available devices are: "
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -457,7 +457,7 @@ class BaseConnection(object):
         elif self.protocol == "serial":
             output = ""
             while self.remote_conn.in_waiting > 0:
-                output += self.remote_conn.read(self.remote_conn.in_waiting)
+                output += self.remote_conn.read(self.remote_conn.in_waiting).decode('utf-8', 'ignore')
         log.debug("read_channel: {}".format(output))
         self._write_session_log(output)
         return output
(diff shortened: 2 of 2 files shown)


netmiko/utilities.py
```diff
@@ -183,7 +183,7 @@ def check_serial_port(name):
     """returns valid COM Port."""
     try:
         cdc = next(serial.tools.list_ports.grep(name))
-        return cdc.split()[0]
+        return cdc[0]
```
ktbyers: This change I don't understand as it implies either: 1)there was a bug before and it didn't work, 2)pyserial changed, 3)something else is going on.

What is the value of `cdc` here?

ktbyers: I think someone else reported this to me at one point, but I didn't have a good way to test it (I am not sure if it was you or someone else).

ktbyers: @mreams13

mreams13: You opened the bug and suggested the fix. I think someone else might have suggested the bug on Slack. I am currently using netmiko over serial and I ran into the bug. I noticed it was still open with a fix included, but no one had actually changed it: 

"""
print(cdc) = COM6 - Prolific USB-to-Serial Comm Port (COM6)
print(cdc[0]) = COM6
print(cdc[1]) = Prolific USB-to-Serial Comm Port (COM6)
print(cdc[2]) = USB VID:PID=067B:2303 SER=5 LOCATION=1-6

ktbyers [8:14 PM]
Can you `print(type(cdc))`
"""

ktbyers:
Can you fix the linting issues on the code? You can probably fix this by doing:

```
black .
pylama .
```

From the root of the Nornir project (i.e. the same directory as setup.py)

Note, pylama . might require that you do some additional modifications to the code.

Issue #982, opened 2018-10-18
Netmiko edge switch driver needs to be in enable() mode to disable paging.


ktbyers:

ktbyers:
```
By logging in, accessing, or using the Ubiquiti product, you
acknowledge that you have read and understood the Ubiquiti
License Agreement (available in the Web UI at, by default,
http://192.168.1.2) and agree to be bound by its terms.
(Edge_sw) >
(Edge_sw) >terminal length 0
            ^
% Invalid input detected at '^' marker.
(Edge_sw) >
(Edge_sw) >enable
Password:********************
(Edge_sw) #
```

pandvan:
This is also needed if you connect to a Cisco IOS device with a _privilege 0_ user.

ktbyers:
@pandvan Yes, a priv 0 user on Cisco IOS wouldn't be supported. You would need at least a priv 1 user (or would need to write some Python-Netmiko glue code to get it work with your case).

Issue #981, opened 2018-10-18
ubiquiti_edge unable to Execute comands


timofmax:
Hi Kirk!
Here is my code:
```
from netmiko import ConnectHandler
USER = "ubnt"
PASSWORD = "ubnt"
COMMAND = "show arp"

DEVICES_IP = ["192.168.5.1", "192.168.5.2"]

for IP in DEVICES_IP:
    print('Connection to device {}'.format(IP))
    DEVICE_PARAMS = {'device_type': 'ubiquiti_edge',
                     'ip': IP,
                     'username': USER,
                     'password': PASSWORD,
                     'secret': PASSWORD}

    with ConnectHandler(**DEVICE_PARAMS, session_log='output.txt') as ssh:
        ssh.enable()
        result = ssh.send_command(COMMAND, expect_string=r'#')
        print(result)


device_ip = "172.23.5.1"
DEVICE_PARAMS = {'device_type': 'ubiquiti_edge',
                 'ip': device_ip,
                 'username': USER,
                 'password': PASSWORD,
                 'secret': PASSWORD}
```
and here is output.txt (mac address changed for privacy :) )
```Welcome to EdgeSwitch
By logging in, accessing, or using the Ubiquiti product, you
acknowledge that you have read and understood the Ubiquiti
License Agreement (available in the Web UI at, by default,
http://192.168.1.2) and agree to be bound by its terms.
(Edge_sw) >
(Edge_sw) >terminal length 0
            ^
% Invalid input detected at '^' marker.
(Edge_sw) >
(Edge_sw) >enable
Password:********************
(Edge_sw) #
(Edge_sw) #show arp
Age Time (seconds)............................. 300
Response Time (seconds)........................ 1
Retries........................................ 4
Cache Size..................................... 493
Dynamic Renew Mode ............................ Disable
Total Entry Count Current / Peak .............. 167 / 208
Static Entry Count Configured / Active / Max .. 0 / 0 / 128
  IP Address        MAC Address      Interface        Type        Age
---------------  -----------------  --------------  --------  -----------
10.0.4.1       11:23:41:12:aa:bb  4/4             Gateway    0h  0m  2s
10.0.4.2       11:23:41:12:aa:bb  4/4             Local         n/a
10.0.4.3       11:23:41:12:aa:bb  4/4             Dynamic    0h  0m  3s
10.0.4.5       11:23:41:12:aa:bb  4/4             Dynamic    0h  0m  6s
[...]

PR #1065, opened 2019-01-14
Fix ubiquity edge enable issue

ktbyers:


ktbyers:
Fixes #982 and fixes #981.

Issue #1058, opened 2019-01-10
Small device input buffers and send_config_set failures


jtdub:
We've recently encountered issues where we send a long list of commands to `send_config_set` and a remote device appears to fill up the input buffers on the device, causing it to not properly execute commands.

This appears to be an artifact of how `send_config_set` operates, by sending the entire list of commands to the device, then grabbing the entire output afterwards.

I believe in instances such as this, we could batch the commands, grab output, then proceed to the next batch of commands until the entire list of commands has been executed.

My thought is something like this:
```
def send_config_set(
    self,
    config_commands=None,
    exit_config_mode=True,
    delay_factor=1,
    max_loops=150,
    strip_prompt=False,
    strip_command=False,
    config_mode_command=None,
    batch_commands=False,
    batch_length=10,
):
    import math
    """
    Send configuration commands down the SSH channel.
    config_commands is an iterable containing all of the configuration commands.
    The commands will be executed one after the other.
    Automatically exits/enters configuration mode.
    :param config_commands: Multiple configuration commands to be sent to the device
    :type config_commands: list or string
    :param exit_config_mode: Determines whether or not to exit config mode after complete
    :type exit_config_mode: bool
    :param delay_factor: Factor to adjust delays
    :type delay_factor: int
    :param max_loops: Controls wait time in conjunction with delay_factor (default: 150)
    :type max_loops: int
    :param strip_prompt: Determines whether or not to strip the prompt
    :type strip_prompt: bool
    :param strip_command: Determines whether or not to strip the command
    :type strip_command: bool
    :param config_mode_command: The command to enter into config mode
    :type config_mode_command: str
    :param batch_commands: Break the list of commands into batches for execution
    :type batch_commands: bool
    :param batch_length: Maximum length of each command batch
    :type batch_length: int
    """
    delay_factor = self.select_delay_factor(delay_factor)
    if config_commands is None:
        return ""
    elif isinstance(config_commands, string_types):
        config_commands = (config_commands,)

    if not hasattr(config_commands, "__iter__"):
        raise ValueError("Invalid argument passed into send_config_set")

    batch_length = batch_length if batch_commands else len(config_commands)
    batches = math.ceil(len(config_commands) / batch_length)
    start_batch = 0
    end_batch = batch_length
    fast_cli = False if batches > 1

    # Send config commands
    cfg_mode_args = (config_mode_command,) if config_mode_command else tuple()
    output = self.config_mode(*cfg_mode_args)
[...]

PR #1085, opened 2019-01-30
Add a config command rate-limiter

aedwardstx:
Rate-limits commands sent on the channel based on prompts seen.

Arbiter is a send command rate limiter that maintains the command buffer at `bucket_size` number of commands. It would allow sending of bucket_size commands, when it sees X prompts come back, it allows another X command(s). If it doesn’t see a prompt come back in some time, it figures “I must have missed it” and sends another command. This continues until the config set is depleted.

ref #1058

Recent relevant article: https://blog.ipspace.net/2019/01/not-so-fast-ansible-cisco-ios-cant-keep.html

### DIFF
--- a/netmiko/arbiter.py
+++ b/netmiko/arbiter.py
@@ -1,5 +1,3 @@
-
-
 class CommandBufferArbiter:
     """
     Acts as a send command rate limiter that maintains the “command buffer” at bucket_size number of commands.
@@ -30,7 +28,7 @@ class CommandBufferArbiter:
         # lines that have yet to be checked for prompts
         self.pending_lines = []
         # The current line that has not yet been completed with a terminating EOL
-        self.incomplete_line = ''
+        self.incomplete_line = ""
 
     def get_token(self, output):
         """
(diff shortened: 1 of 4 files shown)


ktbyers:
FYI, the issue is probably not relevant i.e. that issue is probably not an input buffer overload problem (I don't think there are enough commands to overload the input buffer and the description of the failure is not consistent with an input buffer overload).

Also FWIW I couldn't reproduce the issue described (though I only tested with a copy-and-paste and not with Ansible's ios_config).

aedwardstx:
I agree the article does not exactly fit the situation but I chose to reference it as it provides a bit of background. The referenced issue is on point for what's being addressed in the PR.

ktbyers:
How would the PR fix the referenced issue? I don't think it would even at a command buffer size of 1?

aedwardstx:
Issue #1058 describes a situation where `send_config_set()` sent hundreds of lines to a Nexus3k. A large part of that config was an ACL. The device took a long time to process each ACL line and ended up tail dropping ACL config lines in the limited command buffer. The article touches on this a little bit.
Arbiter watches for prompts in the output to know when the device has processed a command indicating that its safe to send another. It allows for `bucket_size` commands to be unacknowledged at a time. 

There's two basic approaches to solve for this:
- Send one line after seeing one prompt with no outstanding lines
  - This approach leaves the device idle for periods of time in between commands
- Send X lines and allow another to be sent for each prompt seen <- Arbiter
  - This is as fast as `send_config_set()` method of sending the commands in bulk without the risk of overrunning the command buffer.

ktbyers:
@aedwardstx I know this is not exactly what you were looking for, but we fixed this problem by essentially going to a command-buffer of one i.e. Netmiko won't write the next command in send_config_set until the last one has been properly echoed/handled by the device.

That fix is in the Netmiko develop branch and there is a modification to the behavior here:

https://github.com/ktbyers/netmiko/pull/1447

I need to make some slight changes to that PR1447 as I have hard-coded a terminating char '#' in there, but I will update that shortly.

PR #1099, opened 2019-02-18
Add environment markers to support Poetry

ktbyers:


ktbyers:
Tested universal wheel on both Python3.6 and Python2.7. Verified that on Python3.6 there was no enum34 installed and that on Python2.7 both enum34 and ipaddress were installed.

PR #1105, opened 2019-02-26
added the ability to login to a ios device that has enable mode  from…

tybruno:
Currently, there will be a timeout when trying too login to a device (ios) with an enable password from a console server using netmiko. I am aware that using the 'terminal_server' device_type and `redispatch()` can fix this problem. However, this is an issue when using Napalm. Clearing the line before attempting to send the enable password will prevent the session from timing out.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -243,6 +243,10 @@ class BaseConnection(object):
 
             self._modify_connection_params()
             self.establish_connection()
+
+            #clears the line when trying to enter enable mode from console server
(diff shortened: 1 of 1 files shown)


ktbyers:
What do you mean by `clearing the line` here?

carlmontanari:
As this is a bit old and I think it can be addressed by simply sending a return char with `write_channel` or in probably a fair few other ways, I'll close this. If you want to reopen please provide some details in response to Kirk's questions!

Issue #1052, opened 2019-01-08
WLC incorrect login returns false error


jrecords:
If wrong username/password is given using 'cisco_wlc' device type:

raise ValueError("Router prompt not found: {0}".format(repr(prompt)))
ValueError: Router prompt not found: 'User:'

Instead of Invalid login credentials.

ktbyers:
I am not sure this is worth fixing. 

The WLC has the very strange behavior where it adds the extra `login as` message (which forces Netmiko to add special code to handle the login).

```
        login as: user

        (Cisco Controller)

        User: user

        Password:****
```

We loop a certain amount of times and if we ever send the password, the we assume it is correct and then move on to the next step. Above you are failing on the next step.

We could add a check after sending the password whether we successfully read a `>` as one possible solution. Is the `>` guaranteed to always be present post-login (or is that configurable)?

OzNetNerd:
@ktbyers, I've raised https://github.com/ktbyers/netmiko/pull/1107 for this issue.

ktbyers:
Error message improved in #1107

PR #1107, opened 2019-02-27
Appropriate exception message for incorrect WLC credentials

OzNetNerd:
Fix for https://github.com/ktbyers/netmiko/issues/1052.

### DIFF
--- a/netmiko/cisco/cisco_wlc_ssh.py
+++ b/netmiko/cisco/cisco_wlc_ssh.py
@@ -100,7 +100,12 @@ class CiscoWlcSSH(BaseConnection):
         Cisco WLC uses "config paging disable" to disable paging
         """
         self._test_channel_read()
-        self.set_base_prompt()
+
+        try:
+            self.set_base_prompt()
+        except ValueError:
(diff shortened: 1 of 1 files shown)


whitej6:
One thing to keep in mind is that this can give an incorrect error message and should be using a generic authentication fail.  Auth fail is not always just invalid username/password.  My initial PR was because WLCs do not support a password character length of longer than 25.  In that instance the un/pw is sent correctly but the WLC does not forward the auth request to tacacs it however drops the auth request which would be confusing to the enduser seeing incorrect username/password.  Similarly if the authentication was rejected by tacacs, incorrect username/password would not be the correct response.

OzNetNerd:
Thanks for the information @whitej6. In that case your `unable to authenticate` message is more appropriate.

ktbyers:
@whitej6 Yes, PRs are welcome so if you want to create a new PR with a better message, that would be great.

I don't think we should get into trying to parse what the WLC does, however (nor should we try to capture/raise the messages from the WLC) so I think `unable to authenticate` is probably sufficient.

PR #1139, opened 2019-03-15
fix for nxos scp

carlniger:
first draft attempt to fix issue referenced [here](https://github.com/napalm-automation/napalm/issues/931)

tox tests passed; tested scp against both versions of nxos mentioned in above issue, tested various netmiko tasks against IOS/nxos as well. 

this may not be the *best* fix, but wanted to get something opened up so we can address it/talk about it.

-Carl


ktbyers:
Replaced by https://github.com/ktbyers/netmiko/pull/1140

PR #1140, opened 2019-03-18
Fixing MD5 / carriage return issue on NXOS 9K

ktbyers:


ktbyers:
I tested this against the 7K, but it needs to be regression tested against the Nexus 9K.

Fixes https://github.com/ktbyers/netmiko/pull/1140

b-marcum:
This fix has been successfully tested for SCP functionality on the following platforms:

93180YC-EX - 7.0(3)I7(6) 
9000v - 7.0(3)I7(2)
5596UP - 7.3(5)N1(1)

Issue #1125, opened 2019-03-07
Ruckus fastiron does not reliably enter enable mode.


spdkils:
I support devices across the world. Slow and slightly unreliable connections can miss setup on Brocade with it's current implementation.

The current method is a blind call to enable, and if it doesn't work, paging doesn't work. If paging doesn't work, almost nothing works, and you're not in a 'known' position. No error is made, and no second attempt is made to get a good initialization.

I suggest a slight change to the setup to deal with this. I've tested it and it works, the exact error raised etc, I'm ambivalent to.
```        self.set_base_prompt()
        self.enable() #<--- no net
        self.disable_paging(command="skip-page-display")
```

to something like....
```        self.set_base_prompt()
        for _ in range(3):
            self.enable()
            if self.check_enable_mode():
                break
        else:
             msg = "Unable to enter enable mode. blah blah blah fix it"
             raise ValueError(msg)

        self.disable_paging(command="skip-page-display")
```

ktbyers:
Why doesn't the call to `enable` work? Have you tried adjusting `global_delay_factor` so that it works sufficiently reliably?

If enable() fails, doesn't an exception already get raised?

spdkils:
```    def enable(
        self, cmd="enable", pattern=r"(ssword|User Name)", re_flags=re.IGNORECASE
    ):
...
...
                if "ssword" in new_data:
                    self.write_channel(self.normalize_cmd(self.secret))
                    output += self.read_until_prompt()
                    return output <----------- No raise.... just return.
....
...
....
# Below code never gets hit because the early return...
        if not self.check_enable_mode():
            msg = (
                "Failed to enter enable mode. Please ensure you pass "
                "the 'secret' argument to ConnectHandler."
            )
            raise ValueError(msg)
```

spdkils:
Man that above comment it horrible. :( Sorry.

There is a return in the rukus enable, that early outs before the raise of the valueerror.
I suspect that is the problem I hit, I had to open up the code again to look.

ktbyers:
```
 49                     output += self.read_until_prompt()
 50                     return output
```

Okay, we can change line 50 to be a break and then have the return output come after the check_enable_mode:

```
        if not self.check_enable_mode():
            msg = ( 
                "Failed to enter enable mode. Please ensure you pass "
                "the 'secret' argument to ConnectHandler."
            )   
            raise ValueError(msg)
return output         # add this
```

Do you want to submit a pull request on that?


Kirk

[...]

PR #1141, opened 2019-03-18
Enable Logic Fixes

spdkils:
In relation to issue #1125

After looking at the code, there was a dead while loop, and the logic didn't follow as expected.

It setup a while loop, and as soon as it entered the 'secret' it would return the output, without checking if it actually got the enable or not. This is the problem reported in issue #1125 

With this new logic, if you already have enable, it just returns None. (This is the same behavior as before, it had no explicit return for this situation, so it does the same thing.) However it doesn't have to check enable twice, a very minor optimize.

If you do not have enable, rather than going into a loop that always ends in 1 iteration, or a timeout. It now can end with a timeout as per-normal. However, after it sends the 'secret' it checks enable mode, to know if it got it or not.

I was going to try and have it loop 3 times as before, but it didn't seem logical, to just fail 3 password fails in a row, every time you connect. I didn't want to force lockout counters up, or massive failed attempts. So it attempts it, if it works, you return output as per normal, if you fail, it raises ValueError as it should have done previously, but was not.

I also removed the sleep, because it's not looping.. However I tested both ways, and this is now being consistent. I get a good connection on a bad link... Or I get an error. No longer am I in the 3rd state, of connected, but without paging, in an unknown enable state.

I tested it over several dozen connections, and it seems to fix all the problems I was having...

--Allen

### DIFF
--- a/netmiko/ruckus/ruckus_fastiron.py
+++ b/netmiko/ruckus/ruckus_fastiron.py
@@ -19,8 +19,7 @@ class RuckusFastironBase(CiscoSSHConnection):
         self.clear_buffer()
 
     def enable(
-        self, cmd="enable", pattern=r"(ssword|User Name)", re_flags=re.IGNORECASE
-    ):
+            self, cmd="enable", pattern=r"(ssword|User Name)", re_flags=re.IGNORECASE):
         """Enter enable mode.
         With RADIUS can prompt for User Name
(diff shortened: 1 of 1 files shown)


ktbyers:
Hopefully this issue has already been fixed in the current netmiko develop branch. If not, we should revisit the issue in that branch.

Issue #1144, opened 2019-03-19
ExtremeExosSSH.set_base_prompt()


dracoboros:
Running "show system" in Extreme OS makes the device runs other commands like:

show switch
show version
show temperature

However, when "show temperature" is executed, Netmiko assumes the prompt value as "show temperature" as It can be seen in the following Traceback:

```
Traceback (most recent call last):
  File "<omitted>", line 21, in <module>
    out += conn.send_command(cmd)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/netmiko/extreme/extreme_exos.py", line 53, in send_command
    self.set_base_prompt()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/netmiko/extreme/extreme_exos.py", line 37, in set_base_prompt
    cur_base_prompt = super(ExtremeExosBase, self).set_base_prompt(*args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/netmiko/base_connection.py", line 959, in set_base_prompt
    raise ValueError("Router prompt not found: {0}".format(repr(prompt)))
ValueError: Router prompt not found: 'show temperature'
```

I tried to mess up with find_prompt but not success to make it work.

If you need any further information, please feel free to ask for it.

dracoboros:
a-me-sw-ula-142-x460-01.1 # show temperature 
Field Replaceable Units               Temp (C)   Status   Min  Normal   Max
---------------------------------------------------------------------------
Switch         : X460-24t               31.50    Normal   -10    0-45  60
XGM3-2sf-A-1   :                        44.00    Normal   - 5    0-60  65
XGM3-2sf-A-1   : 
a-me-sw-ula-142-x460-01.2 #

dracoboros:
the session_log output with sensitive information omitted:

```
hostname-exos.1 # 

hostname-exos.1 # disable clipaging

hostname-exos.2 # disable cli prompting

hostname-exos.3 # 

hostname-exos.3 # show system


show switch
SysName:          hostname-exos
SysLocation:      
SysContact:       
System MAC:       
System Type:      

SysHealth check:  Enabled (Normal)
Recovery Mode:    All
System Watchdog:  Enabled

Current Time:     
Timezone:        
Boot Time:        
Boot Count:       
Next Reboot:      None scheduled
System UpTime:     

Current State:    OPERATIONAL             
[...]

PR #1145, opened 2019-03-20
Fixing send_command in ExtremeExosBase for long show commands

dracoboros:
Regarding issue #1144 

The line <code>self.set_base_prompt()</code> inside <code>send_command</code> in netmiko.extreme.extreme_exos.ExtremeExosBase was causing the prompt to be bad defined.


dracoboros:
Cutting <code>self.set_base_prompt()</code> only works for long show commands

PR #1146, opened 2019-03-20
extreme_exos.py send_command fixed for show commands

dracoboros:
ExtremeExosBase send_command function fixed regarding Issue #1144

### DIFF
--- a/netmiko/__init__.py
+++ b/netmiko/__init__.py
@@ -23,7 +23,7 @@ NetmikoTimeoutError = NetMikoTimeoutException
 NetmikoAuthError = NetMikoAuthenticationException
 Netmiko = ConnectHandler
 
-__version__ = "2.3.3"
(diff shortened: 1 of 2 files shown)


netmiko/__init__.py
```diff
@@ -23,7 +23,7 @@
 NetmikoAuthError = NetMikoAuthenticationException
 Netmiko = ConnectHandler
 
-__version__ = "2.3.3"
+__version__ = "2.3.4"
```
jdillard: Is there a reason you bumped the version? That looks like a job reserved for the maintainers.

dracoboros: I did not know about that. I have few expierence with Github.

Thank you for telling me that!

ktbyers:
This PR looks like it never passed CI-CD and also needs rebased.

If someone is interested in working on it, just let me know.

PR #1147, opened 2019-03-24
Use class instead of instance for key policies

david-caro:
It turns out that using the instance does not work as expected
(sometimes).

According to the current docs, both instance or class are accepted, but
it seems to behave differently when using one or the other:

http://docs.paramiko.org/en/2.4/api/client.html?highlight=Policy#paramiko.client.SSHClient.set_missing_host_key_policy

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -291,9 +291,9 @@ class BaseConnection(object):
             self.protocol = "ssh"
 
             if not ssh_strict:
-                self.key_policy = paramiko.AutoAddPolicy()
+                self.key_policy = paramiko.AutoAddPolicy
             else:
-                self.key_policy = paramiko.RejectPolicy()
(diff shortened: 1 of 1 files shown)


ktbyers:
I have never seen an issue with this. What is the issue that it causes?

carlniger:
@david-caro is this still an issue / what was the original issue? I'm going to close this for now but let us know if this is still an issue and we can re-open.

david-caro:
> @david-caro is this still an issue / what was the original issue? I'm going to close this for now but let us know if this is still an issue and we can re-open.

I was having issues when trying to parallelize code using this, getting problems with the host authentication complaining about keys, and it seemed to go away if I did this change, passing the instance instead of the bare class.

I know it sounds weird, and I did not have time to follow it up, sorry.

I'll reopen if I hit that issue again, thanks for your reviews though.

PR #1158, opened 2019-03-30
Update save_config method params.

ctopher78:
@carlniger @ktbyers 

Related to: https://github.com/nornir-automation/nornir/issues/352

Nornir's built-in Netmiko task expects Netmiko's `save_config` method to include the `confirm_response` parameter.  The decision in the aforementioned Nornir issue was to make all Netmiko network drivers have the same `save_config` method parameters.  

This PR makes that update ~~and fixes one small incorrect super parameter found while updating the method signatures.~~   I rolled back my change to the dell_dnos6 `super` method since things work fine with the way it was and I don't want to break something I don't understand.

### DIFF
--- a/netmiko/alcatel/alcatel_aos_ssh.py
+++ b/netmiko/alcatel/alcatel_aos_ssh.py
@@ -40,6 +40,10 @@ class AlcatelAosSSH(CiscoSSHConnection):
         """No config mode on AOS"""
         return ""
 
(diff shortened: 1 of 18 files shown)


ktbyers:
Updated in the following PR:

https://github.com/ktbyers/netmiko/pull/1187

ktbyers:
@ctopher78 The super issue was a bug--so I fixed that also (in the new PR).

PR #1165, opened 2019-04-09
Added 'huawei_telnet' device_type, and HuaweiTelnet class

cuong-nguyenduy:
Added device_type and class to support telnet on Huawei routers


ktbyers:
See updated PR

PR #1167, opened 2019-04-10
 - Aruba (recent hp procurve switches)

zhecho:
Tested with:
Switch
  - aruba JL321A 
  - os version: WC.16.04.0008

 	modified:   netmiko/aruba/aruba_ssh.py


ktbyers:
@zhecho Is the Aruba OS the same as HP Provision?

If so, we should convert this over to use a common base. In other words, the Aruba OS driver should just inherit from the `HPProcurveBase` class.

Adding netmanchris also.

zhecho:
> 
> 
> @zhecho Is the Aruba OS the same as HP Provision?
> 
> If so, we should convert this over to use a common base. In other words, the Aruba OS driver should just inherit from the `HPProcurveBase` class.
> 
> Adding netmanchris also.
@ktbyers I'm not aware of HP Provision devices. Tthere are some differences between HP Procurve and Aruba (output of some cli commands and other stuff... ), but generally you're right we can inherit it from HPProcurveBase

PR #1174, opened 2019-04-21
extend cisco_nxos device type to accomodate for cisco ucs

jonesbra:
Cisco UCS will output similar to the following to show version command: 
```

Firmware Version     
-------------------- 
2.0(6d)            

```

@ktbyers  - Show version output.

### DIFF
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -91,7 +91,7 @@ SSH_MAPPER_BASE = {
     },
     "cisco_nxos": {
         "cmd": "show version",
-        "search_patterns": [r"Cisco Nexus Operating System", r"NX-OS", r"Firmware Version"],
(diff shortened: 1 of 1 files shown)


ktbyers:
@jonesbra "Firmware Version" is too generic. If there is something more specific that we can match on that, then that might work.

ktbyers:
If you output the "show version" we can see if there is a string in the output that might work.

ktbyers:
Can you post the "show version" here?

That isn't going to be sufficient i.e. it is too generic of a pattern to search on.


Regards, Kirk

jonesbra:
The original comment has been updated with the show version output.

ktbyers:
You can try something other command that would uniquely identify it, but that is not going to work.

It is too general and would likely cause false matches.

Issue #1097, opened 2019-02-14
Inconsistent calculation of max_loops in base_connection.py?


sulisu:
In the current develop branch, base_connection.py
The first max_loops has no int, while the second one has. Is it inconsistent?

```
    def _read_channel_expect(self, pattern="", re_flags=0, max_loops=150):
        if max_loops == 150:
            max_loops = self.timeout / loop_delay

    def _read_channel_timing(self, delay_factor=1, max_loops=150):
        if delay_factor == 1 and max_loops == 150:
            max_loops = int(self.timeout / loop_delay)
```

carlniger:
@sulisu good eye. I'm assuming this didn't impact anything for you? if it did, let us know. otherwise ill get this into a pr this week to ensure consistency.

Issue #1089, opened 2019-02-04
TextFSM relative path not supported for NET_TEXTFSM


robobeaver6:
I have an environment variable set to the index path  
```python
os.environ["NET_TEXTFSM"] = "lib/ntc-templates/templates"
```
netmiko/utilities.py sets the index_file variable as the path from the environment variable joined to 'index'
```python
def get_structured_data(raw_output, platform, command):
    """Convert raw CLI output to structured data using TextFSM template."""
    template_dir = get_template_dir()
    index_file = os.path.join(template_dir, 'index')
    textfsm_obj = clitable.CliTable(index_file, template_dir)
```

netmiko/_textfsm/_clitable.py then joins the template_dir and the index_file (which already has the full path), resulting in an incorrect path.  
```
fullpath = os.path.join(self.template_dir, self.index_file)
```
The final path is `'lib/ntc-templates/templates/lib/ntc-templates/templates/index'`
which is incorrect, due to the directory being joined twice. 

This is netmiko  v2.3.0

ktbyers:
Okay, strange. I would have thought I tested this...but let me review it.

robobeaver6:
if it is of any use, I was calling Netmiko from Nornir using the following command:
```python
table_advertised_routes = task.run(task=netmiko_send_command,
                                                   name=f'Get Recieved Routes from {task.host} - {peer_ip}',
                                                   command_string=command,
                                                   use_textfsm=True,
                                                   severity_level=logging.DEBUG)
```

ktbyers:
Okay, probably in the interim you will need to install ntc-templates into your home directory as Netmiko will by default search there. I will then need to fix the environment variable issue.

Thanks for reporting this.

robobeaver6:
I hust changed netmiko/utilities.py line 234 so index_file didn't join to the template_dir:
```python
def get_structured_data(raw_output, platform, command):
    """Convert raw CLI output to structured data using TextFSM template."""
    template_dir = get_template_dir()
    # index_file = os.path.join(template_dir, 'index')
    index_file = 'index'
```
This seemed to work fine in the interim.  Not sure if it breaks anything else, but it is working for me.

ktbyers:
@robobeaver6 I dug into this some more, the issue is you need to use an absolute path and not a relative path here:

```
os.environ["NET_TEXTFSM"] = "lib/ntc-templates/templates"
```


os.path.join works fine if you use an absolute.

```
>>> import os
[...]

PR #1177, opened 2019-04-22
support textfsm relative path

carlniger:
support relative path for textfsm templates; ensure max_loops is whole number (for consistency)

https://github.com/ktbyers/netmiko/issues/1089
https://github.com/ktbyers/netmiko/issues/1097


ktbyers:
Replaced by #1178

Issue #756, opened 2018-03-22
use_textfsm is not working when exporting path using ~ symbol


d3nisko:
Hi, I'm using ntc templates to transform output of the "show interfaces status" command. 

First case: ntc templates cloned into home dir.
**NET_TEXTFSM=/home/testuser1/ntc-templates/templates**
Behavior: Everything is working as expected. Dict returned. 

Second case: ntc templates cloned into the working dir with the script. full path specified.
**NET_TEXTFSM=/home/testuser1/scripts/learn/automation/python_for_ne/excercises/lesson_6/ntc-templates/templates**
Behavior: Working as expected. Dict returned. 

Third case: ntc templates cloned into the home dir. ~ is used to indicate the path to home dir
**NET_TEXTFSM=~/ntc-templates/templates**
Behavior: exception appeard -

> > Traceback (most recent call last):
>   File "6_5.py", line 17, in <module>
>     output = net_conn.send_command('show interfaces status', use_textfsm=True)
>   File "/usr/local/lib/python2.7/dist-packages/netmiko/base_connection.py", line 973, in send_command
>     command=command_string.strip())
>   File "/usr/local/lib/python2.7/dist-packages/netmiko/utilities.py", line 219, in get_structured_data
>     template_dir = get_template_dir()
>   File "/usr/local/lib/python2.7/dist-packages/netmiko/utilities.py", line 202, in get_template_dir
>     raise ValueError(msg)
> ValueError: 
> Valid ntc-templates not found, please install https://github.com/networktocode/ntc-templates
> and then set the NET_TEXTFSM environment variable to point to the ./ntc-templates/templates
> directory. 
> 

Fourth case: ntc templates cloned into working dir. ~ is used to indicate the path to home dir
**NET_TEXTFSM=~/scripts/learn/automation/python_for_ne/excercises/lesson_6/ntc-templates/templates** 

> Traceback (most recent call last):
>   File "6_5.py", line 17, in <module>
>     output = net_conn.send_command('show interfaces status', use_textfsm=True)
>   File "/usr/local/lib/python2.7/dist-packages/netmiko/base_connection.py", line 973, in send_command
>     command=command_string.strip())
>   File "/usr/local/lib/python2.7/dist-packages/netmiko/utilities.py", line 219, in get_structured_data
>     template_dir = get_template_dir()
>   File "/usr/local/lib/python2.7/dist-packages/netmiko/utilities.py", line 202, in get_template_dir
>     raise ValueError(msg)
[...]

PR #1178, opened 2019-04-22
Fix NET_TEXTFSM path issues

ktbyers:
Support relative paths
Support ~ in paths


ktbyers:
Fixes #1089 #1097

ktbyers:
Fixes #756

PR #1176, opened 2019-04-22
Updated Huawei telnet support PR

ktbyers:

PR #1179, opened 2019-04-23
Fix calling super() for HuaweiBase

cuong-nguyenduy:
Change from `HuaweiSSH `to `CiscoBaseConnection `when calling` super()` inside class `HuaweiBase`


ktbyers:
Fixed in original PR as this was easier:

https://github.com/ktbyers/netmiko/pull/1176/commits/ea6aa4006b4cac2bb0f2e4a99d20275c8c1bb6a5

PR #1180, opened 2019-04-23
initial precommit hook setup

carlniger:
simple pre commit setup, we can obviously edit line-length and/or add pycodestyle/flake8/etc. but wanted to get something to start with!


ktbyers:
Decided not to implement right now...still required manual steps to install the git hooks. Additionally, probably going to be some issues with Black and the PY36 requirement.

PR #1182, opened 2019-04-24
Raise exception on unsafe charecters

AndersBallegaard:
I hope this added validation can save someone some frustrating troubleshooting
The current way netmiko handles unsafe charecters like a newline at the end is giving you a timeout

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -217,7 +217,8 @@ class BaseConnection(object):
             "^",
             "~",
             "[",
-            "]" "'",
+            "]",
+            "'",
         ]
 
         if ip:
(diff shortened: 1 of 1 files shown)


ktbyers:
I am not seeing why this is needed? What is the use case?

I guess we could call a strip() on the host and IP fields, but the unsafe characters strikes me more as being outside of Netmiko (i.e. if you are wrapping an application that then feeds to Netmiko, you should validate the host/ip other fields in the application)?

AndersBallegaard:
I considered using strip(), but I would rather validate the IP and host field rather than modify them. 

My main reason for adding this verification is that currently if you give it an address that contains unsafe characters, you get a NetMikoTimeoutException witch is definitely not the correct exception to raise for an invalid address.

I agree the application should validate input but that won't always happen so it's important that Netmiko atleast throws a releavant error

ktbyers:
I would rather do something that feeds it into a library like the IPAddress library and a similar library for hostname checking (instead of writing custom checking).

Does that sound reasonable?

AndersBallegaard:
Using a library would be reasonable. Unfortunately the IPAddress library wouldn't work on it's own since it doesn't work on domain names. 

I tried the validators library, but it seems to be bad at spotting unsafe urls and would add a new dependency. But if you know of any that would work it would probably be a better solution

ktbyers:
Yes, I was referring to two separate things like IPAddress and gethostbyname then coupled with IPAddress--but basically reusing existing solutions for both cases.

But anyways I don't want a manual character check as per-what has been included here.

Regards, Kirk

PR #1202, opened 2019-05-07
strip ansi foreground/background colors

carlniger:


carlniger:
I ran black on the tests so there are some changes in there from that (just looks like trailing whitespace trimming)

PR #1214, opened 2019-05-15
CloudGenix ION Driver

ktbyers:


netmiko/cloudgenix/cloudgenix_ion.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def session_preparation(self, *args, **kwargs):
+        """Prepare the session after the connection has been established."""
+        self.ansi_escape_codes = True
+        self._test_channel_read()
+        self.write_channel(self.RETURN)
+        self.set_base_prompt(delay_factor=5)
+
+    def disable_paging(self):
+        """Cloud Genix ION sets terminal height in establish_connection"""
```
ktbyers: CloudGenix ION doesn't have a generic way to disable output paging?

netmiko/cloudgenix/cloudgenix_ion.py
```diff
@@ -0,0 +1,83 @@
...
+    def find_prompt(self, delay_factor=1):
+        prompt = super(CloudGenixIonSSH, self).find_prompt()
+        prompt = self.strip_backspaces(prompt).strip()
+        return prompt
+
+    def strip_command(self, command_string, output):
+        output = super(CloudGenixIonSSH, self).strip_command(command_string, output)
+        # handle command_string being repainted - otherwise return normal output
+        if len(output.split(command_string)) == 4:
+            output = output.split(command_string)[3]
```
ktbyers: @carlniger Why 4?

netmiko/cloudgenix/cloudgenix_ion.py
```diff
@@ -0,0 +1,83 @@
...
+        return output
+
+    def config_mode(self):
+        """No configuration mode on ION SSH"""
+        self._in_config_mode = True
```
ktbyers: @carlniger This is fine here, but I don't think this is the common pattern I have implemented for platforms that don't support config mode in Netmiko. In other words, I wouldn't generally want to propagate this Mikrotik patter for tracking config state forward.

ktbyers:
@carlniger I had a few questions here. I renamed a few things to be more consistent with common Netmiko patterns.

netmiko/cloudgenix/cloudgenix_ion.py
carlniger: Not that I am aware of (nothing obvious in docs that I saw or in the CLI when I was poking around), though I could for sure be wrong.

netmiko/cloudgenix/cloudgenix_ion.py
carlniger: This had to do w/ the repainting the command string char by char. The output ends up looking like this: 

```
SOME-HOSTNAME-01# d
SOME-HOSTNAME-01# du
SOME-HOSTNAME-01# dum
SOME-HOSTNAME-01# dump
SOME-HOSTNAME-01# dump
SOME-HOSTNAME-01# dump i
SOME-HOSTNAME-01# dump in
SOME-HOSTNAME-01# dump int
SOME-HOSTNAME-01# dump inte
SOME-HOSTNAME-01# dump inter
SOME-HOSTNAME-01# dump interf
SOME-HOSTNAME-01# dump interfa
SOME-HOSTNAME-01# dump interfac
SOME-HOSTNAME-01# dump interface
SOME-HOSTNAME-01# dump interface
SOME-HOSTNAME-01# dump interface s
SOME-HOSTNAME-01# dump interface st
SOME-HOSTNAME-01# dump interface sta
SOME-HOSTNAME-01# dump interface stat
SOME-HOSTNAME-01# dump interface statu
SOME-HOSTNAME-01# dump interface status
SOME-HOSTNAME-01# dump interface status
SOME-HOSTNAME-01# dump interface status 2
SOME-HOSTNAME-01# dump interface status 2
SOME-HOSTNAME-01# dump interface status 2
Interface   : 2
Device      : eth0
ID      : 15492940782030183
MAC Address : ff:ff:ff:ff:ff:ff
State       : down
Last Change : 2019-04-29 15:54:41.176167357 +0000 UTC
Status      : Not connected
Address     : 172.16.16.15/24
  DNS Server    : 8.8.8.8


SOME-HOSTNAME-01#
```

So for whatever reason the full command gets written three times... so we snag everything after that last command.

netmiko/cloudgenix/cloudgenix_ion.py
carlniger: In general would we do something more like (for example) the PAN enable mode handling:

```
    def enable(self, *args, **kwargs):
        """No enable mode on PaloAlto."""
        pass
```

Just pass instead? And then pass again in `check_enable_mode`?

netmiko/cloudgenix/cloudgenix_ion.py
ktbyers: I think this would be more reliable:

```
        output = output.split(command_string)[-1]
```

i.e. however many times the `command_string` is repainted doesn't matter, just take the very last entry in the list (which should be everything after the last command_string).

This will also work in the failure case i.e. if there is no command_string, then it will just return output.

netmiko/cloudgenix/cloudgenix_ion.py
ktbyers: CloudGenix engineers stated they don't have this; they are working on it.

PR #1217, opened 2019-05-15
Ruckus fastiron does not reliably enter enable mode. #1125

zeropacket:
I fixed the while loop and limit it to only two tries.


ktbyers:
@zeropacket It looks like you need to rebase your code as you are changing other things here (for example the save_config() method was a fix we put in).

PR #1223, opened 2019-05-21
Simple user pattern

User-Man:
Use check_config_mode function can by run from use simple user connection and need find ">" char for end.


carlniger:
Is there a specific platform/version you're having issues with? Seems like this would catch low priv levels with no config permissions on basically every Cisco box.

ktbyers:
Yes, not understanding this PR--what is the use case? Also the pattern would need to be a raw string.

carlniger:
Going to close this since I dont see it solving a current problem. Feel free to reopen w/ more detail if I'm missing something.

PR #1224, opened 2019-05-21
Local pass pwd pattern

User-Man:
Some cisco ask password like "local-pass"


carlniger:
Hi, what platform(s) are prompting for "local-pass"?

User-Man:
It was on cisco me3400, but I found that prompting was by changed manually. And now I think that there is no need for these changes.

PR #1232, opened 2019-06-06
fixing strip_prompt accuracy

vinpk:
Instead of stripping whole line containing the base_prompt, output should be removed anything after the base_prompt, but not before.

Existing Issue: For below output from IOS-XR
```
(Pdb) output
'\nThu Jun  6 15:32:01.736 IST\n===============================================================\nLocation     Altitude Value (Meters)    Source         \n---------------------------------------------------------------\n0/RP0/CPU0            907          RP/0/RP0/CPU0:ECR7#'
```

Reformatted so it is readable:

```

Thu Jun  6 15:32:01.736 IST
===============================================================
Location     Altitude Value (Meters)    Source         
---------------------------------------------------------------
0/RP0/CPU0            907          RP/0/RP0/CPU0:ECR7#
```

part of output `0/RP0/CPU0            907          ` got stripped down; It is not necessary to strip entire line, instead we could strip down the characters from the prompt and after only; keeping the outputs before the prompt. 

Please suggest if any better options to handle this while using `strip_prompt=True`

Cheers

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1153,8 +1153,8 @@ class BaseConnection(object):
         response_list = a_string.split(self.RESPONSE_RETURN)
         last_line = response_list[-1]
         if self.base_prompt in last_line:
-            mod_line = re.sub(self.base_prompt+".*","",last_line).strip()
-            return self.RESPONSE_RETURN.join(response_list[:-1]+[mod_line])
+            mod_line = re.sub(self.base_prompt+".*","",last_line)
+            return self.RESPONSE_RETURN.join(response_list[:-1]+[mod_line]).strip()
         else:
             return a_string
 
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1153,7 +1153,8 @@ class BaseConnection(object):
         response_list = a_string.split(self.RESPONSE_RETURN)
         last_line = response_list[-1]
         if self.base_prompt in last_line:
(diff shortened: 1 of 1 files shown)


ktbyers:
@vinpk In the above, it looks like you have a show command output that has the full prompt in the actual command output (i.e. as part of the returned data and not just trailing the command):

Looking at this data here:

```
Thu Jun  6 15:32:01.736 IST
===============================================================
Location     Altitude Value (Meters)    Source         
---------------------------------------------------------------
0/RP0/CPU0            907          RP/0/RP0/CPU0:ECR7#
```

Is that correct? I assume `RP/0/RP0/CPU0:ECR7#` would happen again on the next line immediately after this command?

vinpk:
That's right.. because of this behaviour from device through ssh channel, the part of output data is getting stripped down..

ktbyers:
Okay, you probably should look to use send_command_timing() for that one command or set strip_prompt=False as an argument to send_command().

Regards, Kirk

vinpk:
Even `send_command_timing()` also has `strip_prompt` argument; and so it's not possible to parse this output if set to `strip_prompt=True`

ktbyers:
Fair enough...add the argument `strip_prompt=False` to your call and then manually strip it post processing.

There is only so much we can do if the device includes the actual prompt in the show command output (i.e. the device really shouldn't do that).

This is a pretty fundamental Netmiko behavior i.e. very likely that changing this behavior would break other things.


Regards, Kirk

PR #1240, opened 2019-06-12
first draft support for pyats/genie

carlniger:
first crack at adding pyats/genie support. works well in testing with ios-xe so far, but *should* work with any Cisco device I assume.

modified reqs to add pyats/genie for 3.4+

added basic test of genie

added 3.7 to tox env

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -29,7 +29,12 @@ from netmiko.ssh_exception import (
     NetMikoTimeoutException,
     NetMikoAuthenticationException,
 )
-from netmiko.utilities import write_bytes, check_serial_port, get_structured_data
+from netmiko.utilities import (
+    write_bytes,
+    check_serial_port,
+    get_structured_data,
+    get_structured_data_genie,
+)
 
 
 class BaseConnection(object):
@@ -1098,6 +1103,7 @@ class BaseConnection(object):
         strip_command=True,
         normalize=True,
         use_textfsm=False,
+        use_genie=False,
     ):
         """Execute command_string on the SSH channel using a delay-based mechanism. Generally
         used for show commands.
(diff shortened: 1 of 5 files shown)


jesse-mcgraw:
Parses for simple "show version" of an NXOS and IOS lab device, though parser fails on IOS-XE 4507

Minor issues:
   Crash if you don't use IP to connect to the device (eg host)
      self.ip = self.ip.strip()
      AttributeError: 'CiscoNxosSSH' object has no attribute 'ip'

  Typo Faield -> Failed

carlniger:
@jesse-mcgraw on the 4507 -- what netmiko device type did you use? If you use "cisco_ios" it will be "ios" to genie and may blow things up, but it *should* work w/ device type "cisco_xe" (or presumably since that works for show version on my available XE device). 

I'll take a peak at the "ip" issue and see what's up. Appreciate you testing this!

jesse-mcgraw:
I tried three separate physical devices just to get a quick sample: cisco_nxos, cisco_ios, cisco_xe

ktbyers:
Okay, the `ip` issue should be fixed in the `develop` branch.

@jesse-mcgraw Can you re-test it with the updated code (if you haven't done so already)?

carlniger:
I just tried against a 3560cx, 819, 881, 1111, n9kv, and XRv device and parser worked ok. I dont have anything modular to test with though so I cant really tell what's going on there. Does it work on that 4507 when ran manually or through Ansible?  Lastly -- typo fixed! Thanks for all the help!

ktbyers:
Both pyats and genie won't install on 32-bit Linux as they distributed only the wheels pypi and only included 64-bit wheels.

ktbyers:
Genie looks like it requires the canonical form the command (i.e. no command abbreviations).

ktbyers:
Should put a check for Windows also since Genie doesn't support Windows. Might need to test what happens on Windows with imports/other behavior to make sure we handle the error in a reasonable way.

ktbyers:
Is there any Genie support for the ASA?

carlniger:
> Is there any Genie support for the ASA?

Looks like limited but yeah: https://pubhub.devnetcloud.com/media/genie-feature-browser/docs/#/parsers

Will try to make updates per other  notes tomorrow!

ktbyers:
I have been working on updates today. So I will push an updated PR out today so let's use that as the new base PR.

ktbyers:
Replaced by https://github.com/ktbyers/netmiko/pull/1285

Issue #795, opened 2018-04-19
Connection timed-out vs connection refused


tjhannin:
Hi,

Not sure if this is a known issue, but it looks like netmiko doesn't differentiate between a real time-out (ie no response) and connection refused, which can make the NetMikoTimeoutException message thrown from establish_connection when it gets a socket.error a bit misleading.

I poked around a bit more, and it seems part of the issue is that paramiko mangles the exceptions so netmiko can't see the actual error codes, but wondering if either the wording on the error message should be changed, or if netmiko should try to find "timed out" from the socket.error to see if the exception was caused by an actual timeout, or if the connection failed for some other reason?

thanks

ktbyers:
@tjhannin Feel free to submit a PR on it.

Alan-R:
I'll submit a PR for this in a few minutes - including specific Python3 support while maintaining Python2 compatibility.

Alan-R:
There is now a clear, clean, Pythonic MR to fix this behavior, which works well in Py2 and Py3.

ktbyers:
@Alan-R Yes, we haven't forgotten about this--just haven't got to it yet.


Regards, Kirk

Alan-R:
Didn't mean this as criticism. This was more for the original poster than you. I also have a driver for the Casa CMTS I've written. Should be sending it along as soon as we have it in production here (probably a few weeks).

ktbyers:
@Alan-R No worries.  Regards, Kirk

PR #1243, opened 2019-06-13
Added support for several kinds of connection exceptions

Alan-R:
In Python3, it uses the native exception types, and in Python2,
it raises NetMikoTimeoutException with text explaining what
kind of underlying exception occurred.

The associated issue is https://github.com/ktbyers/netmiko/issues/795.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -18,6 +18,7 @@ import time
 from collections import deque
 from os import path
 from threading import Lock
+import errno
 
 import paramiko
 import serial
@@ -30,6 +31,12 @@ from netmiko.ssh_exception import (
     NetMikoAuthenticationException,
 )
 from netmiko.utilities import write_bytes, check_serial_port, get_structured_data
+try:
+    assert ConnectionRefusedError
+except NameError:
+    ConnectionRefusedError = NetMikoTimeoutException
+    ConnectionResetError = NetMikoTimeoutException
+    ConnectionAbortedError = NetMikoTimeoutException
 
 
 class BaseConnection(object):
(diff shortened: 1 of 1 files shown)


Alan-R:
This patch changes the behavior of the code in this way for a socket error:

- The text of the messages indicates what kind of socket error triggered the exception, and the text of the original exception is tacked on to the end of the exception text.
- In Python3, the various native socket exceptions are raised except for timeout which remains the NetMikoTimeout. Python2 continues to raise the same (NetMikoTimeout) exception as before for all these conditions.

If the code didn't need to run under Python2, I would have taken a different approach, but this seemed the cleanest way to provide the most detailed and Pythonic result for Python3 and also Python2.

carlmontanari:
This looks super cool to me. Since netmiko is going to be py3.6+ only going forward this could be reswizzled to drop the py2 bits and use f-strings. @ktbyers this look like something we could merge if updated for just py3.6+?

Alan-R:
I mentioned I would have done it differently if it were py3 only. That's still true. But as written, it's just fine for py2 or py3. I have to say that mapping connection denied as timeout is mondo annoying. We would like to try telnet if ssh fails with connection denied, for example...

ktbyers:
Yes, Netmiko `develop` branch is now all Python3.6 or greater.

So Py2 support is now no longer supported (i.e. release 2.4.2 is the last Netmiko PY2 release).

Hopefully, we can look at this in a bit more detail here in about 2 weeks.


Regards, Kirk

Alan-R:
OK. I guess I should redo this. Will try and get to that this week.

PR #1257, opened 2019-06-25
move docs to pdoc3, update readme

carlniger:
```
rm -rf docs
pdoc3 --html --output-dir docs netmiko
```

^ remove old docs, then generate into docs dir w/ pydoc. repo will need to have GitHub pages enabled and set to master branch. testing creating a PDF w/ pandoc to see if its worth it but I kind of think its not going to be very handy as compared to the html which you can get to just by opening index.html locally anyway.


README.md
```diff
@@ -1,116 +1,76 @@
...
 
+
+## Supported Platforms
+
+Netmiko supports a wide range of devices. These devices fall into three categories:
+- Regularly Tested
+- Limited Testing
+- Experimental
+
+Devices in the "Limited Testing" and "Experimental" categories of course have much less testing and are not garuanteed to be 100% functional, so please take care when interacting with those devices.
```
ktbyers: Spelling guaranteed. Also I would change that text as it implies a bit that the regularly tested are guaranteed in some way. Regularly tested means we try to run our full test suite against that set of devices prior to each Netmiko release. Limited testing means the `config` and `show` operation system tests passed against a test on that platform at one point in time so we are reasonably comfortable the driver should generally work. Experimental means that we reviewed the PR and the driver seems reasonable, but we don't have good data on whether the driver fully passes the unit tests or how reliably it works.

carlniger: updated!

README.md
```diff
@@ -1,116 +1,76 @@
...
 [![Downloads](https://pepy.tech/badge/netmiko)](https://pepy.tech/project/netmiko)
+![GitHub contributors](https://img.shields.io/github/contributors/ktbyers/netmiko.svg)
 [![Code Style](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
 
 
 Netmiko
 =======
 
-Multi-vendor library to simplify Paramiko SSH connections to network devices  
+Multi-vendor library to simplify Paramiko SSH connections to network devices
```
ktbyers: I think the line wraps/white-spacing start getting messed up in Markdown if those two spaces at the end of the line go away.

README.md
```diff
@@ -1,116 +1,80 @@
...
+
+## Documentation
+
+<a href="https://ktbyers.github.io/netmiko/docs/netmiko/index.html" title="Docs">API Documentation</a>
+
+Below are some of the particularly handy Classes/functions for easy reference:
+- [Base Connection Object](https://ktbyers.github.io/netmiko/docs/netmiko/base_connection.html)
+- [SSH Autodetect](https://ktbyers.github.io/netmiko/docs/netmiko/index.html#netmiko.SSHDetect)
+- [SSH Dispatcher](https://ktbyers.github.io/netmiko/docs/netmiko/index.html#netmiko.ssh_dispatcher)
+- [Redispatch](https://ktbyers.github.io/netmiko/docs/netmiko/index.html#netmiko.redispatch)
```
ktbyers: I would probably add the SCP file_transfer and some of that in here (probably as the second most common one after BaseConnection.

ktbyers: It would also be good to link to these example/reference code here for this set of common use cases (in some way). Not sure if this is linked to on the readme in any way.

README.md
ktbyers: This might be wrong or not relevant...so you can probably ignore it...

README.md
ktbyers: Meant to add this link: https://github.com/ktbyers/netmiko/tree/develop/examples/use_cases

Issue #1092, opened 2019-02-07
 Request to modify condition: .netmiko.yml to be located in either your home directory or your current working directory


kalmanoharan:
Current requirement specified is that .netmiko.yml file must be in the current directory or in your home directory.

Is there a way to hardcode .netmiko.yml file location in the script similar changing to os.environ['NETMIKO_DIR'] ?

Because its going to be hard to keep a copy of .netmiko.yml for all my team mates home directory.

========================
 Kirk Byers about 20 hours ago 
Yes, the current Netmiko/Netmiko-tools requires .netmiko.yml to be located in either your home directory or your current working directory.

In order to change that you would probably have to make a change in netmiko/utilities.py.

I am open to some sort of a PR or GitHub discussion on the issue, but that is the way it works currently.

Regards, Kirk

=========

carlniger:
see #1258, closing

PR #1258, opened 2019-06-25
netmiko tools cfg file enhancements

carlniger:
support `yaml` or `yml` file suffix, support hidden or normal files, support providing path to netmiko.yaml via `NETMIKO_TOOLS_DIR`

forgot to open pr on this. first pass at addressing #1092

### DIFF
--- a/netmiko/utilities.py
+++ b/netmiko/utilities.py
@@ -2,6 +2,7 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from glob import glob
 import sys
 import io
 import os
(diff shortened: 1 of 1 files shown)


carlniger:
To add some clarity to this. This PR makes netmiko-tools look in the following directories (in this order) for the `netmiko.yaml` file:

1) current directory
2) ~/
3) directory in the `NETMIKO_TOOLS_DIR` environment variable

This will look for the following variations of the file:
`.netmiko.yml`
`.netmiko.yaml`
`netmiko.yml`
`netmiko.yaml`

ktbyers:
I think we should add tests to this if we can do this in a reasonable way.

ktbyers:
Made some updates here:

https://github.com/ktbyers/netmiko/pull/1269

PR #1285, opened 2019-07-13
Integrate Genie and PyATS into Netmiko

ktbyers:


ktbyers:
Need ASA, S300, any other Cisco (Genie supported) platforms wired in.

Need to expand on unit tests, repeat pattern that was used for TextFSM except for only Cisco specific platforms.

ktbyers:
Things we need to document:
1. No windows support.
2. No 32 bit Linux support. Should ask if on the next release they could include the binaries for this.
3. No support for Cisco platforms outside of IOS, IOS-XE, IOS-XR, NX-OS, ASA (as far as I can tell).
4. Have to manually pip install genie and pyats

https://pubhub.devnetcloud.com/media/genie-feature-browser/docs/#/parsers

PR #1286, opened 2019-07-14
Fortinet access profile

zhecho:
Handle Fortinet devices when they are configured with pre-login
        banner:
            FG621B # config system global
            FG621B (global) # set pre-login-banner enable
            FG621B (global) # end


zhecho:
Closed

PR #1295, opened 2019-07-22
Updates to keymile driver

ktbyers:


ktbyers:
@krmnn

Significant changes I made here:

1. In the KeymileNOSSSH removed a lot of the duplicated code in the `_test_channel_read()` that did require I change the base `_test_channel_read()` to return the retrieved data, but I didn't see doing this as breaking anything.

2. In KeymileSSH made the set_base_prompt the standard Netmiko behavior which is to strip off the last character (in this case the '>'). In this case this is really for driver consistency across various Netmiko drivers.

ktbyers:
@krmnn     

Can you explain why this is needed in more detail? What the use case is here?

```
def send_command(self, *args, **kwargs):
        """Make cd /path work"""
        if len(args) >= 1:
            command_string = args[0]
        else:
            command_string = kwargs["command_string"]

        if "cd" in command_string:
            dest = command_string.split(" ", 1)[1]
            self.base_prompt = dest
            kwargs["expect_string"] = dest

        output = super(KeymileSSH, self).send_command(*args, **kwargs)
        return output
```

This also probably relates to the set_base_prompt item that I changed in point 2 above.

ktbyers:
Adding @carlniger also.

krmnn:
> 1. In the KeymileNOSSSH removed a lot of the duplicated code in the `_test_channel_read()` that did require I change the base `_test_channel_read()` to return the retrieved data, but I didn't see doing this as breaking anything.

Perfect, thanks. I just need to detect that string.

> 2. In KeymileSSH simplified set_base_prompt method a bit. I didn't really see why your extra modifications were needed?

Sorry this is leftover from debugging and can be removed.

> Can you explain why this is needed in more detail? What the use case is here?
> def send_command(self, *args, **kwargs):

If I execute 'cd directory' in the Keymile shell, the prompt changes so I tried to set_base_prompt implicitly. 

Tests for both classes still pass after the updates.
Thank you two for your help!

ktbyers:
@krmnn Okay, I don't think we need that update of base_prompt for send_command as Netmiko dynamically finds the prompt in send_command anyways:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/base_connection.py#L1271

Also I wouldn't expect that behavior in Netmiko so it mostly about keeping the driver behavior consistent across platforms.

So I am probably going to strip that out.

ktbyers:
@krmnn I am going to go ahead and merge this. I removed that one send_command behavior and the `cd/set_base_prompt`.

It would be great if you could regression test this against a real device and make sure I didn't break anything.

Thanks for the contribution.

PR #1297, opened 2019-07-22
Junos without root

daanvdsanden:
SCP files to junos devices without shell access. Tested on vQFX en vSRX


ktbyers:
@daanvdsanden Is this very common on Juniper (that you wouldn't have shell access for doing an OS upgrade)?

Sorry for the delay on this...

daanvdsanden:
I never used shell access to do system upgrades on junos. All documentation regarding upgrades are without shell access (and cli only).
I only used the shell on boxes when asked to by JTAC to get certain information that is not available via the CLI. But for "normal" operation shell access isn't necessary. As far as I now it is not very common to use shell during upgrades.
The code was tested on the vQFX

PR #1300, opened 2019-07-24
Update base_connection.py

vineeth122:
Telnet driver unable to read the string `Username`. So updated in the regex pattern.


ktbyers:
Not needed as capitalization is already ignored in the regular expression pattern.

PR #1350, opened 2019-09-04
first swag at adding pydocstyle

carlmontanari:
Lots of changes here, but all cosmetic. It appears `pep8` is basically replaced w/ `pycodestyle` (per Pylama GitHub repo at least) so I switched to that ++ it allows for checking doc length if we wanted to have that be different than "normal" line length.

Then added pydocstyle and ran through it and got it to be quiet when being ran. It doesn't fix linting for consistent *style* but it is a good start. 

Opening this just as a placeholder/conversation for fixing docstrings/pdoc3 output.

### DIFF
--- a/netmiko/__init__.py
+++ b/netmiko/__init__.py
@@ -1,3 +1,4 @@
+"""Multi-vendor library to simplify Paramiko SSH connections to network devices"""
 from __future__ import unicode_literals
 import logging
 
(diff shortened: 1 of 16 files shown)


carlmontanari:
This one is going to take me a loooong time :) will whittle away at at for the next few weeks!

ktbyers:
@carlmontanari Is this PR dead now?

carlmontanari:
maybe/probably? I think its probably worth adding docstring linting and such but would need to be refreshed for sure if it was something you wanted to pull in

ktbyers:
Okay, I will leave it open for now and see if I get to it in the next few months.

PR #1356, opened 2019-09-08
Enable testing for Python 3.8

Jamim:
Hello @ktbyers,

I believe it would be nice to have tests for the upcoming [Python 3.8](https://www.python.org/dev/peps/pep-0569/).

These changes also:
 - split testing into the `Lint` and `Test` stages

Best regards!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,25 @@ language: python
 python:
   - "3.6"
   - "3.7"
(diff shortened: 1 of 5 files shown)


ktbyers:
Minor change to this PR here:

https://github.com/ktbyers/netmiko/pull/1360

PR #1360, opened 2019-09-10
Enable py38 testing

ktbyers:


.travis.yml
```diff
@@ -22,6 +22,6 @@ matrix:
     - python: 3.8-dev
     - python: nightly
 install:
-  - pip install tox-travis
+  - pip install tox==3.13.2 tox-travis==0.12
```
Jamim: Hello @ktbyers,
What's the reason to freeze these packages?

ktbyers: To reduce the CI failures caused by changes in tooling dependencies.

PR #1361, opened 2019-09-11
add telnet,serial for cisco nxos, huawei, hpcomware

xiruofeng:
add telnet,serial for cisco nxos,huawei,hpcomware
deal with paging and ansi ESC codes in 'send_command'
modify save_config,serial_login function for huawei_ssh device

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1266,6 +1266,14 @@ class BaseConnection(object):
         while i <= max_loops:
             new_data = self.read_channel()
             if new_data:
+            
+                # to deal with paging,if "disable_paging" isn't effective ,such as user has no privilege
+                if re.search(r'-+\s*More\s*-+', new_data, re.I):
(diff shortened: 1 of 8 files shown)


ktbyers:
You should split these into separate PRs (for each platform). It will make it more likely that we will be able/willing to merge it.

This is a no-go i.e. we won't support this:

```
                # to deal with paging,if "disable_paging" isn't effective ,such as user has no privilege
                if re.search(r'-+\s*More\s*-+', new_data, re.I):
                    new_data = re.sub(r'-+\s*More\s*-+', repl="", new_data)
                    new_data = self.strip_ansi_escape_codes(new_data)
                    self.write_channel(chr(32))
                # end 
```

If a platform is so messed-up that they don't allow output paging to be disabled then we are very unlikely to support them in Netmiko (it will cause all sorts of problems and reliability issues).

Or if you don't have AAA privileges, then the user will need to obtain the privileges to disable AAA.

PR #1364, opened 2019-09-12
Add UnifiSwitchSSH driver for Ubiquiti UniFi switches

an-empty-string:
This PR adds a driver for Ubiquiti's UniFi switches.

Ubiquiti's UniFi series of switches have a similar CLI to their EdgeSwitches, but they are not usually meant to be configured using the command line (they talk HTTP to a controller host, which provisions multiple switches). As such, SSHing to them doesn't immediately give a CLI that can be used to get information about the switch, it instead lets you configure the connection to the controller (among a few other things):

```
tris.wilson@pc005-linux ~ % ssh admin@switch04
...snip...
switch04-US.v4.0.14# help
UniFi Command Line Interface - Ubiquiti Networks

   info                      display device information
   set-default               restore to factory default
   set-inform <inform_url>   attempt inform URL (e.g. set-inform http://192.168.0.8:8080/inform)
   upgrade <firmware_url>    upgrade firmware (e.g. upgrade http://192.168.0.8/unifi_fw.bin)
   fwupdate --url <firmware_url|firmware_name> [--dl-only] [--md5sum <sum_of_fw>]
            [--keep-firmware] [--keep-running] [--reboot-sys] 
                                   new firmware update command
   reboot                    reboot the device
```

By typing `telnet localhost` at the prompt, you're dropped to a more familiar switch interface. Thus, this driver extends the EdgeSwitch driver and makes the `telnet localhost` change. While it's not useful to make configuration changes at this interface, since they'll be overwritten by the controller, `show` commands still work as expected and provide the same level as usefulness as they would on an EdgeSwitch.

I also added a test script modeled off of the EdgeSwitch test script; all tests pass when running it against an actual UniFi switch.

### DIFF
--- a/netmiko/ssh_dispatcher.py
+++ b/netmiko/ssh_dispatcher.py
@@ -74,6 +74,7 @@ from netmiko.ruckus import RuckusFastironTelnet
 from netmiko.terminal_server import TerminalServerSSH
 from netmiko.terminal_server import TerminalServerTelnet
 from netmiko.ubiquiti import UbiquitiEdgeSSH
+from netmiko.ubiquiti import UnifiSwitchSSH
 from netmiko.vyos import VyOSSSH
 
 
@@ -158,6 +159,7 @@ CLASS_MAPPER_BASE = {
     "ruckus_fastiron": RuckusFastironSSH,
     "ubiquiti_edge": UbiquitiEdgeSSH,
     "ubiquiti_edgeswitch": UbiquitiEdgeSSH,
+    "ubiquiti_unifi": UnifiSwitchSSH,
     "vyatta_vyos": VyOSSSH,
     "vyos": VyOSSSH,
 }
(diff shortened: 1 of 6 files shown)


an-empty-string:
Hi, just wanted to bump this and see if I needed to make any changes / if there were otherwise any objections to this PR.

ktbyers:
So you entirely skipped the configuration tests and only executed the "show tests"--correct?

That is probably fine (assuming you did that).

On quick glance it looks reasonably. Let me see if I can fully review it.

an-empty-string:
I ran all tests -- the config tests still pass as the UniFi switches will allow you to make temporary changes, they will just be overwritten the next time an admin provisions the device from the controller.

It has been a few months so I will run the tests again on a device with latest firmware as well.

ktbyers:
Sounds good. Let me know how that test re-run goes.


Thanks, Kirk

ktbyers:
Updated the PR here:

https://github.com/ktbyers/netmiko/pull/1657

I added a `cleanup` method as I thought we would need an extra `exit` to exit from the `telnet localhost`. Let me know if this is incorrect in some way.

PR #1376, opened 2019-09-24
Eltex ESR driver

ktbyers:


ktbyers:
@Uburro I made some updates to this driver--generally to make the behavior more consistent with standard Netmiko behavior.

I converted `_confirm` and `_restore` methods to be private methods as these are specific to Eltex ESR driver, but not standard in Netmiko.

Let me know if you see any issues that I caused with my changes.

ktbyers:
@Uburro I also removed the `exit_config_mode` as it looked like the method it inherits from should work without any issues (so I didn't any reason to override it).

I also removed the `save_config` method as the default Netmiko behavior for drivers with `commit` is to specify `save_config` is not implemented.

ktbyers:
This PR should be ready to go (pending with @Uburro wants to do some more testing).

Should be merged on 2019-10-01 (i.e. give @Uburro some time to respond), but otherwise go ahead and merge.

PR #1381, opened 2019-09-26
Updated Dlink pull-request

ktbyers:


ktbyers:
@gmassawe Updated DLink PR is here.

ktbyers:
I made some updates:
1. It looks like there are ansi escape codes in the output (based on earlier testing output post). So I stripped those.
2. Added disable_paging() method to the driver.
3. Changed some of the default behavior for config mode and enable mode methods to be more consistent with Netmiko (and what I thought it should be).
4. Made save_config() call parent method
5. Removed out that custom strip_prompt(). Netmiko pattern of behavior is strip the command echo, but DLink does this:

```
DGS–3426:5#show account
Command: show account

Current Accounts:
Username Access Level
--------------- ------------
dlink Admin
Total Entries: 1
DGS–3426:5# 
```

So they actually include the command as part of the output...so Netmiko typically wouldn't strip this out (this is in addition to the normal command echo which is Netmiko does strip out).

It is possible the unit tests might fail because of that...so we would have to look at that if they do.

Converted telnet enter over to `\r\n` (as that should be the typical/standard form for telnet enter).

ktbyers:
@gmassawe any chance to re-run the unit tests against this?

Fikirsiz:
@ktbyers I ran tests and receive one error with test_netmiko_show.py::test_strip_command FAILED because this device returns command with output.
[config_test.txt](https://github.com/ktbyers/netmiko/files/4130003/config_test.txt)
[show_test.txt](https://github.com/ktbyers/netmiko/files/4130004/show_test.txt)


The output from a device:
myswitch1:5# show switch 
Command: show switch
Device Type                       : DGS-1210-28/ME

ktbyers:
Okay, are there any show commands that don't do this:

```
myswitch1:5# show switch

Command: show switch
Device Type : DGS-1210-28/ME
```

i.e. that don't do that extra `Command: <command syntax>`

it is good that it is just a testing issue (i.e. that the driver looks to be working correctly, but would be also good to get the tests 100% passing).


Regards,

Kirk

Fikirsiz:
Unfortunately, I didn't find command for disabling "Command: " in output. Also, you can see in file config_test.txt another FAILED because Dlink doesn't have config mode and check_config_mode always return False.

ktbyers:
Okay, the config items we should be able to fix in the driver. We just have to make the driver consistent with what the test expects for a driver with no config mode.

dynax60:
Hey, guys, I did some changes for tests in the new pull request #1588 (dunno how to merge it to this pull request). Meanwhile I didn't pass linter check for some reason. I have to check it tomorrow, I need to sleep now.

dynax60:
Everything seems to be ok now. Test someone on your D-Link hardware. For the ssh test you need to add ~/.ssh/config something like that (for old hardware):
```
Host 192.168.50.10
    KexAlgorithms +diffie-hellman-group1-sha1
    Ciphers +aes256-cbc
```

Kirk, I don’t remember whether I said or not, but D-Link hardware has two types of CLI: their own (D-Link) and very similar to IOS. I just remember there were doubts about how to name the type of device. It’s probably suitable for them to use a driver **cisco_ios**. I have nothing to try.

About config mode: there is a command ``enable admin`` to gain access to admin level privileges. You can cofigure it. So, this is an example of logged user 'test' with _user level_ **operator**:

```
DGS-3120-24TC:operator#enable admin
Command: enable admin

PassWord: ******
Success.

DGS-3120-24TC:admin#enable admin
```

So, the word after `:' is a user level type. It might be: user|admin|operator|power_user. Privileged mode is when the user level is admin. We need to add config mode feature.

dynax60:
> Unfortunately, I didn't find command for disabling "Command: " in output. Also, you can see in file config_test.txt another FAILED because Dlink doesn't have config mode and check_config_mode always return False.

I'm agreed, there's no command to disable "Command:" line that is right after user's input. But D-Link have config mode though, it's a ``enable admin``.

ktbyers:
Okay, let me read through this and your test updates and then circle back to it. Thanks for your work here @dynax60

dynax60:
I take my words back about config mode.  I mixed up config mode with enable mode ;) Sorry. Anyway, tests are fine now.

Vadims06:
is there any updates on this?

ktbyers:
I still need to test/review it. I also need to rebase/fix the merge conflicts, but that should be pretty easy/quick.

dynax60:
> I still need to test/review it. I also need to rebase/fix the merge conflicts, but that should be pretty easy/quick.

Hello, Kirk! Could you please merge my pull request #1588 with the dlink branch? My tests results is in this pull request. I also look forward to including the dlink branch in the next version of Netmiko.

Anybody, could someone else test it with your own hardware? I would be grateful to you!

dynax60:
> is there any updates on this?

ping. Could you please test the last updates?

dynax60:
> Unfortunately, I didn't find command for disabling "Command: " in output. Also, you can see in file config_test.txt another FAILED because Dlink doesn't have config mode and check_config_mode always return False.

Could you please test my last update? See pull request #1588

Vadims06:
> > Unfortunately, I didn't find command for disabling "Command: " in output. Also, you can see in file config_test.txt another FAILED because Dlink doesn't have config mode and check_config_mode always return False.
> 
> Could you please test my last update? See pull request #1588

Ok, I can test. How can I install your code into my env? Actually, I have never tested Pull Requests before.

dynax60:
> Ok, I can test. How can I install your code into my env? Actually, I have never tested Pull Requests before.

See https://github.com/ktbyers/netmiko/blob/develop/TESTING.md

You can do it like this, for example:

```
virtualenv -p python3.6 ~/virtualenv/python3.6
source ~/virtualenv/python3.6/bin/activate
pip install
git clone https://github.com/dynax60/netmiko
cd netmiko
pip install -r requirements-dev.txt
git checkout dlink
cd tests
cp etc/commands.yml.example etc/commands.yml
cp etc/responses.yml.example etc/responses.yml
cp etc/test_devices.yml.example etc/test_devices.yml
```

Edit these .yml files to suit your test device (see dlink_ds in case of ssh and dlink_ds_telnet for telnet) and run tests:
```
py.test -v test_netmiko_config.py --test_device dlink_ds
py.test -v test_netmiko_config.py --test_device dlink_ds_telnet
```

You can then post the output of your test here. Don't forget to include your commands.yml and responses.yml here as well. 

Thank you very much!

PR #1384, opened 2019-09-30
Updated Ciena Driver

ktbyers:


ktbyers:
This is probably the main method that needs re-tested:

```
    def remote_space_available(self, search_pattern=""):
```

To make sure I didn't break anything.

ktbyers:
@daanvdsanden Let me know if you can re-test these updates (especially the secure copy tests).

Mainly I wanted to eliminate a lot of the index references and clarify on the longest_match as it was hard to follow in the code.

I also converted over to F-strings (as only Py3.6 or greater support in the develop branch now).

ktbyers:
This is good-to-go from my perspective and I will merge on 10/6 (to give @daanvdsanden time to test).

ktbyers:
Adding @georgesnow to this if he wants to regression test.

georgesnow:
@ktbyers I will try and test this week. I do know the previous SCP in @daanvdsanden didn't work on my SAOS 6.15 box. I don't actually know why. I know off hand I had to modify "remote_md5_cmd" formatting and I had to add a function for scp_get_file in the subclass in order to get work for me.

georgesnow:
I just took quick look (remembered a little more) so the formatting was one of the issues I had. the / (slashes) are included in the string values that are getting passed into the command. in case of  @daanvdsanden code  I would end up with // (double slashes)

example output:
//filesystem//folder//whatever.txt 

instead of:
/filesytem/folder/whatever.txt

Also the command on Ciena boxes required me to use "file scp" depending if the user was dropping into a shell. If the user privileges are set to "diag" level. They can get a shell prompt and perform the md5 hash, but super, admin etc... cannot. And when the netmiko driver attempts to do a SCP it doesn't appear to drop into a shell to perform the SCP transfer. It performs in the CLI interface. However, the md5 hash seems to be performed in a shell prompt

ktbyers:
Okay the first part of that should be easy to fix (the extra slashes).

The second part you are saying the Secure Copy commands will fail if you don't have the required permissions to drop in the shell? Is there a workaround for this where the methods still work and you aren't required to drop into the shell?

ktbyers:
It would also be helpful for me to see the output of the following command: `file vols` from an actual Ciena device so I can verify my modifications work properly.

georgesnow:
3916 (the 5160s have 120GB of available space):

Filesystem           1K-blocks      Used Available Use% Mounted on
ubi0:appA               121352     98688     22664  81% /
devtmpfs                237492         0    237492   0% /dev
run                       1024        48       976   5% /run
tmpfs                   131072       604    130468   0% /tmp
var                       1024       408       616  40% /var
shm                      10240      1664      8576  16% /dev/shm
tmpfs                     1024       456       568  45% /etc
ubi0:config              29152      2988     24664  11% /mnt/sysfs
ubi0:logs                29152        92     27560   0% /mnt/log
mtd:kernel-a             16384      4408     11976  27% /mnt/kernelA
mtd:kernel-b             16384      4408     11976  27% /mnt/kernelB
ubi0:appA               121352     98688     22664  81% /mnt/imageA
ubi0:appB               121352     98716     22636  81% /mnt/imageB

ktbyers:
@georgesnow Okay, great...thanks!

georgesnow:
so that's the output for "file vols"

No, better way to describe is SCP can be performed at the CLI level (with super permissions) so **without** dropping into a shell, but requires the "file" modifier in the beginning of the command.

Example:
test-ce> file scp <TAB>
    [source] [destination]
test-ce> file scp

and when I tested with @daanvdsanden code it seemed to be performing at the CLI level **not** the shell, which should allow you to use regular linux formatted commands. 

And that's why I have that other function in my code for SCP using the CLI method for SCP.

does that make more sense?

ktbyers:
I think so...let me try to update with no dropping into the shell and everything at the CLI. I will then need you to regression test it.

Sound good?

georgesnow:
yup, sounds good!

ktbyers:
Looking at the code...I don't think there should be any double slashes i.e. it should all be single-slash.

I will update the other issue and then let me know if you are still seeing the double-slash problem.

ktbyers:
So it looks like the only place we drop into the Shell is for the MD5 command.

Is there a way to do the MD5 command without dropping into the shell?

The SCP is from the server to the network device (i.e. the server initiates the SCP for both gets and puts so `file scp` is never executed on the networking device.

Here is what we do for the MD5:

```
        remote_md5_cmd = f"{base_cmd} {self.file_system}/{remote_file}"

        self.ssh_ctl_chan._enter_shell()
        dest_md5 = self.ssh_ctl_chan.send_command(remote_md5_cmd, expect_string=r"[$#]")
        self.ssh_ctl_chan._return_cli()
```

Where base_cmd = `md5sum`

So `md5sum /tmp/users/my_user/my_file`


Let me know if that makes sense or not...

daanvdsanden:
I'm testing the code now (and looking how I was able to successful transfer a file without dropping to shell, which was my intention). Also when using file scp the ciena switch will be the ssh client, and the code I submitted, netmiko is the client and the switch is the server.

ktbyers:
@daanvdsanden Yes, that is correct...the network device must be the SCP server and the remote machine is the SCP client.

georgesnow:
Ok that’s my bad. I misunderstood how it was implemented for SCP. And from what I can tell on the ciena docs the only way to do md5sum is from a shell prompt. I can double check that tomorrow

georgesnow:
I can double check the md5sum but pretty sure that was shell only based on ciena docs. I can check tomorrow. 

As far the scp. I will need look at what I did and how the proper implementing on. Cause I think I misunderstood that part

daanvdsanden:
> So it looks like the only place we drop into the Shell is for the MD5 command.
> 
> Is there a way to do the MD5 command without dropping into the shell?

I went through the documentation of ciena (the command reference guide) and I was unable to find any command that would enable me to calculate an MD5 hash without dropping to shell.

I see three options to get around this
1. Disable the md5sum when opening a shell is not possible. The correct rights can checked as follows:
```
SAOS is True Carrier Ethernet TM software.
Welcome to the shell.
ciena-switch> user whoami
username: su  access-level: super
ciena-switch> diag shell
SHELL PARSER FAILURE: 'diag' - no matching entry found
ciena-switch> user relogin-as user ciena
Password: 

SAOS is True Carrier Ethernet TM software.
Welcome to the shell.
ciena-switch> diag shell
ciena@ciena-switch:~$ exit
logout
ciena-switch> 
```
2. Change the default behavior for ciena devices to disable md5sum checking
3. Document the behavior and leave it up to the user to disable md5 checking

ktbyers:
It really needs to be option 3 (from my perspective)--default behavior will require it and user would have to override that behavior.

I say this as the expected default behavior in many contexts is an indempotent transfer which requires the MD5 to operate--so I would probably be unwilling to change that default behavior for a particular driver...i.e. the end-user would be forced to overwrite the behavior (or get the required permissions).

An end-user can easily overwrite the behavior in the file_transfer function call by using different arguments.

ktbyers:
So I think this is implemented right...if people want to test it.

daanvdsanden:
I did some testing on different platforms and versions on all platforms I have at my disposal it works, except for one platform.  On the 5170 it is not working since the output of `file vols` is using two lines for certain file systems:
```
Filesystem           1K-blocks      Used Available Use% Mounted on
overlay               20511356    581780  18864616   3% /
tmpfs                  8172616         0   8172616   0% /dev
tmpfs                  8172616         0   8172616   0% /sys/fs/cgroup
/dev/mapper/EN--VOL-apps
                      84566512   1488828  78758864   2% /mnt/rel
tmpfs                     1024        52       972   5% /run
tmpfs                  1048576       648   1047928   0% /tmp
/dev/mapper/EN--VOL-log
                       5029504     55248   4695728   1% /var/crash
/dev/dm-0              1998672    902400    975032  48% /lib/modules
shm                      65536        12     65524   0% /dev/shm
/dev/mapper/EN--VOL-config
                      20511356    581780  18864616   3% /mnt/config
/dev/mapper/EN--VOL-log
                       5029504     55248   4695728   1% /mnt/logs
/dev/mapper/EN--VOL-config
                          4096      1476      2620  36% /etc/resolv.conf
/dev/mapper/EN--VOL-config
                          4096      1476      2620  36% /etc/hostname
/dev/mapper/EN--VOL-config
                          4096      1476      2620  36% /etc/hosts
tmpfs                     1024        60       964   6% /var/run
/dev/mapper/EN--VOL-secure
                         95054      1571     86315   2% /mnt/config/saos/cert
/dev/mapper/EN--VOL-secure
                         95054      1571     86315   2% /mnt/config/saos/ssh
/dev/mapper/EN--VOL-secure
                         95054      1571     86315   2% /mnt/config/saos/secure
run                       1024        52       972   5% /run
/var/run                  1024        60       964   6% /var/run
/etc                      4096      1476      2620  36% /etc
tmpfs                     4096         0      4096   0% /var/taskstats
/dev/loop0                 128       128         0 100% /ciena/npu_ucode
```

So after reading the documentation I'm seeing we can also add the volume to the command:
image

So longest match is not needed at all O:-) , during testing I saw that the command `file vols` is passed to df so with `f"file vols -P {self.file_system}"`, this is however not documented (and thus I'm not sure if it will stay):
```
ciena-5170> file vols /tmp
Filesystem           1K-blocks      Used Available Use% Mounted on
tmpfs                  1048576       560   1048016   0% /tmp
ciena-5170> file vols /etc/hostname
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/mapper/EN--VOL-config
                          4096      1476      2620  36% /etc/resolv.conf
ciena-5170> file vols /etc/asdf
Filesystem           1K-blocks      Used Available Use% Mounted on
df: /etc/asdf: can't find mount point
ERROR: df -k command failed
ciena-5170> file vols -P /etc/hostname
Filesystem           1024-blocks    Used Available Capacity Mounted on
/dev/mapper/EN--VOL-config      4096      1476      2620  36% /etc/resolv.conf
```

daanvdsanden:
I tested it successfully on the ciena 3905,3916,3928,3930,5142,5160
For the 5170 see the previous comment.

The 8700 (chassis based) has a different prompt when entering shell mode. This can be fixed by changing `expect_string=r"[$#]"` into `expect_string=r"[$#]|P>"` on line 29 and 199.

ktbyers:
Okay, let me look at the notes you added into this.

georgesnow:
Presumably based on @daanvdsanden testing i am expecting it to work. Wonder if Part of my issue was user permissions issue between diag and super? I had tested with a super level. But I know the double slash was issue cause the command history on the switch is how I figured out why it wasn’t working for me. Either I will re-test on on my gear. All I care about is that it works

ktbyers:
Does this work:

```
# where my_user is the user that connected
file vols -P /tmp/users/my_user      
```

I say this as `/tmp/users/my_user` will be the default directory we specify (self.file_system).

ktbyers:
Okay, I think I fixed the two issues (in the last commit).

The line-wrapping issue, I just ran the output through a regex preprocessor so if it sees the line-wrap problem it just removes the extra newline (i.e. it normalizes the data to make it the standard format we expect).

I updated the shell prompt character check to be "r[$#>]" which looked like it should work.

georgesnow:
> Does this work:
> 
> ```
> # where my_user is the user that connected
> file vols -P /tmp/users/my_user      
> ```
> 
> I say this as `/tmp/users/my_user` will be the default directory we specify (self.file_system).

this is the output on 5160:


Screen Shot 2019-10-01 at 2 03 48 PM

ktbyers:
Okay, I will switch to that as it is better solution than the longest match pattern and if there is an issue in the future (with this command), we will adjust then

georgesnow:
@ktbyers and @daanvdsanden do you have a set of test commands and responses? I ran this against me 5150. I got a bunch of issues


test_netmiko_scp.py::test_scp_put FAILED                                         [  8%]
test_netmiko_scp.py::test_remote_space_available ERROR                           [ 16%]
test_netmiko_scp.py::test_local_space_available PASSED                           [ 25%]
test_netmiko_scp.py::test_verify_space_available_put FAILED                      [ 33%]
test_netmiko_scp.py::test_remote_file_size FAILED                                [ 41%]
test_netmiko_scp.py::test_md5_methods FAILED                                     [ 50%]
test_netmiko_scp.py::test_disconnect PASSED                                      [ 58%]
test_netmiko_scp.py::test_verify_space_available_get ERROR                       [ 66%]
test_netmiko_scp.py::test_scp_get ERROR                                          [ 75%]
test_netmiko_scp.py::test_md5_methods_get ERROR                                  [ 83%]
test_netmiko_scp.py::test_disconnect_get ERROR                                   [ 91%]
test_netmiko_scp.py::test_file_transfer FAILED                                   [100%]

georgesnow:
just a side note: I used the same set of commands and responses that worked with my code.    🤷‍♂️

ktbyers:
Okay, I converted over to that simpler disk usage format i.e. `file vols -P {self.file_system}`

ktbyers:
Here is what @daanvdsanden posted in the earlier PR:

```
# commands.yml
ciena_saos:
  version: "software show"
  basic: "interface show"
  extended_output: "configuration show"
  config:
    - "system set time-offset 7200" # base command
    - "system set timestamp local"
    - "system set time-offset 3600" # something you can verify has changed
  config_verification: "configuration search running-config string \"set time-offset\""
  config_file: "ciena_saos_commands.txt"
  save_config_cmd: "configuration save"
  save_config_confirm: False
  save_config_response: False
```

```
# responses.yml:
ciena:
  base_prompt: router1
  router_prompt: router1>
  enable_prompt: router1>
  version_banner: saos-
  interface_ip: 192.0.2.1
  multiple_line_output: "NETCONF CONFIG"
  file_check_cmd: "system set time-offset 3600"
  scp_remote_space: 10000
```

ktbyers:
@georgesnow I would isolate with a single test:

```
test_netmiko_scp.py::test_scp_put FAILED [ 8%]
```

And get the error from what is occurring there and you can post that exception here and we can see why it is failing.

ktbyers:
This would usually be run as follows:

` py.test -s -v test_netmiko_scp.py::test_scp_put --test_device ciena`

georgesnow:
$ py.test -s -v test_netmiko_scp.py::test_scp_put --test_device cienalab
================================= test session starts ==================================
platform linux -- Python 3.6.8, pytest-5.1.0, py-1.8.0, pluggy-0.12.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/gfn4/netmiko-cienatests/netmiko, inifile: setup.cfg
collected 1 item

test_netmiko_scp.py::test_scp_put FAILED

======================================= FAILURES =======================================
_____________________________________ test_scp_put _____________________________________

scp_fixture = (<netmiko.ciena.ciena_saos.CienaSaosSSH object at 0x7f5746a83358>, <netmiko.ciena.ciena_saos.CienaSaosFileTransfer object at 0x7f574653ca90>)

    def test_scp_put(scp_fixture):
        ssh_conn, scp_transfer = scp_fixture
        if scp_transfer.check_file_exists():
            assert False
        else:
>           scp_transfer.put_file()

test_netmiko_scp.py:28:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../netmiko/scp_handler.py:314: in put_file
    self.scp_conn.scp_transfer_file(self.source_file, destination)
../netmiko/scp_handler.py:37: in scp_transfer_file
    self.scp_client.put(source_file, dest_file)
/usr/local/lib/python3.6/dist-packages/scp.py:166: in put
    self._send_files(files)
/usr/local/lib/python3.6/dist-packages/scp.py:271: in _send_files
    self._send_file(fl, name, mode, size)
/usr/local/lib/python3.6/dist-packages/scp.py:281: in _send_file
    self._recv_confirm()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <scp.SCPClient object at 0x7f5746559470>

    def _recv_confirm(self):
        # read scp response
        msg = b''
        try:
            msg = self.channel.recv(512)
        except SocketTimeout:
            raise SCPException('Timeout waiting for scp response')
        # slice off the first byte, so this compare will work in py2 and py3
        if msg and msg[0:1] == b'\x00':
            return
        elif msg and msg[0:1] == b'\x01':
>           raise SCPException(asunicode(msg[1:]))
E           scp.SCPException: scp: /tmp/users/ciena/test9.txt: No such file or directory

/usr/local/lib/python3.6/dist-packages/scp.py:363: SCPException
================================== 1 failed in 9.11s ===================================

georgesnow:
ugh that was ugly

georgesnow:
so it can't find the file

georgesnow:
so this path is incorrect:
/tmp/users/ciena/test9.txt: 

I believe it should be: 
/tmp/users/$WHOAMI/test9.txt

where $WHOAMI = the user account

daanvdsanden:
@georgesnow just to double check scp server is enabled?
```
> system server scp show 

+----------- SCP SERVER ------------+
| Admin State | Enabled             |
| Oper State  | Enabled             |
+-------------+---------------------+
```

Can be enabled with `system server scp enable`

georgesnow:
but you can't change directories on Ciena box without the "file" modifer unless you drop into a shell

georgesnow:
nornir-ll> system server scp show

+----------- SCP SERVER ------------+
| Admin State | Enabled             |
| Oper State  | Enabled             |
+-------------+---------------------+

nornir-ll>

ktbyers:
@georgesnow I didn't understand this statement?

```
so this path is incorrect:
/tmp/users/ciena/test9.txt:

I believe it should be:
/tmp/users/$WHOAMI/test9.txt

where $WHOAMI = the user account
```

Those two should be identical--the user account and $WHOAMI should be the same?

i.e. we are using the username that you connect with via SSH for this.

georgesnow:
I also confirmed the file is **not** on the ciena switch

georgesnow:
users name is grabbed from the test_devices.yml correct?

georgesnow:
the test used a username of "ciena" but in my test_devices.yml the username is defined as "nornir"

ktbyers:
One second...

daanvdsanden:
In the file `tests/conftest.py` in line 434 I have put in the username hardcoded (where the test fixtures are defined :-( )..

georgesnow:
#1569953703
md5sum /tmp/users/ciena/test9.txt
#1569953704
exit
#1569953711
md5sum /tmp/users/ciena/test9.txt
#1569953712
exit


**from the ciena switch**

georgesnow:
AH seriously ok thanks @daanvdsanden

ktbyers:
Yes, some of these tests are hard...(especially having a bunch of flexibility in them is hard):

So as per earlier this is defined in conftest.py (but it sounds like it already was worked out):

```
433         "ciena_saos": {
434             "file_system": "/tmp/users/ciena",
435             "enable_scp": False,
436             "delete_file": delete_file_ciena_saos,
437         },
438     }
```

ktbyers:
Also be careful as this delete file operation is going to execute as part of the test:

```python
def delete_file_ciena_saos(ssh_conn, dest_file_system, dest_file):
    """Delete a remote file for a ciena device."""
    full_file_name = "{}/{}".format(dest_file_system, dest_file)
    cmd = "file rm {}".format(full_file_name)
    output = ssh_conn.send_command_timing(cmd, strip_command=False, strip_prompt=False)
    return output
```

ktbyers:
So did the test work now?

georgesnow:
ok failed the first time even though I updated the conftests.py 

but now its working. we are getting somewhere: 

 py.test -s -v test_netmiko_scp.py::test_scp_put --test_device cienalab
================================= test session starts ==================================
platform linux -- Python 3.6.8, pytest-5.1.0, py-1.8.0, pluggy-0.12.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/gfn4/netmiko-cienatests/netmiko, inifile: setup.cfg
collected 1 item

test_netmiko_scp.py::test_scp_put PASSED

================================== 1 passed in 7.64s ===================================

georgesnow:
re-ran the full test: (space one I think I know why - the 5160 has 120GB and the test # is to small)

test_netmiko_scp.py::test_scp_put PASSED
test_netmiko_scp.py::test_remote_space_available ERROR
test_netmiko_scp.py::test_local_space_available PASSED
test_netmiko_scp.py::test_verify_space_available_put FAILED
test_netmiko_scp.py::test_remote_file_size PASSED
test_netmiko_scp.py::test_md5_methods PASSED
test_netmiko_scp.py::test_disconnect PASSED
test_netmiko_scp.py::test_verify_space_available_get FAILED
test_netmiko_scp.py::test_scp_get PASSED
test_netmiko_scp.py::test_md5_methods_get PASSED
test_netmiko_scp.py::test_disconnect_get PASSED
test_netmiko_scp.py::test_file_transfer FAILED

georgesnow:
ok that didn't fix it

daanvdsanden:
> I updated the shell prompt character check to be "r[$#>]" which looked like it should work.

This won't work. If I don't have the right to get into diag the prompt still matches because the `>` at the end of the cli prompt, than the script will try to execute the md5sum command. This will fail on the wrong place, instead of not finding the expected prompt.

I did some more testst  on the 8700 (also logged in on the backup CTX card via ssh) and we have the following prompts:

1. Primary CTX cli: `hostname>`
2. Primary CTX shell: `hostnameP>`
3. Secondary CTX cli: `hostname-S>`
4. Secondary CTX shell: `hostnameS>`

for all the other cienas i can test:
1. cli: `hostname>`
2. shell: `username@hostname:~$`

I'm going to break my head on a rexexp for mathing when in the shell and when not.

georgesnow:
hmmm.. ok yeah I had heck of a time getting it working to match the right prompts.

I thought it was my yml file cause I did direct copy and paste, but it still isn't working....
🤯

georgesnow:
I thought I modified your regex in my code to get it working. let me look....

georgesnow:
yeah so my function for remote_space_available is this:



`    def remote_space_available(self, search_pattern=""):
        """Return space available on Ciena SAOS"""
        remote_cmd = "file vols"
        remote_output = self.ssh_ctl_chan.send_command_expect(remote_cmd)

        # Try to ensure parsing is correct:
        # Filesystem           1K-blocks      Used Available Use% Mounted on
        # var                       1024       528       496  52% /var
        remote_output = remote_output.strip()
        output_lines = remote_output.splitlines()

        # First line is the header; second is the actual file system info
        header_line = output_lines[0]
        filesystem_line = output_lines[1]

        if "Filesystem" not in header_line or "Avail" not in header_line.split()[3]:
            # Filesystem  1K-blocks  Used   Avail Capacity  Mounted on
            msg = "Parsing error, unexpected output from {}:\n{}".format(
                remote_cmd, remote_output
            )
            raise ValueError(msg)

        longest_match = (0, 0)
        for filesystem_line in output_lines[1:]:
            mounted_on = filesystem_line.split()[5]
            if (
                self.file_system.startswith(mounted_on)
                and len(mounted_on) > longest_match[0]
            ):
                longest_match = (len(mounted_on), filesystem_line.split()[3])
        space_available = longest_match[1]

        if not re.search(r"^\d+$", space_available):
            msg = "Parsing error, unexpected output from {}:\n{}".format(
                remote_cmd, remote_output
            )
            raise ValueError(msg)

        return int(space_available) * 1024`

georgesnow:
actually looking at my class for file transfer is different, but it definitely works on 3916, 5150, and 5160s. 

you can take a look:

https://github.com/georgesnow/netmiko/blob/cc8b62f8b4ad00b6970ccccd3e2f98239849f48c/netmiko/ciena/ciena_saos_ssh.py#L79

daanvdsanden:
> I'm going to break my head on a rexexp for mathing when in the shell and when not.

The following regexp  `r"[$#]|P>|(?<!-)S>"` (we could loose the #) test all the cases (I know of now). I have time to test again on friday.

https://regex101.com/r/bBUcZM/1

georgesnow:
yeah so I re-ran mine. I noticed I modified the conftest.py and removed "ciena" from the hard coding it made it bit more dynamic. and it passes on all the tests. @daanvdsanden you willing to give it go on your gear? then maybe we merge @daanvdsanden code and mine?

ktbyers:
What happens when you fail to get into the `shell` is there any message output from the Ciena device?

georgesnow:
nornir-ll> diag shell
SHELL PARSER FAILURE: 'diag' - no matching entry found
nornir-ll>

ktbyers:
@georgesnow We need to work on the next test that is failing:

```
test_netmiko_scp.py::test_remote_space_available ERROR
```

With this current PR (i.e. what is failing in the current code).

georgesnow:
for SCP nothing was failing. the other config stuff failed

ktbyers:
So for SCP nothing is failing with this current Pull Request / code?

georgesnow:
yeah, so you have to modify conftest.py and remove the "ciena" in the path

and just replace the file transfer class from mine into @daanvdsanden 




just ran this:


py.test -s -v test_netmiko_scp.py --test_device cienalab
================================= test session starts ==================================
platform linux -- Python 3.6.8, pytest-5.1.0, py-1.8.0, pluggy-0.12.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/gfn4/netmiko-githubpull/netmiko, inifile: setup.cfg
collected 12 items

test_netmiko_scp.py::test_scp_put PASSED
test_netmiko_scp.py::test_remote_space_available PASSED
test_netmiko_scp.py::test_local_space_available PASSED
test_netmiko_scp.py::test_verify_space_available_put PASSED
test_netmiko_scp.py::test_remote_file_size PASSED
test_netmiko_scp.py::test_md5_methods PASSED
test_netmiko_scp.py::test_disconnect PASSED
test_netmiko_scp.py::test_verify_space_available_get PASSED
test_netmiko_scp.py::test_scp_get PASSED
test_netmiko_scp.py::test_md5_methods_get PASSED
test_netmiko_scp.py::test_disconnect_get PASSED
test_netmiko_scp.py::test_file_transfer PASSED

============================ 12 passed in 69.97s (0:01:09) =============================

ktbyers:
We aren't rolling back to the old code.

There are several items in the old code that aren't acceptable to me and reverting back will just require I fix them again...and then we will be right back here.

georgesnow:
well I am willing to go back and reapply your changes to my code. does that work?

ktbyers:
@daanvdsanden I think I fixed the 8700 prompt issue with the last commit:

```diff
$ git diff
diff --git a/netmiko/ciena/ciena_saos.py b/netmiko/ciena/ciena_saos.py
index 17e20c8..60b4d2c 100644
--- a/netmiko/ciena/ciena_saos.py
+++ b/netmiko/ciena/ciena_saos.py
@@ -26,7 +26,11 @@ class CienaSaosBase(BaseConnection):
 
     def _enter_shell(self):
         """Enter the Bourne Shell."""
-        return self.send_command("diag shell", expect_string=r"[$#>]")
+        output = self.send_command("diag shell", expect_string=r"[$#>]")
+        if "SHELL PARSER FAILURE" in output:
+            msg = "SCP support on Ciena SAOS requires 'diag shell' permissions"
+            raise ValueError(msg)
+        return output

```

Basically I search for that error condition and detect it and raise an exception. We probably want to do this anyways (for clarity to users so they can see this issue and know what the cause was).

georgesnow:
that matches on both > and $

I used this in my code for the regex:
expect_string=r"[\$#]"

georgesnow:
https://regex101.com/r/d2KYF9/1

ktbyers:
@georgesnow Yes, I am intentionally trying to match on either `>`, `#`, or `$`. I wanted to avoid the regex or expression intentionally (if we can).

The shell prompt for Ciena 8700 uses a `>`

georgesnow:
my mistake. it got a little confusing

daanvdsanden:
> Basically I search for that error condition and detect it and raise an exception. We probably want to do this anyways (for clarity to users so they can see this issue and know what the cause was).

I really like this solution the best raising an exception when it fails. I'll test the latest commits on friday.

georgesnow:
I just ran the latest commit. still failed on:
test_verify_space_available_get

___________________________ test_verify_space_available_get ____________________________

scp_fixture_get = (<netmiko.ciena.ciena_saos.CienaSaosSSH object at 0x7f73361a0588>, <netmiko.ciena.ciena_saos.CienaSaosFileTransfer object at 0x7f73361af8d0>)

    def test_verify_space_available_get(scp_fixture_get):
        ssh_conn, scp_transfer = scp_fixture_get
        assert scp_transfer.verify_space_available() == True
        # intentional make there not be enough space available
        scp_transfer.file_size = 100000000000
>       assert scp_transfer.verify_space_available() == False
E       assert True == False
E         -True
E         +False

test_netmiko_scp.py:80: AssertionError

ktbyers:
@georgesnow Okay, let me look at it.

georgesnow:
but everything else passed! so that's awesome 👍

ktbyers:
@georgesnow that made me happy...the everything else passing :-)

ktbyers:
@georgesnow Are you on Windows/Mac/Linux for the machine you are testing from?

Do you have over 100GB of local space available on that machine?

ktbyers:
Might need to adjust this number here (that 100 billion number upwards)

```
def test_verify_space_available_get(scp_fixture_get):
    ssh_conn, scp_transfer = scp_fixture_get
    assert scp_transfer.verify_space_available() == True
    # intentional make there not be enough space available
    scp_transfer.file_size = 100000000000
    assert scp_transfer.verify_space_available() == False
```

georgesnow:
yes 230GB it looks like:
252G  8.5G  230G   4% /

georgesnow:
yeah so I made:

300000000000

and now it passes:

py.test -s -v test_netmiko_scp.py::test_verify_space_available_get --test_device cienalab
================================= test session starts ==================================
platform linux -- Python 3.6.8, pytest-5.1.0, py-1.8.0, pluggy-0.12.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/gfn4/netmiko-cienatests/netmiko, inifile: setup.cfg
collected 1 item

test_netmiko_scp.py::test_verify_space_available_get PASSED

================================== 1 passed in 7.40s ===================================

ktbyers:
Okay, let me know if any of the other tests are still failing.

Then we'll see what @daanvdsanden testing on the Ciena 8700 shows on Friday.

georgesnow:
@ktbyers passed on all the tests for me. 

the only thing (and really just small thing) if at some point it would be nice is remove the hard coding in conftest.py for the user "ciena". And make it dynamic based on the user that's connecting. I had done that in my version of the code and I remember it being pretty simple to do. (I think) I bring it up cause I re-ran the test on separate machine (also as different user) and it failed. And it took me few minutes to remember. Give it a month and I won't remember to change it....

georgesnow:
@ktbyers  it is a relatively simple change change conftest.py path to:

"file_system": "/tmp/users/",

and change
https://github.com/ktbyers/netmiko/blob/d28c7a9b32437d89e665084d1664ae9e4fa9e7fa/netmiko/ciena/ciena_saos.py#L144:

to:

elif re.search(search_string, remote_out, flags=re.DOTALL)

ktbyers:
Using PR https://github.com/ktbyers/netmiko/pull/1384 for Ciena support so closing as duplicate.

ktbyers:
Ooops, close the wrong one: re-opening.

ktbyers:
@georgesnow Okay, I put a fix in for the test issue...so it should now use the SSH connection username for the unit tests.

georgesnow:
and it works! 👍

I will go back at some point and work on the other part of the Ciena gear for config/set commands when I get a chance.

georgesnow:
@ktbyers are you still planning merging this pull request this week?

ktbyers:
@georgesnow @daanvdsanden This all good from both of your's perspective?

georgesnow:
@ktbyers  everything worked for me. so it would just be if @daanvdsanden has any issues I think

daanvdsanden:
Just tested the code against three different platforms and all looks good to me.

PR #1409, opened 2019-10-07
add command verify to send_config_set

carlmontanari:
decided to leave the wlc mostly alone since im not sure if there was anything special going on there. cloud genix I rolled back to super the base connection as the only issue would be w/ entering/exiting config mode but that driver returns an empty string for those methods so that should work ok!


ktbyers:
I am going to close this as it is been sitting here for so long...I still think it would be a good idea to this, but should just start over on it.

PR #1424, opened 2019-10-17
Updates for Nokia SR OS (former: Alcatel-Lucent, Alcatel, Timetra)

wisotzky:
Support for Nokia SR OS devices using Netmiko. The updated module supports both classic CLI and model-driven CLI.

Integration notes:
- When running against MD-CLI the option `exit_config_mode=False` must be used. Use the commit operation to apply changes from candidate to running.
- If trying to exit the configuration mode (quit-config) without commit (or discard) CLI would raise a warning dialogue aka `Discard uncommitted changes? [y,n]`. Therefore we automatically discard changes before `quit-config` to avoid those exceptions.
- `enable-admin` has been removed. It did not really add value. In theory this allows a regular user to get admin privileges. However, if there is a need for admin rights, typically providers would assign the requires access-rights to the user account that is used by Netmiko.
- Classic CLI writes changes directly to the running configuration. No commit/discard exists.
- File operations (SCP) have been integrated. Corresponding CLI commands are "file ..." which is supported in classic CLI and from SR OS 19.10 onwards also in MD-CLI.
- There is no file operation to calculate/print MD5 hashes. Therefore it is not possible to validate copied files using checksums. Therefore option disable_md5 must be set to True. The `verify_file()` method is just comparing file sizes.

### DIFF
--- a/netmiko/alcatel/alcatel_sros_ssh.py
+++ b/netmiko/alcatel/alcatel_sros_ssh.py
@@ -50,7 +50,6 @@ class AlcatelSrosSSH(BaseConnection):
             self.base_prompt = match.group(1)
             return self.base_prompt
 
-
     def enable(self, *args, **kwargs):
         """Nokia SR OS does not support enable-mode"""
         pass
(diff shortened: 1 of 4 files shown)


wisotzky:
@ktbyers, when do you expect those merge this PR? Please reach out on me, if there is any intent to add test-automation for devices running Nokia SR OS. We can help out router images (VNF) and licenses.

netmiko/alcatel/alcatel_sros_ssh.py
```diff
@@ -1,70 +1,258 @@
-"""Alcatel-Lucent SROS support."""
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2019 - NOKIA Inc. All Rights Reserved.
+# Please read the associated COPYRIGHTS file for more details.
```
ktbyers: The "All Rights Reserved" clause would need dropped (sounds like it is generally inconsistent with the MIT license)

https://opensource.stackexchange.com/questions/2121/mit-license-and-all-rights-reserved


This probably need to change to:

# Copyright (c) 2014 - 2019  Kirk Byers 
# Copyright (c) 2014 - 2019  Twin Bridges Technology
# Copyright (c) 2019 Nokia Inc.
# MIT License - See License file at https://github.com/ktbyers/netmiko/blob/develop/LICENSE

I think that is more proper for the MIT License.

hellt:
I wonder if it would make more sense to change the file names to `nokia` instead of alcatel and phase out alcatel with the next couple releases?

Alcatel brand name wears off, so keeping it in the driver/file names is ok for now but will be less and less actual with the time

netmiko/alcatel/alcatel_sros_ssh.py
wisotzky: License info updated.

wisotzky:
Hi @ktbyers,
I would agree with @hellt to move naming from 'Alcatel' to 'Nokia'. However this would be only true for Nokia SROS as AOS is for OmniSwitches that belong to ALE (Alcatel-Lucent Enterprise) these days. I keep my contribution as is for the time being. Might need to keep 'alcatel_sros_ssh' for some time to avoid breaking compatibility for tools that have already integrated this.

ktbyers:
PR updated #https://github.com/ktbyers/netmiko/pull/1442

ktbyers:
Fixed nokia / alcatel naming in new PR. It is backwards compatible so old device_type will still work as well.

ktbyers:
Superseded by new PR

PR #1438, opened 2019-10-28
file_md5 modified to read file in chunks to compute md5 hash

javidaslan:
md5 hash computation fails for large files on RHEL OS. I tested computing md5 hash of file by reading the file in chunks and it worked. Therefore modified file_md5 function to read the file in chunks and generate md5.


carlmontanari:
Thanks @javidaslan; updated in #1439 to add some other improvements. Closing this one and will get the updated one pulled in.

PR #1448, opened 2019-11-06
Ensure file_size gets created regardless of match.

ronwellman:
I noticed when we accidentally passed in the full path to a file as the source_file of the file_transfer function.  This gets passed along to the remote_file_size function which utilizes regex to search for the filename in the output of a "dir" command.  If there is no match then the variable file_size never gets created.  However, since there was also no error, the remote_file_size function attempts to convert the file_size to an int before returning.  This causes a crash since file_size does not exist.  By forcing it to be the max size before the check for a match, at worst, the file will not be transferred and the program will not crash.

### DIFF
--- a/netmiko/scp_handler.py
+++ b/netmiko/scp_handler.py
@@ -219,7 +219,7 @@ class BaseFileTransfer(object):
         escape_file_name = re.escape(remote_file)
         pattern = r".*({}).*".format(escape_file_name)
         match = re.search(pattern, remote_out)
-        filesize = sys.maxsize
+        file_size = sys.maxsize
         if match:
             line = match.group(0)
             # Format will be 26  -rw-   6738  Jul 30 2016 19:49:50 -07:00  filename
--- a/netmiko/scp_handler.py
+++ b/netmiko/scp_handler.py
@@ -10,7 +10,7 @@ Currently only supports Cisco IOS and Cisco ASA.
 import re
 import os
 import hashlib
-
+import sys
 import scp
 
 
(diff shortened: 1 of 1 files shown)


ktbyers:
The program should crash though as you put an invalid source_file name?

ronwellman:
In our case, we put something to the affect of:

    source_file = "testfolder/testfile.txt"
    file_system = "system:"

This makes a valid path when joined so the "dir system:/testfolder/testfile.txt" inside of the file_transfer function returns a valid result.  However the regex attempts to match "testfolder/testfile.txt" which it cannot.  This results in no match and no error in the output (valid "dir") but attempts to cast file_size as an int before returning it.  Since file_size doesn't exist, that's where it crashes.   

So yes, we used the source_file and file_system variables incorrectly but the crash we got was because "file_size" didn't exist.  We only saw this error when attempting to transfer a file that was inside of a folder so we didn't get this same error when transferring files in the root of a file system.  After looking at your code, we discovered our error but thought the way the code was laid out, it allows for a condition where file_size does not exist and an error was not generated.  I hope my rambling makes sense.

ktbyers:
@ronwellman Yes, your statement above makes sense.

What networking platform are you on?

ronwellman:
Right now we're narrowly focused on Cisco devices.  In testing we've run our scripts across older Catalyst switches and ISR routers, semi-recent IOS-XE routers, and even IOL devices running inside of EVE-NG.  Essentially we're putting together a script that records various settings on the device, does a directory walk over the various file systems, and attempts to SCP each file off of the device for further analysis.  It's nothing all that exciting but we're using the project as a proof of concept that we can develop a few things in house rather than paying a vendor to develop it for us.  Assuming all goes well, we may look to do similar things with Brocade, Juniper, and Dell.

ktbyers:
This one I need to look at some more. I don't really think that is the right fix, but I haven't had much time to look at it.

ktbyers:
Fixed underlying issue in this PR:

https://github.com/ktbyers/netmiko/pull/2191

PR #1460, opened 2019-11-19
Brocade autodetect updated PR

ktbyers:


ktbyers:
Removed fastiron from the PR as the pattern was too generic.

ktbyers:
Those changes were already merged in previously...earlier this year so nothing left to do here.

Issue #1431, opened 2019-10-23
Disable complete on space for Juniper


carlmontanari:
Send `set cli complete-on-space off` as part of session preparation. Without this juniper tries to be "smart" and auto complete command syntax which may cause some issues.

carlmontanari:
closing; see #1464

PR #1464, opened 2019-11-22
disable junos complete on spacel

carlmontanari:
for #1431


ktbyers:
Reimplemented in below PR.

@carl I basically took your first pattern:

https://github.com/ktbyers/netmiko/pull/1483

I realized this was a better way to do it (after thinking about it a bit more).

PR #1473, opened 2019-12-03
Added sock param support in Netmiko

### LINKED ISSUE #818: Support for ssh parameter BindAddress
Would it be possible to add support for the 'BindAddress' client option for SSH?  We have management consoles with multiple interfaces and we need to specify which 'source ip address' for the ssh session with the 'BindAddress' statement.  Even though paramiko/netmiko supports user ssh config files, the code does not seem to handle the 'BindAddress' statements.  Ref; https://www.ssh.com/ssh/config/#sec-Listing-of-client-configuration-options

This is already documented on https://github.com/paramiko/paramiko/issues/206

ljm625:
Added a sock param support in Netmiko since Paramiko already supported.
This should also fix #818 which enables Netmiko to bind address.

Example Usage:
```python
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.bind(('10.75.58.221', 0)) # The address to bind
sock.connect(('128.9.0.5', 22)) #The server address
net_connect = Netmiko(
    "128.9.0.5",
    username="root",
    password="Test_123",
    device_type="huawei",
    sock=sock
)
```

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -834,6 +834,7 @@ class BaseConnection(object):
             "timeout": self.timeout,
             "auth_timeout": self.auth_timeout,
             "banner_timeout": self.banner_timeout,
+            "sock": self.sock,
         }
 
         # Check if using SSH 'config' file mainly for SSH proxy support
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -75,6 +75,7 @@ class BaseConnection(object):
         session_log_file_mode="write",
         allow_auto_change=False,
         encoding="ascii",
+        sock=None,
     ):
         """
         Initialize attributes for establishing connection to target device.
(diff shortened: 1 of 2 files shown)


ktbyers:
That won't work as that `self.sock` is never used nor passed to Paramiko (with the code in the PR).

What value is there in doing this? I.E. why is the sock parameter needed?

ljm625:
Ooops, my bad, thanks for noticing. I forgot to add this back after doing code reformatting.
The sock param enables you to bind to a specific address when initiating an SSH connection.
It's pretty useful when you have several interfaces and you need to specify the source IP to initiate the connections.
In my case, the device I am trying to configure when it's in DCN mode, it blocks all the SSH connections from its subnet IP, and I need to specify another source IP to successfully connected.
It also might be useful if some ACL rules are applied on devices.

ktbyers:
This looks fine. I just need to add some tests around it.

PR #1491, opened 2019-12-17
Bug Fix in SNMP autodetect function

kwrobert:
Fixed a small bug in the SNMP autodetect function. There was just a
whitespace issue which caused only the last entry for a particular
vendor in the SNMP_MAPPER_BASE dict to be executed, instead of trying
all of them in the order of priority. My guess is this passed all the
tests because the SNMP_MAPPER_BASE currently only has one entry for each
vendor.

### DIFF
--- a/netmiko/snmp_autodetect.py
+++ b/netmiko/snmp_autodetect.py
@@ -330,15 +330,15 @@ class SNMPDetect(object):
                 oid = v["oid"]
                 regex = v["expr"]
 
-            # Used cache data if we already queryied this OID
-            if self._response_cache.get(oid):
-                snmp_response = self._response_cache.get(oid)
-            else:
-                snmp_response = self._get_snmp(oid)
-                self._response_cache[oid] = snmp_response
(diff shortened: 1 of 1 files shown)


ktbyers:
Can you provide your example that was failing? That will help me better understand the issue.

kwrobert:
Unfortunately, I don't have a specific example that failed. I was just reading through the source code because the functionality was relevant to a project I'm working on, and noticed this issue. 

But, if you look at the source code, I think the issue is clear. Upstream, lines 329-331 just loop through the items in `entry` and do nothing but set the variables `oid` and `regex`. These two variables are unused by the code in lines 333-342 because those lines are not indented to be within the for loop. The code works as it is because Python doesn't restrict the scope of variables set inside for loops, so after you leave the loop the `oid` and `regex` variables are set to whatever their values were in the last iteration of the for loop. 

All I did was indent lines 333-342 so all of the items in the `entry` dict are actually used, and not just the last entry. 

Does that help? If not, I can try to make an example when I find the time.

PR #1494, opened 2019-12-20
Provide access to device output received before a timeout.

melewitz:
- Seeing the actual output is very helpful when debugging device interaction.
- A new "output" attribute has been added to NetmikoTimeoutException that
  contains all output received prior to the timeout.  The output text is not
  sanitized in any way so the caller has access to exact device output.
- For backward compatibility, the stringified exception remains as just the
  string error, i.e. it does not include any device output.
- An IOError that arguably should have been a timeout was converted to a
  NetmikoTimeoutError to allow for returning output.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -543,7 +543,7 @@ class BaseConnection(object):
                     self._write_session_log(new_data)
                 except socket.timeout:
                     raise NetmikoTimeoutException(
-                        "Timed-out reading channel, data not available."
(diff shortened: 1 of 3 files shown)


ktbyers:
@melewitz Any reason, you can't get this through the `session_log`?

melewitz:
I hadn’t seen it. Thanks for the pointer!

ktbyers:
I am going to close this as I think session_log is probably a better option.

PR #1499, opened 2020-01-02
New Driver: Ruijie

ktbyers:


ktbyers:
@chengzhengnan I removed some code that I thought should be unnecessary looking at the `ruijie` CLI documentation.

Any possibility you can re-run the test suites against this?

Is special_login_handler() required here? I.E. usually this isn't required for SSH (i.e. that would be a rare situation).


Thanks for all the work and contribution here.

czn92:
> @chengzhengnan I removed some code that I thought should be unnecessary looking at the ruijie CLI documentation.
> Any possibility you can re-run the test suites against this?
> Is special_login_handler() required here? I.E. usually this isn't required for SSH (i.e. that would be a rare situation).
> Thanks for all the work and contribution here.

@ktbyers Yes, those codes as well as special_login_handler() can all be removed. Thanks for this review.
I have rerun the test suites against this modification. It works.
[test_results_v2.txt](https://github.com/ktbyers/netmiko/files/4015040/test_results_v2.txt)

PR #1501, opened 2020-01-03
Simplifying huawei telnet regex slightly

ktbyers:


ktbyers:
@cuong-nguyenduy I simplified the regex slightly...basically removed the `re.escape("[Y/N]")` from the end of the regex.

Let me know if this causes any issues, but I am pretty sure dropping this won't affect anything.

Kirk

cuong-nguyenduy:
@ktbyers I originally wanted to have more specific match, but it is fine to have the `re.escape("[Y/N]")` removed.

Tested and worked well.

Cuong.

PR #1516, opened 2020-01-15
Minor Huawei Fix

ktbyers:
Just a minor change with respect to how the data is cleared.


ktbyers:
FYI, @cuong-nguyenduy I made this minor change as I thought it was slightly safer.

I doubt it will cause any issues with your earlier solution.

PR #1539, opened 2020-01-28
Connection timed-out vs connection refused exception

Alan-R:
This resolves the issue "Connection timed-out vs connection refused #795"

It replaces an earlier attempt at solving this problem from a few months ago. This code is Py3 only.


Alan-R:
I've been able to test this some more. It needs a bit more help. Paramiko makes it extra-complicated. Sigh...

Alan-R:
I'll see if I can get you a better version this week.

ktbyers:
I am going to close this as it is stale. Can probably discuss as a Netmiko4 item.

PR #1540, opened 2020-01-30
Update dell_dnos6.py

kroot-dev:
Fixed issue with function save_config for Dell os6


ktbyers:

PR #1542, opened 2020-01-30
Fix ansi escape code in output of paloalto_panos device type

wfleisher:
Added special_login_handler function to paloalto_panos.py
Sends command set cli scripting-mode on
This disables interactive cli access, and eliminates the ansi escape codes in the output.


ktbyers:
I think this is probably fixed here:

https://github.com/ktbyers/netmiko/pull/2223

PR #1545, opened 2020-01-31
Update vyos_ssh.py

kroot-dev:
Disable cmd_verify to be working


ktbyers:

PR #1546, opened 2020-01-31
Update vyos_ssh.py

kroot-dev:
Create save_config function


ktbyers:

PR #1557, opened 2020-02-04
Fixing output bug in send_config_set

ktbyers:


ciscorat:
Thanks for the prompt response.
I applied your fix in base_connection:
       # Send config commands
      output = ""
       if enter_config_mode:
            cfg_mode_args = (config_mode_command,) if config_mode_command else tuple()
            output += self.config_mode(*cfg_mode_args)

However, it still doesn't enter config mode with the same error:
` File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/netmiko/base_connection.py", line 1753, in send_config_set`
`    output += self.exit_config_mode()`

PR #1568, opened 2020-02-10
Update Huawei telnet login process

ktbyers:


ktbyers:
Adding @cuong-nguyenduy 

Sorry for the delay on this...

PR #1606, opened 2020-03-06
Mikrotik fix strip_prompt

ktbyers:


ktbyers:
I don't think this solves the underlying issue.

PR #1613, opened 2020-03-11
Refactored format strings using shorthand f"" to use "".format()

a25484189:
Hello,

this pull request changes all format strings which use the shorthand syntax

`f"{var}"`

which was introduced in python3.6 back to the regular format string syntax 

`"{}".format(var)`

to keep compatibility with older python versions like for example python3.5 which is the default in ubuntu 16.04 .

### DIFF
--- a/netmiko/arista/arista.py
+++ b/netmiko/arista/arista.py
@@ -24,13 +24,13 @@ class AristaBase(CiscoSSHConnection):
 
         Can also be (s2)
         """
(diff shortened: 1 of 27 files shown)


ktbyers:
We don't support Python3.5. You need to use Python 3.6 or greater or alternatively Netmiko 2.4.2.

We intentionally switched almost entirely over to f-strings.

You should also be able to install Python 3.6 and build a Python 3.6 virtual environment.

Issue #1623, opened 2020-03-18
Connection refused not captured


petermcd:
If a device chooses to refuse a connection Netmiko is not currently capturing the relevant exception from Paramiko. We identified this under a circumstance in which the wrong credentials were being tried in a short amount of time causing the device to lock out on the 4th attempt, This is generating output such as:

```
:paramiko.transport:Exception: 
 reading SSH protocol banner[Errno 104] Connection reset by peer<br>
:paramiko.transport:Traceback (most recent call last):<br>
:paramiko.transport:  File /PATH_TO_FILES/lib/python3.6/site-packages/paramiko/transport.py, line 2211, in _check_banner<br>
:paramiko.transport:    buf = self.packetizer.readline(timeout)<br>
:paramiko.transport:  File /PATH_TO_FILES/lib/python3.6/site-packages/paramiko/packet.py, line 380, in readline<br>
:paramiko.transport:    buf  = self._read_timeout(timeout)<br>
:paramiko.transport:  File /PATH_TO_FILES/lib/python3.6/site-packages/paramiko/packet.py, line 607, in _read_timeout<br>
:paramiko.transport:    x = self.__socket.recv(128)<br>
:paramiko.transport:ConnectionReset
: [Errno 104] Connection reset by peer<br>
:paramiko.transport:<br>
:paramiko.transport:During handling of the above exception, another exception occurred:<br>
:paramiko.transport:<br>
:paramiko.transport:Traceback (most recent call last):<br>
:paramiko.transport:  File /PATH_TO_FILES/lib/python3.6/site-packages/paramiko/transport.py, line 2039, in run<br>
:paramiko.transport:    self._check_banner()<br>
:paramiko.transport:  File /PATH_TO_FILES/lib/python3.6/site-packages/paramiko/transport.py, line 2216, in _check_banner<br>
:paramiko.transport:    
 reading SSH protocol banner   str(e)<br>
:paramiko.transport:paramiko.ssh_exception.SSHException: 
 reading SSH protocol banner[Errno 104] Connection reset by peer<br>
:paramiko.transport:REDACTED
: paramiko.ssh_exception.SSHException: 
 reading SSH protocol banner[Errno 104] Connection reset by peer 
```

Some of the above information has been redacted but gives you the idea of what is happening.

This issue has previously been reported on issue #795

tars01:
You have to enable exception handling:

`from netmiko.ssh_exception import NetMikoTimeoutException
[...]

PR #1624, opened 2020-03-18
Corrected failure to capture exception when connection refused

petermcd:
Resolves issue #1623 in which an exception thrown for a refused connection is not captured.

I have introduced a new Netmiko exception as the 2 current exceptions did not make sense for this scenario as although it can be caused by an Authentication issue, on the occasion it happens connection is refused prior to authentication attempting.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -22,6 +22,7 @@ from netmiko import log
 from netmiko.netmiko_globals import MAX_BUFFER, BACKSPACE_CHAR
 from netmiko.ssh_exception import (
     NetmikoTimeoutException,
+    NetmikoConnectionException,
     NetmikoAuthenticationException,
 )
 from netmiko.utilities import (
@@ -903,6 +904,16 @@ class BaseConnection(object):
                 )
                 msg += self.RETURN + str(auth_err)
                 raise NetmikoAuthenticationException(msg)
+            except paramiko.ssh_exception.SSHException as fail_err:
+                self.paramiko_cleanup()
+                msg = (
+                    "Connection Failure: The remote host rejected the connection. "
+                    "Potential lock-out {device_type} {ip}:{port}".format(
+                        device_type=self.device_type, ip=self.host, port=self.port
+                    )
+                )
+                msg += self.RETURN + str(fail_err)
+                raise NetmikoConnectionException(msg)
 
             if self.verbose:
                 print(f"SSH connection established to {self.host}:{self.port}")
--- a/netmiko/ssh_exception.py
+++ b/netmiko/ssh_exception.py
@@ -8,6 +8,12 @@ class NetmikoTimeoutException(SSHException):
     pass
 
 
+class NetmikoConnectionException(SSHException):
+    """SSH connection failed. Based on Paramiko ssh_exception on intial connection."""
+
+    pass
+
+
 class NetmikoAuthenticationException(AuthenticationException):
     """SSH authentication exception based on Paramiko AuthenticationException."""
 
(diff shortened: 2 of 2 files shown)


petermcd:
Updated so should pass the pylama linter now.

petermcd:
Can someone help with the formatting expected, On line 909 of base_connection.py. Originally the code passed black but failed pylama, I resolved the issue with pylama but now instead black fails.


I am not familiar with black but is having 2 linters that appear to have contradictory rules helpful?

ktbyers:
What type login failures do we have:

1. authentication fails
2. no device exists at IP address/port (either device doesn't exist, IP is wrong, ACL is blocking access in some way)
3. DNS failure
4. VTY ports all used up.

It would be good if we could differentiate between item 2 and item 3 (DNS failure versus other no device at IP address/port).

Also tests need added for these new exceptions. Probably in here:

https://github.com/ktbyers/netmiko/blob/develop/tests/test_netmiko_show.py

ktbyers:
Just run `black` again...that should fix it.

Black doesn't correct everything...there are somethings it gets stuck on.

ktbyers:
I think this exception is too generic also:

`paramiko.ssh_exception.SSHException`

I don't think that will work for this. For example, this same exception will happen if the SSH banner is too slow to arrive from the remote device.

At a minimum I think we would need to inspect the exception message back from Paramiko.

petermcd:
Unfortunately Netmiko can only rely upon the exception that is thrown from Paramiko and this is the exception that is thrown.

In my case the exception message mentioned the protocol banner but this of course would differ depending on why it failed. Personally we do not use DNS due to what our use case is.

At the very least I think this exception needs to be captured, maybe parse the error with known conditions and throw a generic exception for unknown conditions, As is the code is simply generating an uncaught exception error.

ktbyers:
You should be able to inspect the message from Paramiko to differentiate between the various cases.

Based on past experience I don't believe a wrong IP address, no TCP connection would behave this same way (i.e. I don't think that would generate Banner error). Your case is a bit different if you were locked out as you would be getting a TCP connection.

I am not sure this is net helpful in its current state as it just replaces a more specific Paramiko error message with a generic Netmiko error. I think we need more granularity on these various failure cases.

ktbyers:
Related PR here:

https://github.com/ktbyers/netmiko/pull/1539

ktbyers:
From a different post that someone made...that might be helpful (for the various cases that need covered):

```
try:
    net_connect = ConnectHandler(**ios_device)
except (AuthenticationException):
    print('Authentication failure: ' + ip_address_of_device)
    application_log.write(ip_address_of_device + ', Authentication failure' + '\r\n')
    continue
except (NetMikoTimeoutException):
    print('Timeout to device: ' + ip_address_of_device)
    application_log.write(ip_address_of_device + ', Timeout to device' + '\r\n')
    continue
except (EOFError):
    print('End of file while attempting device ' + ip_address_of_device)
    continue
except (SSHException):
    print('SSH Issue. Are you sure SSH is enabled? ' + ip_address_of_device)
    application_log.write(ip_address_of_device + ', Possible ssh issue' + '\r\n')
    continue
except Exception as unknown_error:
    print('Some other error: ' + str(unknown_error))
    application_log.write(ip_address_of_device + ', Unknown issue' + '\r\n')
    continue`
```

petermcd:
I do not believe this helps as this would have the same problem as you have raised with my approach. The above would generate the same exception for a banner timeout than a hard refusal by the device.


Also looking at the other pull request, unfortunately I do not believe that approach will work as he is trying to capture the condition using either ConnectionError or OSError however Paramiko first fails to capture a ConnectionResetError exception then throws a paramiko.ssh_exception.SSHException exception, neither of which extend from ConnectionError or OSError

ktbyers:
Okay, I am going to close this and open a more generic issue pertaining to Netmiko exception handling on failed connections.

PR #1629, opened 2020-03-20
Update tests for Dlink

ktbyers:


ktbyers:
dynax60 FYI, I updated this here.

Note, I modified the stripping of the "command" in the echo to not strip the last line (as per my question in the other PR). In other words I changed this from [2:-1] to [2:].

Let me know if that breaks something in the tests.

ktbyers:
There also is this new section in the config mode test...so the other fix should be unnecessary:

```python
def test_config_mode(net_connect, commands, expected_responses):
    """ 
    Test enter config mode
    """
    # Behavior for devices with no config mode is to return null string
    if net_connect.config_mode() != "": 
        assert net_connect.check_config_mode() is True
    else:
        assert True
```

PR #1642, opened 2020-03-25
Add support huawei OLT

Fikirsiz:
I added support Huawei OLT(8xxx) for netmiko. All tests are passing, but this device requires "enable" almost for all commands so I used command "display version" for the most values in commands.yaml. I did not find a way to run all the test commands with "enable".

[output_tests.txt](https://github.com/ktbyers/netmiko/files/4378660/output_tests.txt)


BSpendlove:
Hey, you should really use `return` and not `quit` in the `exit_config_mode` because if you configure something like an ACL, `quit` will return you back from ACL config mode to config mode, so it doesn't actually exit out of config mode properly. I had this issue with the SSH driver that I'm writing

Fikirsiz:
I agree it is more correct to use "return".

ktbyers:
@Fikirsiz Can you post the `py.test` output into this pull request?

ktbyers:
No SSH support only telnet?

Fikirsiz:
I added SSH and telnet support. Tested on Huawei MA5801.

[tests_results.txt](https://github.com/ktbyers/netmiko/files/4398103/tests_results.txt)

Fikirsiz:
I just tested the driver HuaweiSmartAXSSH on my OLT devices Huawei MA5801 and Huawei MA5608T - it works well for SSH and for Telnet. Test results below. 
[output_test.txt](https://github.com/ktbyers/netmiko/files/4398201/output_test.txt)

Can we just add an alias in file netmiko/huawei/huawei_smartax.py?
```
class HuaweiOLT(HuaweiSmartAXSSH):
    pass


class HuaweiOLTTelnet(HuaweiSmartAXSSH):
    pass
```

ktbyers:
I made an alias here:

https://github.com/ktbyers/netmiko/pull/1654

Note, this will only support SSH and your testing above would actually have been SSH (even when you tested telnet). In other words, this:

```
class HuaweiOLTTelnet(HuaweiSmartAXSSH):
    pass
```

Was still using SSH.

ktbyers:
If you want to add telnet support, then I think we should add it to the SmartAX driver and implement a PR that uses that as the base.

Let me know if you have any questions on how to do that.

Thanks for your contributions here...

ktbyers:
I made an error above regarding 'telnet' when I stated you were still using SSH...you were using 'telnet' so that was just an incorrect statement on my part. Telnet still needs re-added if you want it.

PR #1657, opened 2020-04-05
Unifi switch driver

ktbyers:


ktbyers:
@an-empty-string Made some updates here. Let me know if you see any issues. I will probably just go ahead and merge this and we can make modifications if necessary.

Issue #1663, opened 2020-04-14
SROS - Commit triggers NetmikoTimeoutException


jtishey:
I'm seeing an issue where I use the Nokia SROS driver to connect to a VM on SROS 16.0R7 in MD-CLI mode.   When I call the commit() method after send_config_set() - the configuration on the router commits, but netmiko raises a timeout: `netmiko.ssh_exception.NetmikoTimeoutException: Timed-out reading channel, data not available.`

Python 3.6.7
netmiko==3.1.0
TiMOS-C-16.0.R7-1 cpm/x86_64 Nokia 7750 SR
configuration-mode model-driven

@wisotzky - have you seen anything similar working on #1442  ?

Example:
```
>>> from netmiko import ConnectHandler
>>> d = {
...  'device_type': 'nokia_sros',
...  'ip': 'router1',
...  'username': username,
...  'password': password,
...  'port': 22,
...  'verbose': False
... }
>>> nc = ConnectHandler(**d)
>>> nc.send_config_set(["configure port 1/1/c10 description test"])
'edit-config exclusive\nINFO: CLI #2060: Entering exclusive configuration mode\nINFO: CLI #2061: Uncommitted changes are discarded on configuration mode exit\n\n(ex)[]\nA:user@router1# configure port 1/1/c10 description test\n\n*(ex)[]\nA:user@router1# '
>>> nc.commit()
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/paramiko/channel.py", line 699, in recv
    out = self.in_buffer.read(nbytes, self.timeout)
  File "/usr/lib/python3.6/site-packages/paramiko/buffered_pipe.py", line 164, in read
    raise PipeTimeout()
paramiko.buffered_pipe.PipeTimeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/netmiko/base_connection.py", line 550, in _read_channel_expect
    new_data = self.remote_conn.recv(MAX_BUFFER)
  File "/usr/lib/python3.6/site-packages/paramiko/channel.py", line 701, in recv
    raise socket.timeout()
socket.timeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/netmiko/nokia/nokia_sros_ssh.py", line 128, in commit
    output += self.read_until_pattern(r"@")
[...]

PR #1670, opened 2020-04-16
Update read after commit for fast commits #1663

jtishey:
PR for issue #1663  It appears that for MD-CLI devices, when committing configuration changes,  if the prompt returns quickly, the second read_until_pattern() call was not getting anything from the buffer.  The first one matches the commit command on the prompt, and can read past that line to the return prompt if the commit is completed quickly.  

This change updates to only one read operation after sending the commit command to search for "^(ex)[]" as opposed to "^*(ex)[]" - the * indicating uncommitted changes.

This does have the limit of requiring of being in config mode via "edit-config exclusive" in order to match the (ex) line, which is the default for the driver.  Reference:

```
def config_mode(self, config_command="edit-config exclusive", pattern=r"\(ex\)\["):
    """Enable config edit-mode for Nokia SR OS"""
```

Tested against TiMOS-C-16.0.R7-1 cpm/x86_64 Nokia 7750 SR in model-driven configuration mode.

### DIFF
--- a/netmiko/nokia/nokia_sros_ssh.py
+++ b/netmiko/nokia/nokia_sros_ssh.py
@@ -124,7 +124,7 @@ class NokiaSrosSSH(BaseConnection):
             log.info("Apply uncommitted changes!")
             cmd = "commit"
             self.write_channel(self.normalize_cmd(cmd))
-            output += self.read_until_pattern(r"[^\*]\(ex\)\[]"))
+            output += self.read_until_pattern(r"[^\*]\(ex\)\[]")
         return output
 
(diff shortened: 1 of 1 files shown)


jtishey:
Testing in a different enviornment with actual hardware is throwing errors with this..  Looking for different solutions.

ktbyers:
@jtishey For a different issue, I made changes on this line today:

```
output += self.read_until_pattern(pattern=re.escape(cmd))
```

See this PR here:

https://github.com/ktbyers/netmiko/pull/1671

PR #1681, opened 2020-04-20
Adds support for Cisco FTD

micahculpepper:
Adds basic SSH support


ktbyers:
@micahculpepper Can you run both the show tests and the config tests on this driver and paste the output here. We probably need both of these set of tests to be passing.

Details are here:

micahculpepper:
Will do. I wasn't intending to support configuration changes in this PR (and you wouldn't want to make them in the basic FTD CLI anyway), so I'll have to consider what this means for the tests.

ktbyers:
Okay, on my side, I am very reluctant to having have-working drivers (as it tends to create a lot of confusion, and support issues).

So we might need to discuss more what the FTD can reasonable do (or at least what you would want to do via the CLI on the FTD) and then see if there is a way to map that into Netmiko (and raise exceptions for things it can't do or that aren't supported by the Netmiko driver).

ktbyers:
I am going to close this as it is not a fully working driver.

Issue #1520, opened 2020-01-16
Avoid NET_TEXTFSM environment variable


yoonghm:
Hi,

  We could install ntc-template via 
```bash
pip install ntc_templates
```
We cloud find the path of installed package via

```python
mport inspect
inspect.getfile(ntc_templates)
```
With these, we do not need to create `NET_TEXTFSM` environment variable which is platform dependent and make user code non-portable.

I could create a pull request to handle it.

ktbyers:
I would be open to this if you want to do a PR...the pattern of preference should be:

1. Use NET_TEXTFSM
2. Try to find "pip" installed version
3. Fall back to looking in ~/ntc-templates/templates


This is for backwards compatibility and also I am not sure how frequently they are going to release new versions of ntc-templates to pypi. Also if you specify an environment variable it is pretty logical that it should have precedence.



Regards,

Kirk

yoonghm:
I have put up a pull request on https://github.com/ktbyers/netmiko/pull/1831

ktbyers:
This issue was solved here:

https://github.com/ktbyers/netmiko/pull/1961

Now support pip installed ntc-templates. Order is ENV_VAR > PIP installed ntc-templates > ~/ntc-templates/templates

PR #1696, opened 2020-04-25
Solve #1520 - Finds `pip` installed ntc-templates

lykinsbd:
Per conversation in #1520, the following is now the hierarchy for
 finding the `ntc-templates/templates` directory:

1. Find directory in `NET_TEXTFSM` Environment Variable
2. Check for pip installed `ntc-templates` location in this environment
3. ~/ntc-templates/templates

If a valid directory is not found, that is even if one is referenced in
 the environment variable or found in the home directory, a ValueError
 will be raised.

Added two dependencies into the `setup.py` file:

1. `ntc-templates` so that going forward it will be automatically
 installed with Netmiko
2. `importlib_resources` if Python is less than 3.7

### DIFF
--- a/netmiko/utilities.py
+++ b/netmiko/utilities.py
@@ -18,6 +18,12 @@ try:
 except ImportError:
     GENIE_INSTALLED = False
 
+try:
+    import importlib.resources as importlib_resources
+except ImportError:
+    # Handle Python < 3.7
+    import importlib_resources
+
 # Dictionary mapping 'show run' for vendors with different command
 SHOW_RUN_MAPPER = {
     "juniper": "show configuration",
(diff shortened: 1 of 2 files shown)


lykinsbd:
The failed checks all look like line-length and blackening problems, which I'm happy to address.  Mainly, I feel a bit slow because I can't find the reference anywhere in the repo for what line length it's checking for.

I assume that means the Black default of 88 lines, but I'm seeing things in these files that were already longer than 88 lines.  I was trying to make a minimal PR by only touching the lines I was making changes too, but if I need to blacken `setup.py` and `netmiko/utilities.py` at a specific line length, just let me know.

Thanks!

lykinsbd:
Blackened it with default settings, and it's passing Pylama and Black tests are failing, sorry about that. Fixing those now!

ktbyers:
@lykinsbd Looks like the unit tests broke after this change?

lykinsbd:
@ktbyers All fixed up and passing, sorry about the back/forth.

ktbyers:
No worries at all. Thanks for the contribution.

ktbyers:
Superceded by PR here: https://github.com/ktbyers/netmiko/pull/1961

PR #1710, opened 2020-04-30
Updated PR on Netmiko support for pip installed ntc-templates

ktbyers:


ktbyers:
@lykinsbd Made some updates here mainly related to the `importlib` behavior. New solution should be quite a bit simpler.

I removed direct ntc-templates dependency. I probably won't be willing to do that until this issue outlined here is fixed:

https://github.com/networktocode/ntc-templates/pull/672

Basically, I don't want that other terminal.py dependency in Netmiko (as that is unmaintained code) and caused me quite a bit of pain in the past.

Still some more review that I need to do.

lykinsbd:
I totally get waiting on making the direct dependency until `terminal.py` issue is sorted. 👍 

As for the `importlib.resources` fix, that works for me.  It is simpler and closer to what I originally did in in my first commit ( 13913fd ).

What I put afterwards, while more complicated, was focused on making `importlib.resources` importable by any module without needing the `try...except` block.  And would hold true anywhere else in the program after that as it modified `sys.modules`.

I've used it in the past at the top level of a module/library initiation so that other developers could simply do `import importlib.resources` without caring what python version they are running.

It may not make sense to either use it at all in Netmiko, but more specifically may not make sense to do in utilities.py.   Like I said, I totally get making it simpler, but wanted to give some background information on why I put the complicated version in there to start with.

ktbyers:
@lykinsbd Okay, that makes sense. 

I would say for now we just leave it as the simpler solution and then we can see across time if there are other needs for it (in Netmko).

ktbyers:
Superceded by PR here: https://github.com/ktbyers/netmiko/pull/1961

PR #1748, opened 2020-05-22
Centec fixes

ktbyers:
Minor fixes to the Centec driver.

### DIFF
--- a/docs/netmiko/centec/centec_os.html
+++ b/docs/netmiko/centec/centec_os.html
@@ -34,7 +34,6 @@ class CentecOSBase(CiscoBaseConnection):
         self._test_channel_read(pattern=r&#34;[&gt;#]&#34;)
         self.set_base_prompt()
         self.disable_paging()
-        self.set_terminal_width(command=&#34;terminal width 511&#34;)
         # Clear the read buffer
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()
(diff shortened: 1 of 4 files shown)


ktbyers:
@feng891005 I made a couple of minor modifications here.

Can you test if `terminal width 511` works on these Centec devices? I couldn't really determine this from looking online.

The other change was the `config_mode` call now looked unnecessary as it now was identical to the parent class (note, I made a change to the parent class yesterday to use the canonical command of `configure terminal` instead of an abbreviated command).

feng891005:
> @feng891005 I made a couple of minor modifications here.
> 
> Can you test if `terminal width 511` works on these Centec devices? I couldn't really determine this from looking online.
> 
> The other change was the `config_mode` call now looked unnecessary as it now was identical to the parent class (note, I made a change to the parent class yesterday to use the canonical command of `configure terminal` instead of an abbreviated command).

Hello ktbyers,
The Centec switch does not support the "terminal width" command，but the default terminal width is set to the maximum. So I deleted this command，

Sorry, I didn't pay attention to the changes to the "configure terminal" command. If you have already changed, then my changes are unnecessary

feng891005:
@ktbyers 
I have updated the centec_os.py file and deleted the config_mode function

ktbyers:
Okay, I will just go ahead and delete that `set_terminal_width` method and then we should be good to go.

Regards, Kirk

PR #1749, opened 2020-05-22
6wind driver update

ktbyers:


ktbyers:
@paunadeu I wanted to have an OS name--I debated between "turbo" and just generic "os".

Netmiko has convention for `device_type` of:

`vendor_osname_transport`

For example:

`cisco_ios_ssh`

SSH will be assumed if a transport is not provided.


So I converted this to `sixwind_os`

paunadeu:
Perfect! :-) I’m discussing with 6WIND R&D to implement disable pager function! :-)

Thanks and have a nice weekend!

PR #1750, opened 2020-05-23
update doc  and update centec_os.py

feng891005:
update doc  and update centec_os.py

### DIFF
--- a/netmiko/centec/centec_os.py
+++ b/netmiko/centec/centec_os.py
@@ -1,7 +1,6 @@
 """Centec OS Support"""
 from netmiko.cisco_base_connection import CiscoBaseConnection
 import time
-import re
 
 
 class CentecOSBase(CiscoBaseConnection):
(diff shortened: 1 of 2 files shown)


netmiko/centec/centec_os.py
```diff
@@ -1,31 +1,19 @@
...
 
 
 class CentecOSBase(CiscoBaseConnection):
     def session_preparation(self):
         """Prepare the session after the connection has been established."""
         self._test_channel_read(pattern=r"[>#]")
         self.set_base_prompt()
-        self.disable_paging()
-        self.set_terminal_width(command="terminal length 0")
+        self.enable()
```
ktbyers, line 11: Was this `enable()` call needed for some reason?

ktbyers:
@feng891005 I am going to close this PR as it should be a duplicate with the PR here:

https://github.com/ktbyers/netmiko/pull/1748

I still had the question on whether the `enable()` call was necessary, however.

netmiko/centec/centec_os.py
feng891005: hi ktbyers
Thank you for checking the code for me
centec_os.py is no longer a problem. Actually, you don't need to add enable (), but actually I want to add centec to the list of platforms. Could you please help to add this?

feng891005: Sorry,
has been added to the platform,
But https://ktbyers.github.io/netmiko/docs/netmiko/index.html has not been added here

PR #1753, opened 2020-05-25
add centec to index.html for api doc

feng891005:
@ktbyers 
hi ktbyers
I added centec to the index.html file so that it can be viewed in the api document


ktbyers:
@feng891005 Okay, I refreshed all of the docs here:

https://github.com/ktbyers/netmiko/pull/1756

PR #1774, opened 2020-06-04
cisco_ios.py: InLineTransfer.config_md5 restore correct check

icovada:
After #1408 `config_md5` just calls `file_md5`

`file_md5` computes hash of file contents from filename, `config_md5` should compute hash of a string


ktbyers:
Good catch...yeah, we totally messed that up earlier.

PR #1787, opened 2020-06-12
Add example of using concurrent connections

YAPNie:
Example of using Multithreading (concurrent, class based) connections.
Please, pay attention that I'm a newbie in programming (Python, Git, etc.) and just adapted the example from the Article: https://developer.ibm.com/articles/au-threadingpython/
I used Python v3.6.8.

### DIFF
--- a/examples/connect_multiple_devices/connect_multithread_class.py
+++ b/examples/connect_multiple_devices/connect_multithread_class.py
@@ -1,0 +1,131 @@
+#!/usr/bin/env python3
+"""
+This example is using Multithread Working Class.
+
+The example based on: https://developer.ibm.com/articles/au-threadingpython/
+
+There are some ways for concurrency:
+- Multi-threading:
+  - with a Worker class.    <-- this example about it.
+  - with a Worker funcion.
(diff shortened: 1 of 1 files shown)


BSpendlove:
Why use threading when you can just use concurrent.futures? I think it'd be much cleaner and more friendly to those looking to using threads (or processes with [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor)

ktbyers:
Agreed...any example should use concurrent.futures.

Here is a pretty simple example:

https://github.com/ktbyers/pyplus_course/blob/master/class10/collateral/ssh_threads_simple.py

YAPNie:
> https://github.com/ktbyers/pyplus_course/blob/master/class10/collateral/ssh_threads_simple.py
Hmm, I did not see it before. I tried to find the example inside the project's documentation and examples but did not find it.
My fail, sorry.

ktbyers:
@YAPNie No worries at all...it is not very easy to find.

PR #1792, opened 2020-06-15
ZTE Driver Pull Request

ktbyers:


ktbyers:
@TanY0Y0 I assume the OS here is ZXROS?

ktbyers:
@TanY0Y0 I just split broke out ZTE in your earlier pull-request that had a bunch of different drivers so that it was easier for me to process. I will likely do this for the other drivers as well.

Note, I changed the naming slightly to include the Network OS name in the device_type.

ktbyers:
This should be good to go...assuming that @TanY0Y0 is okay with the new naming.

TanY0Y0:
> @TanY0Y0 I assume the OS here is ZXROS?

sure,it is

PR #1794, opened 2020-06-15
New Driver for H3C

ktbyers:
@TanY0Y0


ktbyers:
@TanY0Y0 I assume this is using "Comware7" OS?

TanY0Y0:
> @TanY0Y0 I assume this is using "Comware7" OS?

ok

ktbyers:
@TanY0Y0 Is that a yes, that the underlying OS here is essentially ComWare7?

TanY0Y0:
> @TanY0Y0 Is that a yes, that the underlying OS here is essentially ComWare7?

yes,its OS is comware7

ktbyers:
@TanY0Y0 Are you still interested in having this H3C driver included in Netmiko?

TanY0Y0:
@ktbyers yes，I still interested in

ktbyers:

PR #1807, opened 2020-06-23
Strip ansi codes on Aruba devices and add BaseConnection constructor parameter

brasswood:
I was having problems testing netmiko on an Aruba 2930M switch, and stripping the ANSI codes fixed it. Not sure what effect this has on other Aruba devices such as controllers.

I also modified the BaseConnection constructor. I'm completely new to this project but it seems like it would make sense for `ansi_escape_codes` to be a constructor parameter that specific platforms can pass if they have to deal with ANSI codes.


brasswood:
Looking at other examples, it looks like I'm setting `self.ansi_escape_codes` the wrong way. Let me fix that.

PR #1810, opened 2020-06-24
More accurate ArubaOS-Switch driver

brasswood:
I rescind my previous pull request and present this one instead. ArubaOS is a unique operating system for Aruba controllers. ArubaOS-Switch is different, and it more closely resembles the HP Procurve operating system.

Fixes #1809 

```
================================================= test session starts ==================================================
platform linux -- Python 3.6.9, pytest-5.1.2, py-1.8.2, pluggy-0.13.1 -- /mnt/c/Users/me/workspace/netmiko/venv/bi
n/python3
cachedir: .pytest_cache
rootdir: /mnt/c/Users/me/workspace/netmiko, inifile: setup.cfg
plugins: pylama-7.7.1
collected 19 items

test_netmiko_show.py::test_disable_paging PASSED                                                                 [  5%]
test_netmiko_show.py::test_terminal_width PASSED                                                                 [ 10%]
test_netmiko_show.py::test_ssh_connect PASSED                                                                    [ 15%]
test_netmiko_show.py::test_ssh_connect_cm PASSED                                                                 [ 21%]
test_netmiko_show.py::test_send_command_timing PASSED                                                            [ 26%]
test_netmiko_show.py::test_send_command PASSED                                                                   [ 31%]
test_netmiko_show.py::test_cmd_verify_decorator PASSED                                                           [ 36%]
test_netmiko_show.py::test_send_command_global_cmd_verify PASSED                                                 [ 42%]
test_netmiko_show.py::test_send_command_juniper SKIPPED                                                          [ 47%]
test_netmiko_show.py::test_send_command_textfsm SKIPPED                                                          [ 52%]
test_netmiko_show.py::test_send_command_genie SKIPPED                                                            [ 57%]
test_netmiko_show.py::test_base_prompt PASSED                                                                    [ 63%]
test_netmiko_show.py::test_strip_prompt PASSED                                                                   [ 68%]
test_netmiko_show.py::test_strip_command PASSED                                                                  [ 73%]
test_netmiko_show.py::test_normalize_linefeeds PASSED                                                            [ 78%]
test_netmiko_show.py::test_clear_buffer PASSED                                                                   [ 84%]
test_netmiko_show.py::test_enable_mode PASSED                                                                    [ 89%]
test_netmiko_show.py::test_disconnect PASSED                                                                     [ 94%]
test_netmiko_show.py::test_disconnect_no_enable PASSED                                                           [100%]

=============================================== short test summary info ================================================
[...]

Issue #1699, opened 2020-04-27
prompt not stripped on some hp procurve switches


sliddjur:
For instance, the `show lldp info remote device` does not strip the prompt from the command. This causes some headache when using textfsm.

Here is log from `lldp = con_procurve.send_command("show lldp info remote-device")`

```
DEBUG:netmiko:In strip_ansi_escape_codes
DEBUG:netmiko:repr = ''
DEBUG:netmiko:Stripping ANSI escape codes
DEBUG:netmiko:new_output = 
DEBUG:netmiko:repr = ''
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: b'\n'
DEBUG:netmiko:In strip_ansi_escape_codes
DEBUG:netmiko:repr = '\x1b[1;0H\x1b[1M\x1b[24;1H\x1b[1L\x1b[24;31H\x1b[24;1H\x1b[2K\x1b[24;1H\x1b[?25h\x1b[24;1H\x1b[1;24r\x1b[24;1H\x1b[1;24r\x1b[24;1H\x1b[24;1H\x1b[2K\x1b[24;1H\x1b[?25h\x1b[24;1H\x1b[24;1H65432p-swi009-warehouse-1etg# \x1b[24;1H\x1b[24;31H\x1b[24;1H\x1b[?25h\x1b[24;31H'
DEBUG:netmiko:Stripping ANSI escape codes
DEBUG:netmiko:new_output = 65432p-swi009-warehouse-1etg# 
DEBUG:netmiko:repr = '65432p-swi009-warehouse-1etg# '
DEBUG:netmiko:read_channel: 65432p-swi009-warehouse-1etg# 
DEBUG:netmiko:In strip_ansi_escape_codes
DEBUG:netmiko:repr = ''
DEBUG:netmiko:Stripping ANSI escape codes
DEBUG:netmiko:new_output = 
DEBUG:netmiko:repr = ''
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:[find_prompt()]: prompt is 65432p-swi009-warehouse-1etg#
DEBUG:netmiko:In strip_ansi_escape_codes
DEBUG:netmiko:repr = ''
DEBUG:netmiko:Stripping ANSI escape codes
DEBUG:netmiko:new_output = 
DEBUG:netmiko:repr = ''
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: b'show lldp info remote-device\n'
[...]

PR #1819, opened 2020-06-29
substitute ESC[(\d)L with return characters for Aruba/ProCurve switches

brasswood:
Aruba and HP ProCurve switches appear to use this control sequence to print newlines.
This allows netmiko to handle it properly.

This is somewhat hacky because without a real VT100 emulator, we don't know where the
cursor is positioned and consequently where the newline should actually go. So this may or may
not cause issues on other switches using VT100 codes.

I changed the variable name code_form_feed to code_insert_line.
See https://www.xfree86.org/current/ctlseqs.html#VT100%20Mode

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1101,15 +1101,17 @@ class BaseConnection(object):
 
         # Initial attempt to get prompt
         prompt = self.read_channel()
+        log.debug(f"find_prompt initial attempt: prompt is {prompt}")
 
         # Check if the only thing you received was a newline
         count = 0
         prompt = prompt.strip()
         while count <= 12 and not prompt:
             prompt = self.read_channel().strip()
+            log.debug(f"find_prompt: count={count}, prompt={prompt}")
             if not prompt:
                 self.write_channel(self.RETURN)
-                # log.debug(f"find_prompt sleep time: {sleep_time}")
+                log.debug(f"find_prompt sleep time: {sleep_time}")
                 time.sleep(sleep_time)
                 if sleep_time <= 3:
                     # Double the sleep_time when it is small
@@ -1120,7 +1122,9 @@ class BaseConnection(object):
 
         # If multiple lines in the output take the last line
         prompt = self.normalize_linefeeds(prompt)
+        log.debug(f"find_prompt: normalized prompt: {prompt}")
         prompt = prompt.split(self.RESPONSE_RETURN)[-1]
+        log.debug(f"find_prompt: last line, split around {self.RESPONSE_RETURN}: {prompt}")
         prompt = prompt.strip()
         if not prompt:
             raise ValueError(f"Unable to find prompt: {prompt}")
@@ -1814,7 +1818,7 @@ class BaseConnection(object):
         code_erase_line = chr(27) + r"\[2K"
         code_erase_start_line = chr(27) + r"\[K"
         code_enable_scroll = chr(27) + r"\[\d+;\d+r"
-        code_form_feed = chr(27) + r"\[1L"
+        code_insert_line = chr(27) + r"\[(\d)L"
         code_carriage_return = chr(27) + r"\[1M"
         code_disable_line_wrapping = chr(27) + r"\[\?7l"
         code_reset_mode_screen_options = chr(27) + r"\[\?\d+l"
@@ -1837,7 +1841,6 @@ class BaseConnection(object):
             code_erase_line,
             code_enable_scroll,
             code_erase_start_line,
-            code_form_feed,
             code_carriage_return,
             code_disable_line_wrapping,
             code_erase_line_end,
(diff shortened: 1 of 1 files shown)


brasswood:
#1699

ktbyers:
@brasswood Can you expand on the reasoning in your PR? In almost all cases ProCurve's have a normal <enter>, why do they only rarely do this?

brasswood:
@ktbyers When just pressing enter at the command prompt, our Aruba 2930s and 3810s consistently do not print newline characters or `ESC E`. Here is the unfiltered result from the switch from just pressing enter three times at the prompt:
```
\x1b[1;0H\x1b[1M\x1b[232;1H\x1b[1L\x1b[1000;16H\x1b[232;1H\x1b[2K\x1b[1000;1H\x1b[?25h\x1b[1000;1H\x1b[1;232r\x1b[1000;1H\x1b[1;232r\x1b[1000;1H\x1b[232;1H\x1b[2K\x1b[1000;1H\x1b[?25h\x1b[1000;1H\x1b[232;1HCOMP-ES-2E-01# \x1b[1000;1H\x1b[232;16H\x1b[1000;1H\x1b[?25h\x1b[1000;16H\x1b[1;0H\x1b[1M\x1b[232;1H\x1b[1L\x1b[1000;16H\x1b[232;1H\x1b[2K\x1b[1000;1H\x1b[?25h\x1b[1000;1H\x1b[1;232r\x1b[1000;1H\x1b[1;232r\x1b[1000;1H\x1b[232;1H\x1b[2K\x1b[1000;1H\x1b[?25h\x1b[1000;1H\x1b[232;1HCOMP-ES-2E-01# \x1b[1000;1H\x1b[232;16H\x1b[1000;1H\x1b[?25h\x1b[1000;16H\x1b[1;0H\x1b[1M\x1b[232;1H\x1b[1L\x1b[1000;16H\x1b[232;1H\x1b[2K\x1b[1000;1H\x1b[?25h\x1b[1000;1H\x1b[1;232r\x1b[1000;1H\x1b[1;232r\x1b[1000;1H\x1b[232;1H\x1b[2K\x1b[1000;1H\x1b[?25h\x1b[1000;1H\x1b[232;1HCOMP-ES-2E-01# \x1b[1000;1H\x1b[232;16H\x1b[1000;1H\x1b[?25h\x1b[1000;16H
```
No `\n`, `\r`, or `ESC E`. Instead, it is using the control sequences to clear the line at the top of the screen, then insert a new line at the bottom of the screen (using `ESC[1M` and `ESC[1L` respectively).

While I don't have a ProCurve switch, I assume that they all behave this way too. Firstly, the fact that the `HPProcurveBase` class works almost perfectly on ArubaOS Switch (the code in that base class precisely expresses the behavior of our Aruba switches), combined with the fact that HPE has acquired Aruba Networks makes me believe that ArubaOS Switch is based on the ProCurve operating system (see #1810). 

Secondly, the @sliddjur at #1699 had this error:
>```
>OSError: Search pattern never detected in send_command_expect: 65432p\-swi004\-new\-office\-2etg\#\ 65432p\-swi004\-new\-office\-2etg\#
>```

His switch's raw output:
>```
>DEBUG:netmiko:repr = '\x1b[1;0H\x1b[1M\x1b[24;1H\x1b[1L\x1b[24;32H\x1b[24;1H\x1b[2K\x1b[24;1H\x1b[?25h\x1b[24;1H\x1b[1;24r\x1b[24;1H\x1b[1;24r\x1b[24;1H\x1b[24;1H\x1b[2K\x1b[24;1H\x1b[?25h\x1b[24;1H\x1b[24;1H65432p-swi004-new-office-2etg# \x1b[24;1H\x1b[24;32H\x1b[24;1H\x1b[?25h\x1b[24;32H\x1b[1;0H\x1b[1M\x1b[24;1H\x1b[1L\x1b[24;32H\x1b[24;1H\x1b[2K\x1b[24;1H\x1b[?25h\x1b[24;1H\x1b[1;24r\x1b[24;1H\x1b[1;24r\x1b[24;1H\x1b[24;1H\x1b[2K\x1b[24;1H\x1b[?25h\x1b[24;1H\x1b[24;1H65432p-swi004-new-office-2etg# \x1b[24;1H\x1b[24;32H\x1b[24;1H\x1b[?25h\x1b[24;32H'
>```
Notice that this is also using `ESC[1M` and `ESC[1L` to create the newline on the terminal.

This is the exact issue that I was having, and telling netmiko to interpret `ESC[<n>L` as n newlines fixed it for me.

ktbyers:
Yeah, I have a ProCurve in my lab and there has been a ProCurve driver in Netmiko for around 6 years and it doesn't behave this way in general (or else the entire ProCurve driver would break).

Also even in the referenced issue #1699 there are other newlines in the output and Netmiko would break much earlier so that isn't a sufficient explanation for 1699. By that I mean, there would be the question of why does it normally output <enters> and then in this one rare circumstance not?

Anyways let me look at it some more...

ktbyers:
Okay, looking at it some more...I understand it a bit better. Let me research it some more.

brasswood:
>Netmiko would break much earlier
As long as `find_prompt()` loops only once, the switch only prints one prompt, and there are no issues. This is usually the case.

Sometimes we don't receive the data until the second or third loop of `find_prompt()`, and the switch gives us 2 or 3 prompts. This is when problems arise, as netmiko sees no newlines to split the string around, so it thinks that the three prompts are one.

@ktbyers does your ProCurve behave similarly just when pressing enter at the command prompt? I am aware that in some contexts these switches also print newlines.

ktbyers:
```
As long as find_prompt() loops only once, the switch only prints one prompt, and there are no issues. This is usually the case.
```

Assuming that is true, then you could also fix the issue by using `global_delay_factor`. Of course, you would potentially have an undesirable variability here when it sometimes fails.

It is a bit weird they would do this behavior in the hitting enter at a blank prompt, but not in a bunch of other contexts. But the ProCurves are weird devices so it wouldn't surprise me.

I generally run my test suite against my lab ProCurve on a reasonably regular behavior and haven't ever ran into the issue (as far as I am aware). But yeah, I could test it some more and see if I can reproduce this (I am not sure I will or whether I will just implement a variation of your fix).

I will probably implement some variation of your fix above as it doesn't really hurt things and is not all that different from the below which we are already doing here.

```
return re.sub(code_next_line, self.RETURN, output)
```


Thanks for all your research/work here...I was a bit skeptical this was a real thing, but now I understand it better and it makes sense.

ktbyers:
Replacing with updated PR

PR #1828, opened 2020-07-05
Add progress bar for Secure Copy

ktbyers:


ktbyers:
@kjoyce77 I expanded upon your work to add a `progress_bar` function so that there is a built-in progress bar that can be used with Netmiko.

It looks as follows:

```
Transferring file to ('184.105.247.70', 22): test1.txt

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>       | (92.70%)
```

It supports both `progress` and `progress4`


Thanks for this contribution...good job :-)

Takezo13:
Could you document this new function and show some examples on how to use it because I am a bit confused about the way to use it to be honest.

kjoyce77:
@Takezo13 I'll set a follow-up for myself to document.  In the meantime the basic use is similar to what is documented under the scp package except you are using netmiko's SCPConn instead of SCPClient and the netmiko ConnectHandler in place of ssh.get_transport.  Also, you'd be using scp_put_file instead of put.

https://pypi.org/project/scp/

Here is some sample code that might clarify things for you.  This uses the built-in progress function @ktbyers added to my pull request.
```python
import sys
import datetime
import argparse
import ipaddress
import netmiko

arg_parser = argparse.ArgumentParser()
for argument in ("dst_name", "dst_host", "src_file", "dst_file", "user", "password"):
    arg_parser.add_argument(argument)
args = arg_parser.parse_args()

if args.dst_name == args.dst_host:
    print(f"Transfering {args.src_file} to {args.dst_name}:{args.dst_file}...")
else:
    print(
        f"Transfering {args.src_file} to {args.dst_name}({args.dst_host}):{args.dst_file}"
    )
sys.stdout.flush()

device = {
    "device_type": "cisco_ios",
    "username": args.user,
    "password": args.password,
}

try:
    device["ip"] = str(ipaddress.ip_address(args.dst_host))
except ValueError:
    device["host"] = args.dst_host

with netmiko.ConnectHandler(**device) as net_connect:
    scp_conn = netmiko.SCPConn(net_connect, progress=netmiko.progress_bar)
    scp_conn.scp_put_file(args.src_file, args.dst_file)
    scp_conn.close()
```

ktbyers:
Here was my testing code on it:

```python
from getpass import getpass
from netmiko import ConnectHandler, file_transfer, progress_bar


cisco = {
    "device_type": "cisco_ios",
    "host": "cisco1.lasthop.io",
    "username": "pyclass",
    "password": getpass(),
}

# A secure copy server must be enable on the device ('ip scp server enable')
source_file = "test1.txt"
dest_file = "test1.txt"
direction = "put"
file_system = "flash:"

ssh_conn = ConnectHandler(**cisco)
transfer_dict = file_transfer(
    ssh_conn,
    source_file=source_file,
    dest_file=dest_file,
    file_system=file_system,
    direction=direction,
    # Force an overwrite of the file if it already exists
    overwrite_file=True,
    progress4=progress_bar,
)

print(transfer_dict)
```

ktbyers:
Here was my testing code on. It is a bit hard to capture in a single screenshot since the '>' indicator and the percentage keep increasing when you execute it:

```
Transferring file to ('184.105.247.70', 22): test1.txt

>>>>>>>>>>>>>>>                                   | (31.67%)
```

Takezo13:
Many thanks for the add'l info! I actually found out that it works perfectly fine with a "put" but with a "get", I have the following callback error:

...\lib\site-packages\netmiko\scp_functions.py", line 15, in progress_bar
    filename = filename.decode()
AttributeError: 'str' object has no attribute 'decode'

assadniang:
Also, could you specify a directory to transfer instead of a file?

Takezo13:
May I know if the progress bar is only working for put file transfer or if it's supposed to work also for a get? If so, maybe it's my script which is having an issue (but it works fine without the progress_bar so I am a bit puzzled). Many thanks in advance for your help.

ktbyers:
@Takezo13 I need to test this...the expectation was that it should work for a get. So if it doesn't, then it is a bug.

ktbyers:
@assadniang Nope, no directories just single files (at least by design only single files).

PR #1752, opened 2020-05-23
Textfsm search path to search for pip installed ntc-templates

ktbyers:

PR #1831, opened 2020-07-08
Use path from installed ntc_templates library

yoonghm:
Netmiko could search installed ntc_templates from Python distribution before searching directory pointed by NET_TEXTFSM environmental variable and then user's home directoru.


ktbyers:
This is a duplicate with:

https://github.com/ktbyers/netmiko/pull/1752

So I am going to close this PR.

There would be some other things needed in a PR:
1. Order would need to be:
* environment variable
* pip installed ntc-templates
* ~/ntc-templates/templates/index
2. Unit tests would be needed to verify this was all working properly.

PR #1838, opened 2020-07-09
NewDriver: Netgear ProSafe

cppmonkey:
A new driver for Netgear ProSafe devices.
Currently it has only been tested with M5300 series switches.

### DIFF
--- a/PLATFORMS.md
+++ b/PLATFORMS.md
@@ -39,6 +39,7 @@
 - MikroTik RouterOS
 - MikroTik SwitchOS
 - NetApp cDOT
+- Netgear ProSafe
 - Nokia/Alcatel SR OS
 - OneAccess
 - Palo Alto PAN-OS
--- a/_config.yml
+++ b/_config.yml
@@ -1,1 +1,1 @@
-theme: jekyll-theme-dinky
+theme: jekyll-theme-cayman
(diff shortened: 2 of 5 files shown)


ktbyers:
Is this PR ready or still a work-in-process?  @cppmonkey

cppmonkey:
Hi @ktbyers, it could do with some more testing.
What validation should I be including?

ktbyers:
You should make sure the "show" and "config" tests detailed here all pass and post the results as part of the pull-request.

Thanks, Kirk

cppmonkey:
Done a spot of testing. Did crash the switch, but I think that was because I was spamming _show loginsessions_ from the console
image

Tested the saving ability and it does work. But it has to be called explicitly.
_write memory confirm_ is mentioned within the CLI for [11.0.0 (page 235)](https://www.downloads.netgear.com/files/GDC/M5300/M5300-M6100-M7100_CLI_v11_20Apr2015.pdf) but not for [10.0.0 (page 572)](https://www.netgear.jp/support/manual/M5300/M5300_CLI_27Nov2012.pdf)

So it should work for M5300, M6100, and M7100 Series ProSAFE Managed Switches running 11.0.0.x

I've tested it against an M5300-28G (GSM7228S)


@ktbyers feel free to merge this. Many thanks for the awesome project and your training courses

Issue #1863, opened 2020-07-30
Nokia SR OS/Alcatel AOS - problem with detecting prompt while using save_config after device reboot for software upgrade


akei9:
I found a problem with PROMPT detecting mechanism in Nokia SR OS/Alcatel-Lucent AOS when we are restarting device using **_admin reboot upgrade_** command. 

When we are restarting network device using above command and log on to this device after reboot the prompt is starting from asterisk sign (***X:HOSTNAME#**).  

Problems starts when we want to use **save_config()** method after reboot. The command hangs and then returns the timeout exception because it cannot find the prompt (_because is looking for the one without a asterisk sign_ - **X:HOSTNAME#**).

The solution is to add a condition where we check if the asterisk sign is in the search_pattern variable, and if so, replace it with an empty string:

```
if '*' in search_pattern:
    search_pattern = search_pattern.replace('*', '')
```

The other solution is to modify **save_config()** method in **nokia_sros_ssh.py[113]** (now it looks like below):
```
def save_config(self, *args, **kwargs):
    """Persist configuration to cflash for Nokia SR OS"""
    return self.send_command(command_string="/admin save", expect_string=r"#")
```

and change to that (_with added **kwargs to **send_command** method parameter_):
```
def save_config(self, *args, **kwargs):
    """Persist configuration to cflash for Nokia SR OS"""
    return self.send_command(command_string="/admin save", **kwargs)
```

that we can set our own expected string variable (expect_string).

And the last - best (additional) solution is to write a separate function for the admin reboot upgrade operation, including the change of the prompt detection.

akei9:
> What is the process in the CLI to restore the prompt back to normal in this case (after the device reboot)?
> 
> Kirk

Calling admin save command is changing prompt to normal form. I’m preparing proposition of changes in code and I will pull request this changes soon.

akei9:
I was thinking about this a little bit and it can be solved in yet another way:

```
def save_config(self, expect_string=r"#", *args, **kwargs):
    """Persist configuration to cflash for Nokia SR OS"""
    return self.send_command(command_string="/admin save", expect_string=expect_string)
```

ktbyers:
Does it work post-reboot to just do:

```python
net_connect.save_config()
net_connect.set_base_prompt()
```

ktbyers:
Okay, I am pretty sure the above will work from my testing so I am going to close this and close your PR (just let me know if it doesn't work, however).


Regards, Kirk

PR #1866, opened 2020-07-31
Add admin reboot upgrade and save_config fix in Nokia SR-OS

akei9:
PR includes two changes:
- new **reboot_upgrade** method that allows you to restart the device for software upgrade
- fix in **save_config** method


ktbyers:
Wouldn't this fail?

I.E. the device reboots so you won't have an SSH connection and would need to re-establish a new connection?

ktbyers:
See comments on this PR:

https://github.com/ktbyers/netmiko/issues/1863

PR #1870, opened 2020-08-04
Fix broken "enable" on Nokia SR-OS driver

ktbyers:


ktbyers:
Tested both show and config against both classical CLI and MD-CLI.

ktbyers:
Roman is going to verify that failing 'enable-admin/enable' call does not cause system lockout (cc @hellt)

PR #1871, opened 2020-08-05
Minor change to Netgear Prosafe

ktbyers:


ktbyers:
@cppmonkey Let me know if this change breaks anything. I thought it should work from looking at the code and the Netgear behavior.

cppmonkey:
@ktbyers No problem. Not had a chance yet. Returned the M5300 back to the office yesterday.
Still not very good at dealing with virtualenv. Not the Python Guru you are but I'm learning.
I have all of your [PyNET](https://pynet.twb-tech.com/class-pyauto.html) videos archived for future reference =)

Hoping that other members at work will take an interest and partake in your great course!

PR #1877, opened 2020-08-07
fix yotc driver

TanY0Y0:
fix yotc driver

1.```send_command_more```as default send_command func for yotc device, so its need send command to get the first response, then get ```--more--``` responses
2.fix save_config:
```save_config``` must be in enable mode and not in config mode, otherwise it cant be inputed, and there is a reason why I cant use the ```super.save_config``` that it need confirm_response the "y" without "\n"
,that means once you type "y", it will respond immediately
3.add cleanup():
reset the "command max-lines" to default

are these code styles ok? after this check I will re-run the tests

### DIFF
--- a/netmiko/yotc/yotc.py
+++ b/netmiko/yotc/yotc.py
@@ -42,11 +42,7 @@ class YotcBase(CiscoBaseConnection):
 
         while prompt not in new_data:
             new_data = self.send_command_timing(
-                "m",
-                use_textfsm=False,
-                use_genie=False,
-                strip_prompt=False,
-                **kwargs,
+                "m", use_textfsm=False, use_genie=False, strip_prompt=False, **kwargs,
             )
             output += new_data
         output = re.sub(more_str_re, "", output)
@@ -70,9 +66,7 @@ class YotcBase(CiscoBaseConnection):
 
         if use_genie:
             structured_output = get_structured_data_genie(
-                output,
-                platform=self.device_type,
-                command=command_string.strip(),
+                output, platform=self.device_type, command=command_string.strip(),
             )
             # If we have structured data; return it.
             if not isinstance(structured_output, str):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by TanY0Y0** [comment]
@ktbyers 
image
I black the file, but I dont know why it cant pass black check

File: netmiko/yotc/yotc.py [hunk 1]
```diff
@@ -78,13 +81,16 @@ def check_config_mode(self, check_string=")#", pattern="#"):
     def save_config(self, cmd="wr", confirm=True, confirm_response="y"):
         """Saves Config."""
         self.exit_config_mode()
-        return super().save_config(
-            cmd=cmd, confirm=confirm, confirm_response=confirm_response
```
ktbyers, line 82: Is there some reason the `super()` call doesn't work--it should (given the behavior you are doing with `send_command_timing`?

I have a very strong preference for calling the parent class via super() as otherwise the Netmiko code becomes very hard to maintain (i.e. I have to search and edit a large number of files whenever there is a behavior change).

netmiko/yotc/yotc.py
```diff
@@ -34,12 +34,15 @@ def _send_command_more(
...
         """
         prompt = self.find_prompt()
-        more_str_re = r".*--More--.*"
-        new_data = ""
-        output = ""
+        more_str_re = r"\s--More--\s*"
+        new_data = self.send_command_timing(
+            command_string, use_textfsm=False, use_genie=False, **kwargs,
+        )
+        output = new_data
```
ktbyers, line 41: Why duplicate this code? Why not just have it once inside the loop?

netmiko/yotc/yotc.py
```diff
@@ -78,13 +81,16 @@ def check_config_mode(self, check_string=")#", pattern="#"):
...
+        output += self.send_command_timing(
+            confirm_response, strip_prompt=False, strip_command=False, normalize=False,
         )
+        return output
 
     def cleanup(self):
-        """Don't use 'exit' to disconnect."""
-        pass
+        """Don't use 'exit' to disconnect, and reset command max-lines to default"""
+        self.send_config_set("no command max-lines")
```
ktbyers, line 93: What is the actual command to gracefully logout from the SSH session?

TanY0Y0: sometimes when device is busy, exit is no use, and busy is often,so just disconnect, "no command max-lines" is to reset the "command max-lines 1000" that typed in ```session_preparation()``` , its permanent

netmiko/yotc/yotc.py
TanY0Y0: I use ```send_command_more```  as the default send_command func in yotc, as it have no disable_paging func, and just can expand to 1000 lines，so need use this to show all things in one time，so need enter the first command link ```show run```， if just loop the `` ``,its no place to use this func

netmiko/yotc/yotc.py
TanY0Y0: if the super.save_config() can add ```normalize=False``` option it can be solved, because when type "y\n",  the device think its not "y", so save_config will canelled

ktbyers: Okay, that makes sense...now I understand why it wasn't working in the super-call.

ktbyers:
@TanY0Y0 Are you still interested in having this YOTC driver included in Netmiko?

TanY0Y0:
@ktbyers   I still want YOTC driver in

ktbyers:
Okay, I will need to circle back to it...I really hate that lack of a way to turn off output paging on the platform/device.

TanY0Y0:
@ktbyers   me too,if there has a way to turn off paging, i will absolutely use it 😂

ktbyers:

Issue #1858, opened 2020-07-29
Cisco XR commit comment should not be quoted


tdorssers:
The cisco_xr driver is currently quoting a commit comment, which is redundant. This is a day 0 issue since #89. Please have a look at commitId 1000003264, which is configured by netmiko:

```
RP/0/RSP0/CPU0:XR1(config-if)#commit comment ?
  LINE  Comment for this commit (Max 60 characters)
RP/0/RSP0/CPU0:XR1(config-if)#commit comment no quotes required
Wed Jul 29 17:52:59.910 CEST
RP/0/RSP0/CPU0:XR1(config-if)#
RP/0/RSP0/CPU0:XR1#show configuration commit list detail 
Wed Jul 29 17:53:09.703 CEST

   1) CommitId: 1000003265                 Label: NONE
      UserId:   admin                      Line:  vty0:node0_RSP0_CPU0
      Client:   CLI                        Time:  Wed Jul 29 17:52:59 2020
      Comment:   no quotes required 

   2) CommitId: 1000003264                 Label: NONE
      UserId:   admin                      Line:  vty1:node0_RSP0_CPU0
      Client:   CLI                        Time:  Wed Jul 29 10:18:12 2020
      Comment:   "bulk provision" 

```

For your reference, I manually configured the router to demonstrate that quotes are not needed.
I propose to remove the quoting and add a check to limit the comment length to 60 chars.
If the length exceeds 60, the commit will fail:

```
RP/0/RSP0/CPU0:XR1(config)#commit comment 0123456789012345678901234567890123456789012345678901234567890123456789
                                                                                                      ^
% Invalid input detected at '^' marker.

```

Thanks.

OzNetNerd:
@tdorssers, I've raised a PR to fix the quoting - https://github.com/ktbyers/netmiko/pull/1878. If it's integrated into the code base, you'll be able to disable quotes by passing `quote_comment=False` to `commit()`.

As per the PR comment, I didn't implement a comment length restriction. The reason being that it may create issues in the future.

ktbyers:
Fixed here:

https://github.com/ktbyers/netmiko/pull/1880

PR #1878, opened 2020-08-10
Made Cisco XR commit message quotes optional

OzNetNerd:
Fix for https://github.com/ktbyers/netmiko/issues/1858.

I'm reluctant to remove the quotes entirely in case they're required for some versions of the OS. Similarly, I'm reluctant to limit the commit message length in case it is increased in future versions of the OS.

To ensure the commit message is acceptable, I suggest devs check the command output for the `% Invalid input detected at '^' marker.` string.

### DIFF
--- a/netmiko/cisco/cisco_xr.py
+++ b/netmiko/cisco/cisco_xr.py
@@ -25,7 +25,13 @@ class CiscoXrBase(CiscoBaseConnection):
         )
 
     def commit(
-        self, confirm=False, confirm_delay=None, comment="", label="", delay_factor=1
+        self,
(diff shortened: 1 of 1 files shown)


ktbyers:
@OzNetNerd I made a slightly different change here:

https://github.com/ktbyers/netmiko/pull/1880

Basically, I didn't really want to add a new argument to the commit method. The fix in 1880 will just pass in the comment directly (so if you need double quotes, then you will need to double quote it).

I suspect this was just an implementation error on the original implementation (by me).

Thanks for the contribution.

PR #1899, opened 2020-08-26
Fix autodetect for 3850s

rlaneyjr:
Autodetect for 3850s would use cisco_ios when this should be cisco_xe.  Fixed for both SSH and SNMP.  Also added Panos running config mapper command "show config running".

### DIFF
--- a/netmiko/paloalto/paloalto_panos.py
+++ b/netmiko/paloalto/paloalto_panos.py
@@ -96,7 +96,7 @@ class PaloAltoPanosBase(BaseConnection):
             if device_and_network:
                 command_string += " device-and-network"
             if policy_and_objects:
-                command_string += " device-and-network"
+                command_string += " policy-and-objects"
             if no_vsys:
                 command_string += " no-vsys"
             command_string += " excluded"
--- a/netmiko/snmp_autodetect.py
+++ b/netmiko/snmp_autodetect.py
@@ -59,7 +59,7 @@ SNMP_MAPPER_BASE = {
     },
     "cisco_xe": {
         "oid": ".1.3.6.1.2.1.1.1.0",
-        "expr": re.compile(r".*IOS-XE Software,.*", re.IGNORECASE),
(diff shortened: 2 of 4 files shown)


netmiko/paloalto/paloalto_panos.py
```diff
@@ -96,7 +96,7 @@ def commit(
             if device_and_network:
                 command_string += " device-and-network"
             if policy_and_objects:
-                command_string += " device-and-network"
+                command_string += " policy-and-objects"
```
ktbyers, line 99: This looks like a rebase issue.

netmiko/snmp_autodetect.py
```diff
@@ -59,7 +59,7 @@
     },
     "cisco_xe": {
         "oid": ".1.3.6.1.2.1.1.1.0",
-        "expr": re.compile(r".*IOS-XE Software,.*", re.IGNORECASE),
+        "expr": re.compile(r".*Catalyst L3 Switch Software.*", re.IGNORECASE),
```
ktbyers, line 62: hmmmm, I don't think that will work (i.e. that is too generic and would match platforms that should be categorized as `cisco_ios`)

ktbyers:
It looks like you need to `rebase` as there are changes in your PR that shouldn't be here.

Also, it looks like your SNMP change wouldn't work. It would need to have some identifier that indicated `IOS-XE` (if possible).

netmiko/ssh_autodetect.py
```diff
@@ -93,6 +93,12 @@
         "priority": 99,
         "dispatch": "_autodetect_std",
     },
+    "cisco_xe": {
+        "cmd": "show version",
+        "search_patterns": [r"Cisco IOS XE Software", r"IOS-EX"],
```
IceN9ne, line 98: Looks like a typo of `IOS-XE`?

ktbyers:
No response on issues with the PR...so am going to go ahead and close.

PR #1920, opened 2020-09-03
Updated TPLink JetStream PR

ktbyers:


ktbyers:
FYI, I made a few updates here: @piterpunk

ktbyers:
Changes from original PR:

1. Changed the class naming to include JetStream to be more consistent with the Netmiko format of VendorPlatform.
2. Converted `enable()` method to call super(). Basically, it helps (a lot) with ongoing Netmiko maintenance if I can reduce the number of child class methods that are NOT invoking the parent class.

Let me know if I broke anything with these changes...it looked to me like it should all work fine.

netmiko/tplink/tplink_jetstream.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        prompt = super().set_base_prompt(
+            pri_prompt_terminator=pri_prompt_terminator,
+            alt_prompt_terminator=alt_prompt_terminator,
+            delay_factor=delay_factor,
+        )
+
+        prompt = prompt.strip()
+        self.base_prompt = prompt
+        return self.base_prompt
```
ktbyers, line 104: This looked a bit odd to me since the parent class already sets `self.base_prompt()` is the extra strip and then resetting `self.base_prompt` necessary?

@piterpunk

ktbyers:
I am going to go ahead and merge this.

If there are issues/problems with the driver, then we can just fix them in subsequent PRs, but it looks good to me.

piterpunk:
> Changes from original PR:
>     2. Converted `enable()` method to call super(). Basically, it helps (a lot) with ongoing Netmiko maintenance if I can reduce the number of child class methods that are NOT invoking the parent class.

The "enable" change brokes the functionality. The prompt with enable without administrative rights and with administrative rights is the same, but the available options are different:
```
hermes>enable

hermes#
 broadcast            - Write message to all users logged in,at most 256
                         characters
 enable-admin         - Achieve the admin privilege
 logout               - Logout the system
 ping                 - Ping command
 telnet               - Telnet remote host
 terminal             - Set terminal line parameters
 clear                - Reset functions
 exit                 - Exit current mode
 history              - Display command history
 show                 - Display system information

hermes#enable-admin

hermes#
 broadcast            - Write message to all users logged in,at most 256
                         characters
 configure            - Enter Global Configuration Mode
 copy                 - Config file commands
 debug                - Debugging commands
 enable-admin         - Achieve the admin privilege
 firmware             - Firmware commands
 logout               - Logout the system
 ping                 - Ping command
 reboot               - Reboot the system
 remove               - Config file commands
 reset                - Reset the system
 telnet               - Telnet remote host
 terminal             - Set terminal line parameters
 tracert              - Tracet route to destination
 clear                - Reset functions
 exit                 - Exit current mode
 history              - Display command history
 show                 - Display system information
```
So the "enable" didn't fails to call the failback command. And run "enable-admin" even after the administrative rights are granted didn't fails. That is way I did it this way.

netmiko/tplink/tplink_jetstream.py
piterpunk: The lines 102 and 103 can be removed, they are leftovers of the driver that I used as base.

piterpunk:
@ktbyers need to run `enable` and then `enable-admin`. But when I tried to do this way, the `enable-admin` wasn't executed because it first check if it was in "enable mode" and... it was. So it didn't run the command.

ktbyers:
And I assume you can't just directly call `enable-admin` i.e. that `enable` must be called first?

piterpunk:
> And I assume you can't just directly call `enable-admin` i.e. that `enable` must be called first?

Exactly. `enable` must be called first and then `enable-admin`.

PR #1947, opened 2020-09-21
netmiko telnet multithreading script

sen2diwakar:
this is a sample netmiko multithreading script to access multiple devices simultaneously through telnet


ktbyers:
@sen2diwakar Thanks for the contribution. I have a simpler example for threading that I will use an example.

Regards, Kirk

PR #1962, opened 2020-09-24
add_ttp (ready)

tbotnz:
- adds TTP support
- need to clarify whether to vendor or use the source


tbotnz:
example of this working

```
#!/usr/bin/env python
from pprint import pprint

cisco1 = {
    "device_type": "cisco_ios",
    "host": "10.0.2.33",
    "username": "admin",
    "password": "admin",
}

# write template to file
ttp_raw_template = """
interface {{ interface }}
 description {{ description }}
"""


with open("show_run_interfaces.txt", "w") as writer:
    writer.write(ttp_raw_template)

command = "show run interface fastethernet0/1"
with ConnectHandler(**cisco1) as net_connect:
    # Use TTP to retrieve structured data
    output = net_connect.send_command(
        command, use_ttp=True, ttp_template="show_run_interfaces.txt"
    )

print()
pprint(output)
print()
```

result

```
python3.6 test.py 

[[{'description': 'cat', 'interface': 'FastEthernet0/1'}]]
```

brobare:
Suuuper cool!

brobare:
Yeah, suuuuuuper cool!!! Just normalizing some data from two vendors....

`python cisco.py && python juniper.py
Password:
[[{'loopback': {'cidr': 32,
                'interface': 'Loopback0',
                'ip_address': '10.249.17.252'}}]]
Password:
[[{'loopback': {'cidr': '32',
                'interface': 'lo0',
                'ip_address': '10.249.55.144',
                'unit': '0'}}]]


cat cisco.py
#/usr/bin/python

from pprint import pprint

import getpass
import sys

username = getpass.getuser()
passwd = getpass.getpass(prompt='Password: ')
#ip = input("IP address: ")

device = {
    "device_type": "cisco_ios",
    "host": "10.249.17.252",
    "username": username,
    "password": passwd
}

ttp_raw_template = """
<group name="loopback">
interface {{ interface }}
 description {{ description }}
 ip address {{ ip_address }} {{ cidr | to_cidr }}
</group>
"""

with open("show_run_int_lo0.txt", "w") as writer:
    writer.write(ttp_raw_template)

command = "show run int lo0"
with ConnectHandler(**device) as net_connect:
    # Use TTP to retrieve structured data
    output = net_connect.send_command(
        command, use_ttp=True, ttp_template="show_run_int_lo0.txt"
    )

pprint(output)



cat juniper.py
#/usr/bin/python

from pprint import pprint

import getpass
import sys

username = getpass.getuser()
passwd = getpass.getpass(prompt='Password: ')

device = {
    "device_type": "juniper",
    "host": "10.249.55.144",
    "username": username,
    "password": passwd
}

ttp_raw_template = """
<group name="loopback">
set interfaces {{ interface }} unit {{ unit }} family inet address {{ ip_address }}/{{ cidr }}
</group>
"""

with open("show_conf_lo0.txt", "w") as writer:
    writer.write(ttp_raw_template)

command = "show configuration interfaces lo0 | display set"
with ConnectHandler(**device) as net_connect:
    # Use TTP to retrieve structured data
    output = net_connect.send_command(
        command, use_ttp=True, ttp_template="show_conf_lo0.txt"
    )

pprint(output)`

netmiko/utilities.py
```diff
@@ -344,6 +352,31 @@ def get_structured_data(raw_output, platform=None, command=None, template=None):
...
+    Convert raw CLI output to structured data using TTP template.
+
+    You can use a straight TextFSM file i.e. specify "template"
+    """
+    if not TTP_INSTALLED:
+        msg = "\nTTP is not installed. Please PIP install ttp:\n" "pip install ttp\n"
+        raise ValueError(msg)
+
+    if os.path.exists(template):
+        with open(template) as f:
```
dmulyalin, line 366: No need to open TTP template and read it , better pass it on to TTP as is, TTP will use  [load_files](https://github.com/dmulyalin/ttp/blob/06794d8226bf406b126637208690051e0068c0a1/ttp/utils/loaders.py#L6) function to load it checking if its an OS path to file, OS path to directory of files or a text string.

netmiko/utilities.py
```diff
@@ -344,6 +352,31 @@ def get_structured_data(raw_output, platform=None, command=None, template=None):
...
+            ttp_template = f.read()
+        try:
+            if template:
+                ttp_parser = ttp(data=raw_output, template=ttp_template)
+                ttp_parser.parse()
+                # print(ttp_parser.result(format="json")[0])
+                return ttp_parser.result(format="raw")
```
dmulyalin, line 373: result methods support `structure` keyword, that is very useful to produce overall results combining them in different structure such as list of dictionaries, dictionary of flattening the lists. Moreover, result method support all the same attributes as outputters, all the way to calling functions and returners on results. suggest to introduce `ttp_res_kwarg` argument in addition to `template` keyword, `ttp_res_kwarg` can be a dictionary of attributes to unpack to result method such as:
```
return ttp_parser.result(**ttp_res_kwarg)
```

netmiko/utilities.py
```diff
@@ -344,6 +352,31 @@ def get_structured_data(raw_output, platform=None, command=None, template=None):
...
+        except Exception:
+            return raw_output
+    else:
+        raise ValueError(f"File not found: {template}")
+
```
dmulyalin, line 378: Taking all above into account suggest to try this code:
```
def get_structured_data_ttp(raw_output, template=None, ttp_res_kwarg={}):
    """
    Convert raw CLI output to structured data using TTP template.
    You can use a straight TextFSM file i.e. specify "template"
    """
    if not TTP_INSTALLED:
        msg = "\nTTP is not installed. Please PIP install ttp:\n" "pip install ttp\n"
        raise ValueError(msg)

    if template:
        try:
            ttp_parser = ttp(data=raw_output, template=template)
            ttp_parser.parse(one=True) # force to parse in single process
            # print(ttp_parser.result(format="json")[0])
            return ttp_parser.result(**ttp_res_kwarg)
        except Exception:
            return raw_output
    else:
        raise ValueError(f"Template not found: {template}")
```

tbotnz (commented):
My only concern of exposing the ttp_res_kwarg north to the connecthandler is that it breaks the consistent feel of the parser args (eg ttp & textfsm are pretty similar) however i will leave it up to @ktbyers to decide what he wants or just submitthe PR without it and we can add later

netmiko/utilities.py
dmulyalin: Sure, nw. 

One more thing, please pay attention to `ttp_parser.parse(one=True) # force to parse in single process` line and `one=True` in particular, as TTP can invoke multiprocessing mode if data to parse exceeds 5 MByte in size, that might not work out of the box when TTP used as a module, hence `one=True` a safer way of running.

tbotnz:
@ktbyers this should be ready now, let me know if you are happy,
builds failing because of pre-existing linting issues?

ktbyers:
Updated PR here:

https://github.com/ktbyers/netmiko/pull/1967

Fixes black linting issues.

I probably need to see more examples of using the arguments and the behavior. Let's go with the simpler solution for now and we can look at expanding across time.

PR #1994, opened 2020-10-06
Added two new examples

AlexMunoz905:
I added "examples/save_output_to_file.py", and "examples/send_command_two_devices.py". They do as they say. The "examples/save_output_to_file.py" creates a txt file and writes the output to it. "examples/send_command_two_devices.py" shows how to connect two devices at the same time and output the command. Obviously, they could be combined to save both outputs to two seperate files.

### DIFF
--- a/examples/save_output_to_file.py
+++ b/examples/save_output_to_file.py
@@ -1,0 +1,27 @@
+from netmiko import ConnectHandler
+from getpass import getpass
+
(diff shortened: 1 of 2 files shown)


ktbyers:
@AlexMunoz905 Thanks for the contribution.

I think I am going to pass on these as they were fairly similar to examples that already exist.

PR #2005, opened 2020-10-14
Allow DellPowerConnect to use auth with password

adrianlasota:
Added next pattern for detecting Dell OS9
Introduced Dell PowerConnect switch class for enabling auth by default ( some switches doesn't support no_auth)


ktbyers:
@adrianlasota Wouldn't you also want to add a device_type to ssh_dispatcher? In other words, how would this new class get used?

a-gorczew (approved):

hcymerys (approved):

adrianlasota:
> @adrianlasota Wouldn't you also want to add a device_type to ssh_dispatcher? In other words, how would this new class get used?

Change outdated

PR #2011, opened 2020-10-18
Added support for NIOS, an enterprise DDI solution (Marketleader)

johnneerdael:
Still some work to be done but hopefully I can get some assistance with that :)


ktbyers:
@johnneerdael Can you run the tests against your code and a test device and post the output here (the "show" and "config" tests):

Thanks and thanks for the contribution.

Issue #2014, opened 2020-10-20
Dell os10 switch fails ssh autodetect


sregister:
Hi--I'm seeing an issue with SSHDetect.autodetect() on a few dell os10 switches.

Example code: 
```#!/usr/bin/env python
from netmiko import SSHDetect, ConnectHandler
from getpass import getpass

device = {
    "device_type": "autodetect",
    "host": "switch_ip",
    "username": "admin",
    "password": getpass(),
}

guesser = SSHDetect(**device)
best_match = guesser.autodetect()
print(best_match)  # Name of the best device_type to use further
```

If switch_ip is an os9 switch best_match is `dell_os9`; if its an os10 switch I get `None`. Here is a the full `show version` output I see when I manually ssh into the os10 switch:
```switch1# show version
Dell EMC Networking OS10 Enterprise
Copyright (c) 1999-2019 by Dell Inc. All Rights Reserved.
OS Version: 10.5.0.0P1
Build Version: 10.5.0.0P1.342
Build Time: 2019-08-12T22:06:43+0000
System Type: Z9332F-ON
Architecture: x86_64
Up Time: 5 weeks 5 days 20:55:31
```

I'm not sure exactly at which level this is breaking; I'll update if I find out anything more.

sregister:
After working with @penguin359, this might be caused by the switch being stuck in a pager when `show version` is executed causing an impartial read.  When `autodetect()` starts walking through the elements in `SSH_MAPPER_BASE` it runs alacatel_aos's command "show system" which is a valid command on os10 (but not the one we're looking for) which opens a pager. The next switch it tries is alcatel_sros's "show version" which is the correct command for dell_os10 but the output is incomplete and doesn't contain the string we are looking for. When it comes around to checking for dell_os10's search patterns the cache is incomplete and it fails. 

I'm trying to lookup the command to globally disable paging (for all ttys forever) but this might not exist.

ktbyers:
@sregister Netmiko's autodetect mechanism requires there to be no disabling of paging. In other words, since we don't know the platform, there is no way we can disable the paging of output. Though maybe you were looking to disable paging in your devices as a one-off operation (and not as part of Netmiko's autodetect mechanism).

I guess we could potentially re-order such that `show version` comes first...and `show system` comes after that.

sregister:
> I guess we could potentially re-order such that `show version` comes first...and `show system` comes after that.

[...]

PR #2015, opened 2020-10-21
ssh autodetect: Fix issue with dellos10

sregister:
On dell os10 certain switch commands open a pager so autodetect isn't able to
correctly parse the os version string.

Avoid this issue by reordring SSH_MAPPER_BASE to avoid this edge case.

Fixes: https://github.com/ktbyers/netmiko/issues/2014

### DIFF
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -40,6 +40,7 @@ Examples
 """
 import re
 import time
+from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


adrianlasota:
With this fix you can broke other switches and his flow with paged commands

ktbyers:
I created a more generic fix here (which sorts SSH_MAPPER_BASE by the most common command):

https://github.com/ktbyers/netmiko/pull/2032

PR #2019, opened 2020-10-25
Update AdtranOS enable pattern

josephwhite13:
I'm not sure what flavor of Adtran the current implementation is built against, but below is an example of Adtran's TA (TotalAccess) series cli.
```
SomeHostname>enable   
Password: 
SomeHostname#
```


ktbyers:
@BSpendlove Any comments on this PR? It looks reasonable to me so I will probably merge it unless you see issues.

BSpendlove:
That looks good, just for reference, all my tests were against multiple models of the total access line up @josephwhite13

PR #2032, opened 2020-11-06
Fixing ssh_autodetect so most common command happens first.

ktbyers:


ktbyers:
@adrianlasota Does this break some platform? In other words, is there some platform that is counting on the previous hard-coded order? 

If so, which ones?

adrianlasota:
> @adrianlasota Does this break some platform? In other words, is there some platform that is counting on the previous hard-coded order?
> 
> If so, which ones?

I check Dell OS9 and Dell PowerConnect, it's working correctly

PR #2052, opened 2020-11-23
Add exception info to NetmikoTimeoutException message

mk-fg:
Hi,

I was getting this NetmikoTimeoutException with current netmiko code:
```
TCP connection to device failed.

Common causes of this problem are:
1. Incorrect hostname or IP address.
2. Wrong TCP port.
3. Intermediate firewall blocking access.

Device settings: cisco_ios 10.0.0.98:22
```

One general issue with this mesage is that it's completely inaccurate:

- It was an SSH connection, which I could see clearly established, passing handshake and such - there was no issue of any kind on TCP level.

- Can you guess what was the actual error there?
    `[PermissionError] [Errno 13] Permission denied: '/etc/netbox/ssh_key'` - very unclear from above error, obviously.

I think this happens because socket.error is an alias for OSError, so applies to any kind of weirdness that has nothing to do with TCP or sockets.

This PR tries to address this in an easiest way possible - simply log the actual exception type and message, so that one can tell what the problem is, hopefully, instead of leading into a wrong and confusing direction (debugging ssh protocol and network stuff), and without the need to patch netmiko to get the true error info instead of false placeholder.

One potential caveat I see here is that sensitive information might propagate to exception message if it can end up in OSError.
But not sure if python can raise such OSErrors, and don't think paramiko should (that'd be a bug there), so probably not an issue?

paramiko seem to try actually reading key as late as possible, after establishing ssh connection and after `debug1: KEX done [preauth]` on sshd side (openssh `sshd -dddddd` output), which makes this non-connection error less obvious (as opposed to failing before even trying).
Not sure if maybe it's also a bug there, but pretty sure reporting all OSErrors as TCP issues is still worth addressing here.

Thanks.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -934,6 +934,7 @@ Common causes of this problem are:
 3. Intermediate firewall blocking access.
 
 Device settings: {self.device_type} {self.host}:{self.port}
+Error info: [{conn_error.__class__.__name__}] {conn_error}
 
 """
 


mk-fg:
> One potential caveat I see here is that sensitive information might propagate to exception message if it can end up in OSError.
> But not sure if python can raise such OSErrors, and don't think paramiko should (that'd be a bug there), so probably not an issue?

Also, if this is a reason for masking the real exception here, might be worth adding a comment to exception message to that effect - i.e. something like `It's likely TCP error, but can actually be absolutely any python OSError - permission denied, syscall blocking, out of memory, or whatever else - hidden here for $REASON`.

ktbyers:
@mk-fg Can you post your original full stack trace from the original issue?

Regards,

Kirk

mk-fg:
Oh, sorry, didn't think about it, here:
```
  File "/usr/local/lib/python3.8/site-packages/netmiko/base_connection.py", line 921, in establish_connection
    self.remote_conn_pre.connect(**ssh_connect_params)
  File "/usr/local/lib/python3.8/site-packages/paramiko/client.py", line 435, in connect
    self._auth(
  File "/usr/local/lib/python3.8/site-packages/paramiko/client.py", line 676, in _auth
    key = self._key_from_filepath(
  File "/usr/local/lib/python3.8/site-packages/paramiko/client.py", line 586, in _key_from_filepath
    key = klass.from_private_key_file(key_path, password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/pkey.py", line 235, in from_private_key_file
    key = cls(filename=filename, password=password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/rsakey.py", line 55, in __init__
    self._from_private_key_file(filename, password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/rsakey.py", line 175, in _from_private_key_file
    data = self._read_private_key_file("RSA", filename, password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/pkey.py", line 307, in _read_private_key_file
    with open(filename, "r") as f:
PermissionError: [Errno 13] Permission denied: '/etc/netbox/ssh_key'
```

ktbyers:
And what did it look like all together i.e. the full stack trace and the Netmiko exception message?

mk-fg:
In my case, netmiko was used by NAPALM framework, which itself was used by NetBox.

In the NetBox WebUI, error liiked like this:
`Error connecting to the device at 10.0.0.98: Cannot connect to 10.0.0.98`

So I didn't see netmiko error there, as NAPALM does same thing as netmiko and just hides everything for some reason:
```
        try:
            self._netmiko_device = ConnectHandler(
                device_type=device_type,
                host=self.hostname,
                username=self.username,
                password=self.password,
                timeout=self.timeout,
                **netmiko_optional_args
            )
        except NetMikoTimeoutException:
            raise ConnectionException("Cannot connect to {}".format(self.hostname))
```

If I patch that to do `traceback.print_exc()`, this is the output:
```
  File "/usr/local/lib/python3.8/site-packages/netmiko/base_connection.py", line 920, in establish_connection
    self.remote_conn_pre.connect(**ssh_connect_params)
  File "/usr/local/lib/python3.8/site-packages/paramiko/client.py", line 435, in connect
    self._auth(
  File "/usr/local/lib/python3.8/site-packages/paramiko/client.py", line 676, in _auth
    key = self._key_from_filepath(
  File "/usr/local/lib/python3.8/site-packages/paramiko/client.py", line 586, in _key_from_filepath
    key = klass.from_private_key_file(key_path, password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/pkey.py", line 235, in from_private_key_file
    key = cls(filename=filename, password=password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/rsakey.py", line 55, in __init__
    self._from_private_key_file(filename, password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/rsakey.py", line 175, in _from_private_key_file
    data = self._read_private_key_file("RSA", filename, password)
  File "/usr/local/lib/python3.8/site-packages/paramiko/pkey.py", line 307, in _read_private_key_file
    with open(filename, "r") as f:
PermissionError: [Errno 13] Permission denied: '/etc/netbox/ssh_key'

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.8/site-packages/napalm/base/base.py", line 89, in _netmiko_open
    self._netmiko_device = ConnectHandler(
  File "/usr/local/lib/python3.8/site-packages/netmiko/ssh_dispatcher.py", line 312, in ConnectHandler
    return ConnectionClass(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/netmiko/cisco/cisco_ios.py", line 17, in __init__
    return super().__init__(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/netmiko/base_connection.py", line 346, in __init__
    self._open()
  File "/usr/local/lib/python3.8/site-packages/netmiko/base_connection.py", line 351, in _open
    self.establish_connection()
  File "/usr/local/lib/python3.8/site-packages/netmiko/base_connection.py", line 942, in establish_connection
    raise NetmikoTimeoutException(msg)
netmiko.ssh_exception.NetmikoTimeoutException: TCP connection to device failed.

Common causes of this problem are:
1. Incorrect hostname or IP address.
2. Wrong TCP port.
3. Intermediate firewall blocking access.

Device settings: cisco_ios 10.0.0.98:22
```

But as any higher-level code will never print tracebacks, only error messages - if even that, as neither netmiko nor napalm seem to actually do it atm - I think it'd still be useful to propagate error messages, unless there is a reason not to.

mk-fg:
Also opened https://github.com/napalm-automation/napalm/pull/1328 to maybe also pass netmiko error info there.

ktbyers:
Hmmmm, so if I ran in Python I am pretty sure I would see the full exception stack trace including the original underlying exception so then my first question is why don't you see that?

Is that because of Netbox?

I am pretty sure you would see all of this in straight Python with the current code.

The reason you want to use a Netmiko or NAPALM exception is that it greatly simplifies the job of exception handling for people writing code. In other words, instead of having to capture a bunch of different exceptions they can just capture the `NetmikoTimeoutException` in that context.

mk-fg:
> The reason you want to use a Netmiko or NAPALM exception is that it greatly simplifies the job of exception handling for people writing code.

I understand that, and not arguing for raising raw OSError or anything like that at all, and that's not what change in the PR does.

Pretty sure exceptions like NetmikoTimeoutException are then almost always printed as "ERROR: {error}" in all user-facing applications, so why not include helpful and actionable error message for these, instead of unhelpful static one?

mk-fg:
> When caught in that way, pretty sure exceptions like NetmikoTimeoutException are almost always printed as "ERROR: {error}" in all user-facing applications (just like NetBox does it)

I'd also argue that reporting nested tracebacks is something that maybe user-facing apps should never do - not only it's very confusing to anyone but the developer (i.e. faced with long code traceback, they tend to assume it's a bug), but also just not very helpful, as it's intended for developers, and using error messages written in a reasonably-understandable english are ok solution here, and what they seem to be intended for.

> Hmmmm, so if I ran in Python I am pretty sure I would see the full exception stack trace including the original underlying exception so then my first question is why don't you see that?
> Is that because of Netbox?

Given the above, don't think it's NetBox fault for relying on error messages for reporting errors to users, but rather a problem with underlying modules that these are not being generated properly.

ktbyers:
I probably disagree with you here, but let me think about it some more.

mk-fg:
> I probably disagree with you here, but let me think about it some more.

Regardless of the PR, I'd be interested to learn more of your position on this issue, as not sure if I understand it correctly.

ktbyers:
I am going to close this as it is old and I think I disagree with the general premise. 

Properly handling errors are difficult though (and different people can disagree on what even is "properly handling").

PR #2063, opened 2020-12-02
SCP nokia_sros: Fix to remote_file_size()

jkierklo:
This is supplementary fix to  #1724 - handling file_transfer to non_root directories.

Only filename is shown in the 'file dir' command output, need to isolate file basename for regex match. When uploading file to non-root location remote_file_size() method fails. It tries to match whole destination 'directory/file' against 'file dir directory/file' output, which contains only final path component:

```
*A:GJJAM003PAR01# file dir directory/file
Volume in drive cf3 on slot A has no label.

Volume in drive cf3 on slot A is formatted as FAT16

Directory of cf3:\directory

02/05/2021  03:42p                   2 file
               1 File(s)                      2 bytes.
               0 Dir(s)               146472960 bytes free.
```

```
with netmiko.Netmiko(**crede) as conn:
    file_transfer(
        conn, 
        source_file='file', 
        dest_file='directory/file', 
        file_system='cf3:', 
        overwrite_file=True, 
        direction='put'
)
```

gives:
```
  File "/tmp/crap.py", line 14, in <module>
    file_transfer(conn, source_file='file', dest_file='images/file', file_system='cf3:', overwrite_file=True, direction='put')
  File "/tmp/venv/lib/python3.9/site-packages/netmiko/scp_functions.py", line 137, in file_transfer
    if scp_transfer.verify_file():
  File "/tmp/venv/lib/python3.9/site-packages/netmiko/nokia/nokia_sros_ssh.py", line 297, in verify_file
    return os.stat(self.source_file).st_size == self.remote_file_size(
  File "/tmp/venv/lib/python3.9/site-packages/netmiko/nokia/nokia_sros_ssh.py", line 289, in remote_file_size
    raise ValueError("Filename entry not found in dir output")
ValueError: Filename entry not found in dir output
```

### DIFF
--- a/netmiko/nokia/nokia_sros_ssh.py
+++ b/netmiko/nokia/nokia_sros_ssh.py
@@ -279,7 +279,7 @@ class NokiaSrosFileTransfer(BaseFileTransfer):
         if "File Not Found" in remote_out:
             raise IOError("Unable to find file on remote system")
 
-        dest_file_name = remote_file.replace('\\', '/').split('/')[-1]
+        dest_file_name = remote_file.replace("\\", "/").split("/")[-1]
         # Parse dir output for filename. Output format is:
         # "10/16/2019  10:00p                6738 {dest_file_name}"
 
--- a/netmiko/nokia/nokia_sros_ssh.py
+++ b/netmiko/nokia/nokia_sros_ssh.py
@@ -248,7 +248,7 @@ class NokiaSrosFileTransfer(BaseFileTransfer):
 
         if self.direction == "put":
             if not remote_cmd:
-                remote_cmd = self._file_cmd_prefix() + "file dir {}\\{}".format(
+                remote_cmd = self._file_cmd_prefix() + "file dir {}/{}".format(
                     self.file_system, self.dest_file
                 )
             dest_file_name = self.dest_file.replace("\\", "/").split("/")[-1]
--- a/netmiko/nokia/nokia_sros_ssh.py
+++ b/netmiko/nokia/nokia_sros_ssh.py
@@ -248,7 +248,7 @@ class NokiaSrosFileTransfer(BaseFileTransfer):
 
         if self.direction == "put":
             if not remote_cmd:
-                remote_cmd = self._file_cmd_prefix() + "file dir {}/{}".format(
+                remote_cmd = self._file_cmd_prefix() + "file dir {}\\{}".format(
                     self.file_system, self.dest_file
                 )
             dest_file_name = self.dest_file.replace("\\", "/").split("/")[-1]
@@ -279,10 +279,11 @@ class NokiaSrosFileTransfer(BaseFileTransfer):
         if "File Not Found" in remote_out:
             raise IOError("Unable to find file on remote system")
 
+        dest_file_name = remote_file.replace('\\', '/').split('/')[-1]
         # Parse dir output for filename. Output format is:
-        # "10/16/2019  10:00p                6738 {filename}"
+        # "10/16/2019  10:00p                6738 {dest_file_name}"
 
-        pattern = r"\S+\s+\S+\s+(\d+)\s+{}".format(re.escape(remote_file))
+        pattern = r"\S+\s+\S+\s+(\d+)\s+{}".format(re.escape(dest_file_name))
         match = re.search(pattern, remote_out)
 
         if not match:


jkierklo:
Attaching tests results:
image

pszulczewski:
Dear Kirk,

Is there anything else we can do so you can accept this PR?
I asked @jkierklo to run tests and supply the test output.
Most of our deployments for SROS need extra patch after `pip install netmiko`, accepting this PR would make our lives easier.

Thanks,
--
Patryk

/cc @ktbyers 
/cc @hellt

netmiko/nokia/nokia_sros_ssh.py
```diff
@@ -248,7 +248,7 @@ def check_file_exists(self, remote_cmd=""):
 
         if self.direction == "put":
             if not remote_cmd:
-                remote_cmd = self._file_cmd_prefix() + "file dir {}/{}".format(
+                remote_cmd = self._file_cmd_prefix() + "file dir {}\\{}".format(
```
hellt, line 251: I wonder why we needed to use `\\` here? Why not keeping `/`?

netmiko/nokia/nokia_sros_ssh.py
```diff
@@ -279,10 +279,11 @@ def remote_file_size(self, remote_cmd=None, remote_file=None):
         if "File Not Found" in remote_out:
             raise IOError("Unable to find file on remote system")
 
+        dest_file_name = remote_file.replace("\\", "/").split("/")[-1]
```
hellt, line 282: why this replacement is needed? what use case does it cover?

netmiko/nokia/nokia_sros_ssh.py
pszulczewski: Agree better to keep single \ like in other lines.

netmiko/nokia/nokia_sros_ssh.py
pszulczewski: Replace is indeed redundant, can be done as:
dest_file_name = remote_file.split("\\")[-1]
This is needed to separate filename from the path.

pszulczewski: Just talked to @jkierklo.

In SROS path can be specified with \ and with / they both will work.
This replace is needed to uniform regardless of what file/path is given split is done always on the same character.

pszulczewski: Users can specify file path as:
7705-TiMOS-8.0.R7**/**BOOT.LDR 
```A:LAB-SAR-8-2# file dir cf3:\7705-TiMOS-8.0.R7/BOOT.LDR 


Volume in drive cf3 on slot A has no label.

Volume in drive cf3 on slot A is formatted as FAT32

Directory of cf3:\7705-TiMOS-8.0.R7

09/26/2017  08:47p             2796640 BOOT.LDR
               1 File(s)                2796640 bytes.
               0 Dir(s)              1820561408 bytes free.```

Or like
7705-TiMOS-8.0.R7**\**BOOT.LDR 

```A:LAB-SAR-8-2# file dir cf3:\7705-TiMOS-8.0.R7\BOOT.LDR 


Volume in drive cf3 on slot A has no label.

Volume in drive cf3 on slot A is formatted as FAT32

Directory of cf3:\7705-TiMOS-8.0.R7

09/26/2017  08:47p             2796640 BOOT.LDR
               1 File(s)                2796640 bytes.
               0 Dir(s)              1820561408 bytes free.```

Comment (hellt) [reply netmiko/nokia/nokia_sros_ssh.py:282]: yes, but why double backslashes? If, imagine, my path is `dir1\dir2\file` what will `\\` substitute?

Comment (pszulczewski) [reply netmiko/nokia/nokia_sros_ssh.py:282]: That's python ... first backslash is escape for the second.
```
>>>x = 'cf3:\dir\file'
>>> x
>>>'cf3:\\dir\x0cile'
>>>x = 'cf3:\\dir\\file'
>>>x
'cf3:\\dir\\file'
x.split('\\')
['cf3:', 'dir', 'file']
```

hellt: ah, ok =) makes sense then

pszulczewski: I asked @jkierklo to make line 251 back to / and we all should be good.

ktbyers:
I will see what I can do, but will probably be hard for me to get to this before next week.

ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/2190

pszulczewski:
Thank you Kirk, we really appreciate it.

Issue #2072, opened 2020-12-13
use netmiko-develop to see if IOXR doesnt detect None


muhammad-rafi:
`ios` and `ios_xe` are identical in Netmiko...so autodetect will use the more general `ios` selection.

I think  the IOS-XR issue is likely fixed in the Netmiko `develop` branch (if you want to test that).

Just re-open if there is still an issue here.

Regards, Kirk

_Originally posted by @ktbyers in https://github.com/ktbyers/netmiko/issues/2067#issuecomment-743261596_

muhammad-rafi:
Following my previous issue with ios_xr detecting none with autodetect, as per the Kirk advice I tried with netmiko-develop branch, the results are different and still cannot get the os platform for ios xr 

Here is the output of the ones I tested with develop branch and last one is for the io xr 

```
C:\Users\rafim1\Downloads>python netmiko_autodetect_2.py
cisco_ios

C:\Users\rafim1\Downloads>python netmiko_autodetect_2.py
cisco_nxos

C:\Users\rafim1\Downloads>python netmiko_autodetect_2.py
dell_os10

C:\Users\rafim1\Downloads>python netmiko_autodetect_2.py
cisco_asa

C:\Users\rafim1\Downloads>python netmiko_autodetect_2.py
Traceback (most recent call last):
File "netmiko_autodetect_2.py", line 35, in
main()
File "netmiko_autodetect_2.py", line 30, in main
osdetect(host)
File "netmiko_autodetect_2.py", line 22, in osdetect
best_match = os_detect.autodetect()
File "C:\Users\rafim1\Downloads\netmiko\ssh_autodetect.py", line 286, in autodetect
accuracy = autodetect_method(**tmp_dict)
TypeError: _autodetect_remote_version() got an unexpected keyword argument 'cmd'`

muhammad-rafi:
I have placed netmiko package back to the default package directory and no difference in the results, just in case. 

``
C:\Users\rafim1\Downloads>python netmiko_autodetect_2.py
Traceback (most recent call last):
  File "netmiko_autodetect_2.py", line 35, in <module>
    main()
  File "netmiko_autodetect_2.py", line 30, in main
    osdetect(host)
  File "netmiko_autodetect_2.py", line 22, in osdetect
    best_match = os_detect.autodetect()
[...]

PR #2076, opened 2020-12-16
Fix _autodetect_remote_version in ssh_autodetect

Frazew:
# Error

When running SSHDetect on an unknown device, there is a chance for the `autodetect` method to raise the following exception:

```
_autodetect_remote_version() got an unexpected keyword argument 'cmd'
```

This happens when the `autodetect` passes over the `cisco_wlc` type and calls `_autodetect_remote_version` with an extra unexpected argument `"cmd"`.

# Analysis

- fd67d8ff2741cfd3a2dd391a34fcca8c304cb87c introduced a new detection method based on the SSH signature (`_autodetect_remote_version`).
- 66dccecf5e59bb002ab9b5c91c297b285b2b996c later introduced a method to sort the commands to run based on the most frequent ones and introduced with it a `"cmd"` key in the `cisco_wlc` device dict (https://github.com/ktbyers/netmiko/blob/develop/netmiko/ssh_autodetect.py#L192).
- This led the `autodetect_method` (https://github.com/ktbyers/netmiko/blob/develop/netmiko/ssh_autodetect.py#L286) to have an extra unexpected `cmd` argument when `autodetect_method` is `_autodetect_remote_version`. 

# Fix

This PR addresses this issue by adding the `cmd` keyword argument to the signature of `_autodetect_remote_version`.

I believe that in the future attention should be paid so as to make sure that all new detection methods added match the signature of the base `_autodetect_std` method.

### DIFF
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -347,7 +347,7 @@ class SSHDetect(object):
             return cached_results
 
     def _autodetect_remote_version(
-        self, search_patterns=None, re_flags=re.IGNORECASE, priority=99
+        self, cmd="", search_patterns=None, re_flags=re.IGNORECASE, priority=99
     ):
         """
         Method to try auto-detect the device type, by matching a regular expression on the reported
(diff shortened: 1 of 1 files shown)


Frazew:
Issue #2072 seems to be related to this

ktbyers:
This issue should be fixed here:

https://github.com/ktbyers/netmiko/commit/bc9700a803ccd89e29672dbe544368b946352aa0#diff-f75c75643fef9ba2a42b2fa97929bee7a6d398b33ea445d0dafd73017c728213

PR #2086, opened 2020-12-25
Resolves ktbyers/netmiko#2079

### LINKED ISSUE #2079: ssh_autodect can't detect IOS-XR
Dear developers,

As discussed within #1685, ssh_audotedect relies on routers to returns the output of _show version_ [within 1 second](https://github.com/ktbyers/netmiko/blob/753cb53aca7692e8e67d3386b7614788f2b34d06/netmiko/ssh_autodetect.py#L321), while that can take more than 5 on some IOS-XR routers such as ASR9k.

In their great wisdom, Cisco have developed _show version brief_ which always returns output fast (at least in my empirical tests), but which looks to be available only on certain platforms: https://github.com/ansible/ansible/pull/37609

lamehost:
Make the delay between write to and read from CLI in autodetect() a parameter

### DIFF
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -392,7 +392,12 @@ class SSHDetect(object):
         return 0
 
     def _autodetect_std(
-        self, cmd="", search_patterns=None, re_flags=re.IGNORECASE, priority=99, read_after=1
+        self,
+        cmd="",
+        search_patterns=None,
+        re_flags=re.IGNORECASE,
+        priority=99,
+        read_after=1,
     ):
         """
         Standard method to try to auto-detect the device type. This method will be called for each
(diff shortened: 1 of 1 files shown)


ktbyers:
I made a comment about another possible fix back in the main issue #2079

ktbyers:
@lamehost Does this change here--fix the underlying problem?

https://github.com/ktbyers/netmiko/pull/2131/files#diff-c90931e41c9586e828c2c188c18af6cffa053a56fd099e7425cd410e62d31a27R622-R626

That should be in the Netmiko 3.3.3 released code (basically it will cause the _read_channel_timing() call to wait quite a bit longer to get the entire output).

ktbyers:
I am going to close this as probably fixed (see above).

If it is still an issue, we should re-open a new issue on it.

Issue #2021, opened 2020-10-30
Cisco Nexus 9000 failing on slow command


babilat:
Netmiko is quitting on slow running commands "show run" on Cisco Nexus 9000. I have tried using "expect_string={hostname}#, delay factor of 4 and neither works. Netmiko documentation says it should take atleast 100 seconds, but it seems to be failing around 20 seconds into the run. 
#####My code############
device = {
    "device_type": "cisco_nxos",
    "host": "xx.xx.xx.xx",
    "username": "username",
    "password": "passwd",
    "session_log": "session.out",
}

command = "show run | xml | no-more"
with ConnectHandler(**device) as net_connect:
    print(net_connect.find_prompt())
    output = net_connect.send_command(command, delay_factor=4)

#############error#########################
Traceback (most recent call last):
  File "netmiko_detect.py", line 30, in <module>
    response = connection.send_command(command, delay_factor=4)
  File "/usr/local/lib/python3.6/dist-packages/netmiko/utilities.py", line 429, in wrapper_decorator
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/netmiko/base_connection.py", line 1525, in send_command
    search_pattern
OSError: Search pattern never detected in send_command: #2020

ktbyers:
Is that the full exception stack trace?

You probably need to include the Netmiko log file from the failure case.

https://github.com/ktbyers/netmiko/blob/develop/COMMON_ISSUES.md#enable-netmiko-logging-of-all-reads-and-writes-of-the-communications-channel

Also which version of Netmiko are you using. You need to be using the latest released Netmiko version.


Regards, Kirk

babilat:
I am using netmiko==3.3.2. I don't get much when trying to get full exception stack trace with traceback.

babilat:
[test.log](https://github.com/ktbyers/netmiko/files/5466424/test.log)

babilat:
Attached the log file.

ktbyers:
If you do this manually, how long (roughly) does it take to output data:

```
!Command: show running-config
!Running configuration last done at: Tue Oct 13 13:50:54 2020
!Time: Fri Oct 30 16:38:49 2020

version 7.0(3)I7(6) Bios:version 07.65 

DEBUG:netmiko:read_channel: ******************************************
This may take time. Please be patient.
******************************************

```

ktbyers:
By this, I mean the following command:

`show run | xml | no-more`

[...]

Issue #1996, opened 2020-10-07
Create a netmiko read-loop timeout in Netmiko 4.x


ktbyers:

ktbyers:
This is done in Netmiko 4.x

Issue #1959, opened 2020-09-23
Cisco NX-OS fails "send_command_timing" when executing "show version"


ktbyers:
This is post the performance fixes in Netmiko develop branch.

ktbyers:
I think this is fixed.

Issue #1852, opened 2020-07-17
Fortigate: Search pattern never detected in send_command_expect


lvrfrc87:
I am trying to send some config lines with Netmiko to a Fortigate firewall.
These are the lines I am sending:

```
config global
config system interface
edit "b_t666_vl666"
set vdom "BLUE"
set ip 8.8.8.1 255.255.255.252
set type vlan
set interface "ag1_int"
set vlanid 666
set alias "test_t666_vl666"
end
end
```

When I try to push the same config via netmiko, I get the below error message:

```
Traceback (most recent call last):
  File "/root/.ansible/tmp/ansible-local-2383c0yw6152/ansible-tmp-1594995400.8889592-219300089313818/AnsiballZ_fortios_ssh.py", line 102, in <module>
    _ansiballz_main()
  File "/root/.ansible/tmp/ansible-local-2383c0yw6152/ansible-tmp-1594995400.8889592-219300089313818/AnsiballZ_fortios_ssh.py", line 94, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "/root/.ansible/tmp/ansible-local-2383c0yw6152/ansible-tmp-1594995400.8889592-219300089313818/AnsiballZ_fortios_ssh.py", line 40, in invoke_module
    runpy.run_module(mod_name='ansible.modules.network.fortios_ssh', init_globals=None, run_name='__main__', alter_sys=True)
  File "/usr/lib64/python3.6/runpy.py", line 205, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/ansible_fortios_ssh_payload__ow_58y5/ansible_fortios_ssh_payload.zip/ansible/modules/network/fortios_ssh.py", line 130, in <module>
  File "/tmp/ansible_fortios_ssh_payload__ow_58y5/ansible_fortios_ssh_payload.zip/ansible/modules/network/fortios_ssh.py", line 72, in send_lines
  File "/usr/local/lib/python3.6/site-packages/netmiko/base_connection.py", line 1337, in send_command
    search_pattern
OSError: Search pattern never detected in send_command_expect: Fw1_prd_r3b15_ams7\ \#
```

[...]

PR #2092, opened 2020-12-30
New, promptless mode to collect device output

dmulyalin:
Would like to discuss promptless mode to read data for show commands output. 

# Why?

Existing methods have issues that proved to be solved using promptless mode, in particular:

- Read till pattern based on prompt not always reliable and Netmiko fails with `Search pattern never detected in send_command:` error, increasing timeouts usually makes problem less intense but does not solves it completely and might lead to increased execution time
- Using read timing mode seems prone to problems with either waiting for too long or finishing waiting too early, both resulting in behavior that would be good to avoid

# What?

Promptless mode allows to detect end of output from device without relying on timers or correct prompt matching (hence the name - promptless). This mode still uses pattern to decide if device finished emitting data, but that pattern is not Dependant on device's prompt.

# How?

promptless_mode_v0 1

Each reading cycle, data from device read as fast as possible until device either finishes or pose to prepare more data. Latter case detected and handled using read timeout timer and checking if new data received. To detect when device finishes producing output, algorithm sends two space character to device and checks in next read cycle if last line contains two additional spaces, concluding that end of output detected if so and continue cycling otherwise.

Overall its similar to how Humans interact with device prompt to verify that its still operational - try hitting space or type something in terminal to see if its appears on the screen.

Do not think I am the first one to came up with similar idea, however, this pull request includes code that was successfully tested with a number of drivers.

# Testing done

So far tested show commands output gathering for drivers:

- Cisco IOS
- Cisco NXOS
- Cisco IOSXR
- Juniper Junos
- Linux 

Was having aforementioned issues with Cisco NXOS driver - failing to detect pattern using `send_command` or skipping some output using `send_command_timing` methods, problem gone away once switched to promptless mode. Left it running in my lab for several days,  `send_command_ps` collected 198240 show commands outputs with no errors detected thus far.

As a result, decided to share my findings and discuss if promptless mode could be of the benefit to wider community.

Obviously, more testing required, but new mode could be introduced in codebase in a backward compatible way as it does not changes any other code, same could be true for removing it in the future in case it deemed useless.

Interesting byproduct of promptless mode is that device prompt can change while commands executed with end of output properly detected, i.e. this works fine in prompless mode:

```

cisco = { 
'device_type': 'cisco_ios', 
'host': '192.168.217.10', 
'username': 'cisco', 
'password': 'cisco', 
}  

net_connect_cisco = ConnectHandler(**cisco)

promptless_mode_output = net_connect_cisco.send_command_ps("""
show ip int brief
conf t
interface loopback 100
ip address 1.1.1.100 255.255.255.255
end
show ip int brief
wr
""")
 
print(promptless_mode_output)

# prints:
# lab-rt1#show ip int brief
# Interface                  IP-Address      OK? Method Status                Protoc
# Ethernet0/0.2000           192.168.217.10  YES NVRAM  up                    up
# lab-rt1#conf t
# Enter configuration commands, one per line.  End with CNTL/Z.
# lab-rt1(config)#interface loopback 100
# lab-rt1(config-if)#ip address 1.1.1.100 255.255.255.255
# lab-rt1(config-if)#end
# lab-rt1#show ip int brief
# Interface                  IP-Address      OK? Method Status                Protocol
# Ethernet0/0.2000           192.168.217.10  YES NVRAM  up                    up
# Loopback100                1.1.1.100       YES manual up                    up
# lab-rt1#wr
# Building configuration...
# [OK]
```

Another byproduct, is the fact that majority of devices use space char for pagination, as a result even with pagination turned on, promptless mode will collect full output from such devices.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -2036,6 +2036,157 @@ Device settings: {self.device_type} {self.host}:{self.port}
             self.session_log.close()
             self.session_log = None
 
+    def send_command_ps(
+        self,
+        command_string,
+        read_timeout=30,
+        timeout=120,
+        inter_loop_sleep=0.1,
+        strip_prompt=True,
+        strip_command=True,
+        normalize=True,
+        use_textfsm=False,
+        textfsm_template=None,
+        use_ttp=False,
+        ttp_template=None,
+        use_genie=False,
+    ):
+        """Execute command_string_ps on the SSH channel using promptless (ps) approach. Can be used
+        for any commands, including commands that change prompt. Multiple commands can be sent
+        separated by '\n' newline.
+
+        :param command_string: The command(s) to be executed on the remote device.
+        :type command_string: str
+
+        :param read_timeout: Timeout in seconds to wait for data from devices, default 30s, if
+            set to -1 will wait indefinitely
+        :type read_timeout: int
+
+        :param timeout: Timeout in seconds of overall wait, default 120s, if
+            set to -1 will wait indefinitely
+        :type timeout: int
+
+        :param inter_loop_sleep: Interval in seconds to sleep between loops, default 0.1s
+        :type inter_loop_sleep: int
+
+        :param strip_prompt: Remove the trailing router prompt from the output (default: True).
+        :type strip_prompt: bool
+
+        :param strip_command: Remove the echo of the command from the output (default: True).
+        :type strip_command: bool
+
+        :param normalize: Ensure the proper enter is sent at end of command (default: True).
+        :type normalize: bool
+
+        :param use_textfsm: Process command output through TextFSM template (default: False).
+        :type use_textfsm: bool
+
+        :param textfsm_template: Name of template to parse output with; can be fully qualified
+            path, relative path, or name of file in current directory. (default: None).
+        :type textfsm_template: str
+
+        :param use_ttp: Process command output through TTP template (default: False).
+        :type use_ttp: bool
+
+        :param ttp_template: Name of template to parse output with; can be fully qualified
+            path, relative path, or name of file in current directory. (default: None).
+        :type ttp_template: str
+
+        :param use_genie: Process command output through PyATS/Genie parser (default: False).
+        :type use_genie: bool
+        """
+        data_received = ""
+        previous_last_line = ""
+        no_data_elapsed = 0
+        start_time = time.time()
+
+        if normalize:
+            command_string = self.normalize_cmd(command_string)
+
+        self.clear_buffer()
+        self.write_channel(command_string)
+
+        # Main loop to get output from device
+        while True:
+            if read_timeout != -1 and no_data_elapsed > read_timeout:
+                raise NetmikoTimeoutException(
+                    "send_command_ps {}s read_timeout expired".format(read_timeout)
+                )
+            if timeout != -1 and (time.time() - start_time) > timeout:
+                raise NetmikoTimeoutException(
+                    "send_command_ps {}s timeout expired".format(timeout)
+                )
+
+            time.sleep(inter_loop_sleep)
+            no_data_elapsed += inter_loop_sleep
+
+            # read data from channel
+            chunk = self._read_channel()
+            if chunk:
+                no_data_elapsed = 0
+                data_received += chunk
+                new_last_line = data_received.splitlines()[-1]
+            else:
+                continue
+
+            # handle edge cases
+            if (
+                # ignore empty last line
+                not new_last_line.strip(" ")
+                # Unix systems echo command back before executing, ignore it
+                or new_last_line == command_string
+            ):
+                continue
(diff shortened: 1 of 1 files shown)


dmulyalin:
To anticipate concerns raised in https://github.com/ktbyers/netmiko/issues/1830:
- How would multi-line prompting work? I.E. things like confirm [y/n]? - multiline string could be sent `reboot\nY`
- how reliably I could count on cntl-a and cntl-k to work across 70+ - `send_command_ps` uses spaces to detect end of output, nothing special, nonetheless, found issue with Juniper  echoed `\x07` back instead of space - this can be handled as edge-cases, hence the emphasis that more test data required
- Why would it make it any faster at detecting the prompt compared to a regex pattern? - it would not, it makes it more reliable as per testing done for Cisco NXOS, without sacrificing any speed though
-  need to strip out all the prompttest messages and ANSI codes and backspaces from the output - not the case here, `send_prompt_ps` calls `self.clear_buffer()` to clean up previously sent spaces, so that they would not be collected by following command output

Also, would be good to see if `send_command_ps` helps with:
https://github.com/ktbyers/netmiko/issues/1852, https://github.com/ktbyers/netmiko/issues/1959, https://github.com/ktbyers/netmiko/issues/2021

Issue https://github.com/ktbyers/netmiko/issues/1996 addressed in `send_command_ps` as well by `read_timeout` timer

ktbyers:
@dmulyalin I will need to spend some more time reviewing/looking at this.


Regards, Kirk

dmulyalin:
Confirmed promptless mode working with cisco_ios_telnet driver, will keep updating if more testing results come through.

dmulyalin:
Had to pull long output from IOS-XR devices, some of them happened to be relatively slow, used Netmiko normal mode, but it was producing `Search pattern never detected in send_command` errors for some devices. That issue gone after switching to prompt-less mode, however, noticed that not all output was collected, as end of output was mistakenly detected by ps mode probably due to IOS-XR buffer line was overflown with space characters sent to device for end of output detection, once that overflow happens, IO XR sends `^@` characters back, and in next cycle if no reply from device incoming, ps mode falsefully detects end of output. Fixed that issue by adding another edge-case handling.

Also, added initial sleep timer, as it seems to be useful in some cases like dealing with slow devices, e.g. if it takes about a second or two before device starts producing output data and we aware of that, we can set initial sleep to 2 seconds, to wait for 2 seconds after sending command before starting to collect output, that can help to save some cpu cycles and improve stability.

ktbyers:
@dmulyalin Sorry for delayed response on this. I will try to respond on this in the next couple of weeks.


Regards, Kirk

ktbyers:
@dmulyalin Sorry for delayed response on this.

I am pretty sure that I don't want to do this. My reasons being:
1. I am pretty sure from past experience that we would run into issues trying to use this solution. Devices end up doing a bunch of weird things to make the output more human readable.
2. I don't think I would want to do this just for the multiline prompting reason i.e. I view the embedding of "\n into a string as anti-pattern (so I would not want to encourage this).
3. This area is so core to Netmiko that is really hard to test and it has very big impact to Netmiko users. So I would never want to change the primary method `send_command` over to this pattern.

I am open to more discussions, however, on whether we should add this as an experimental method into Netmiko.



Regards,

Kirk

dmulyalin:
@ktbyers thank you for looking at it.

Let me elaborate on the items you raised.

1. Agree - that is why more testing required, not even sure that it will work for all platforms, quiet sure it will need more edge cases handling. However, to my experience it proved working for major platforms such as Cisco IOS, XR, NXOS, Juniper and Linux.
2. Agree - multiline handling using "\n" is rather a by-product and not the core feature of this method, main functionality ps mode brings to the table is capability to gather (usually very long) output from devices where normal send_command and timing methods tend to fail from time to time, in other words, main feature is increased reliability
3. Agree - plan was to introduce promptless mode as additional, experimental method, it is not positioned to replace or modify any of existing methods of base connection class

Hope above helped to clarify some aspects of this pull request, what do you think?

ktbyers:
Thanks for contributing this, but I definitely am against adding this. Consequently, I am going to go ahead and close it.

Regards, Kirk

dmulyalin:
No worries, thank you @ktbyers for looking at it, appreciate your time and efforts.

For those who might be interested to use promptless mode, it was tested, implemented as a Netmiko based plugin for Nornir framework and distributed as part of Nornir-Salt package - [netmiko_send_commands_ps task](https://nornir-salt.readthedocs.io/en/latest/Tasks/netmiko_send_command_ps.html#netmiko-send-command-ps), in addition SaltStack Nornir based proxy minion can invoke promptless mode using `use_ps=True` argument with [nr.cli](https://salt-nornir.readthedocs.io/en/latest/Nornir%20Execution%20Module.html#nr-cli) execution module  function.

All the best to you all.

PR #2116, opened 2021-01-17
Initial support for Brocade Fabric OS

DanielWeeber:
Only tested with basic commands, but should be fine


ktbyers:
Updated in this PR:

https://github.com/ktbyers/netmiko/pull/2134

PR #2134, opened 2021-02-02
Brocade fabric os support

ktbyers:


ktbyers:
@DanielWeeber Let me know if you see any issues here.

I made a few updates, but I think this is consistent with what you had earlier as far as behavior.

Changes I made:
1. Updated naming to be `brocade_fos` (for fabric os).
2. Changed behavior for both config-mode and enable-mode as it looks like Fabric OS has neither of these. This basically made it consistent with what Netmiko is expecting for a device that doesn't have these.

ktbyers:
@DanielWeeber Any possibility you can run the Netmiko test suite against a test device?

Testing process is here:

PR #2136, opened 2021-02-03
Allied telesis alliedware plus driver

ktbyers:


ktbyers:
FYI @openmymai I inherited from the CiscoBaseConnection class which allowed me to remove some of your code. I also removed the telnet and file transfer classes at least until we can more thoroughly test them (i.e. run the standard Netmiko test suite against them).

If you want to re-add them, we can. We just need to run the tests against them.

Regards, Kirk

ktbyers:
I also updated the naming to be more consistent with the Netmiko naming.

ktbyers:
@openmymai Let me know if you get a quick chance to sanity test this updated driver (to make sure I didn't break anything with my changes).

PR #2140, opened 2021-02-05
Add type hints and fix no_auth channel logic for HP drivers

wrgeorge1983:
~needs~ needed some bugfixing because of this:
```
hp\hp_procurve.py:16: error: "HPProcurveSSHChannel" has no attribute "use_keys"
hp\hp_procurve.py:16: error: "HPProcurveSSHChannel" has no attribute "password"
```
```python
class HPProcurveSSHChannel(SSHChannel):
    def _build_ssh_client(self, no_auth=True) -> paramiko.SSHClient:
        """Allow passwordless authentication for HP devices being provisioned."""
        if not self.use_keys and not self.password:
            return super()._build_ssh_client(no_auth=no_auth)
        else:
            return super()._build_ssh_client(no_auth=False)

```

~that'll come in a following commit~
fixed

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -617,12 +617,16 @@ class BaseConnection(object):
             ChannelClass = SSHChannel
         else:
             ChannelClass = ssh_channel_class
+        assert issubclass(ChannelClass, SSHChannel)
+
         self.channel = ChannelClass(
             ssh_params,
             device_type=self.device_type,
             ssh_hostkey_args=self.ssh_hostkey_args,
             encoding=self.encoding,
             session_log=self.session_log,
+            use_keys=self.use_keys,  # needed for making decision to use no_auth
+            password=self.password,  # needed for making decision to use no_auth
         )
         self.channel.establish_connection()
         self.special_login_handler()
--- a/netmiko/hp/hp_comware.py
+++ b/netmiko/hp/hp_comware.py
@@ -3,14 +3,14 @@ from netmiko.cisco_base_connection import CiscoSSHConnection
 
 
 class HPComwareBase(CiscoSSHConnection):
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
         # Comware doesn't have a way to set terminal width which breaks cmd_verify
         global_cmd_verify = kwargs.get("global_cmd_verify")
         if global_cmd_verify is None:
             kwargs["global_cmd_verify"] = False
         return super().__init__(**kwargs)
 
-    def session_preparation(self):
+    def session_preparation(self) -> None:
         """
         Prepare the session after the connection has been established.
         Extra time to read HP banners.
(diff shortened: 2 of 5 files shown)


netmiko/base_connection.py
```diff
@@ -617,12 +617,16 @@ def _open_ssh(self, ssh_channel_class: Optional[Type[SSHChannel]]) -> None:
             ChannelClass = SSHChannel
         else:
             ChannelClass = ssh_channel_class
+        assert issubclass(ChannelClass, SSHChannel)
```
wrgeorge1983, line 620: I *BELIEVE* this is safe and correct.  if it's not would need to either make it true, or do something else to make sure it's okay to provide `use_keys` and `password` parameters.

netmiko/hp/hp_comware.py
```diff
@@ -34,21 +34,25 @@ def session_preparation(self):
...
+        self, config_command: str = "system-view", pattern: str = "", re_flags: int = 0
+    ) -> str:
         """Enter configuration mode."""
-        return super().config_mode(config_command=config_command)
+        return super().config_mode(
+            config_command=config_command, pattern=pattern, re_flags=re_flags
+        )
 
-    def exit_config_mode(self, exit_config="return", pattern=r">"):
+    def exit_config_mode(self, exit_config="return", pattern=r">") -> str:
```
ktbyers, line 45: Your missing the type hints on the arguments i.e. `exit_config: str = "return"` et cetera.

netmiko/hp/hp_comware.py
```diff
@@ -34,21 +34,25 @@ def session_preparation(self):
...
         """Exit config mode."""
         return super().exit_config_mode(exit_config=exit_config, pattern=pattern)
 
-    def check_config_mode(self, check_string="]"):
+    def check_config_mode(self, check_string="]", pattern: str = "]") -> bool:
```
ktbyers, line 49: Missing type hints on check_string argument.

netmiko/hp/hp_comware.py
```diff
@@ -34,21 +34,25 @@ def session_preparation(self):
...
         """Check whether device is in configuration mode. Return a boolean."""
-        return super().check_config_mode(check_string=check_string)
+        return super().check_config_mode(check_string=check_string, pattern=pattern)
 
     def set_base_prompt(
         self, pri_prompt_terminator=">", alt_prompt_terminator="]", delay_factor=1
-    ):
+    ) -> str:
```
ktbyers, line 55: Should add type hints on the arguments as well.

ktbyers (changes requested):
I think I would probably focus on submitting the HP Compware PR and have that be stand-alone. That will make it a lot easier for me to review/merge.

The biggest missing item is type hints on the arguments for all of the functions.

ktbyers:
Okay, I realized after the fact that I had already done both hp_procurve and hp_comware and just hadn't pushed them up here. That is totally my bad...so sorry about that.

Looking at the mypy output the next drivers that need done are the:

```
fortinet_ssh
flexvnf_ssh
f5_tmsh_ssh
```

So if you want to work on those, let me know.

So I am going to go ahead and close this PR, and then we can start on the "f" through "a" drivers.

Once again I apologize about the wasted effort here.


Regards, Kirk

PR #2155, opened 2021-02-12
mypy coverage for f5, flexvnf, and fortinet drivers

wrgeorge1983:


netmiko/flexvnf/flexvnf_ssh.py
```diff
@@ -41,27 +42,29 @@ def enter_cli_mode(self):
...
+        return ""
 
-    def exit_enable_mode(self, *args, **kwargs):
+    def exit_enable_mode(self, *args, **kwargs) -> str:
         """No enable mode on flexvnf."""
-        pass
+        return ""
 
-    def check_config_mode(self, check_string="]"):
+    def check_config_mode(self, check_string: str = "]", *args, **kwargs) -> bool:
```
wrgeorge1983, line 57: I don't like relying on `*args, **kwargs`.  🤷 

since original signature doesn't match base class (triggering mypy error), existing code would outright break if users tried to follow signature from base class.  This silently swallows extra parameters.  Alternatively we could pass the extras (`pattern` in this case) on to the Super.  Not sure which makes more sense.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -39,8 +39,12 @@ def session_preparation(self):
...
         self.clear_buffer()
 
-    def disable_paging(self, delay_factor=1, **kwargs):
-        """Disable paging is only available with specific roles so it may fail."""
+    def disable_paging(
+        self, command: str = None, delay_factor: float = 1, *args, **kwargs
+    ) -> str:
+        """Disable paging is only available with specific roles so it may fail.
+        Does not use `command`, `cmd_verify`, or `pattern` parameters"""
+
```
wrgeorge1983, line 47: similar concern as above.  since users might be depending on `command` being first positional parameter I thought might be worth special handling here.  IMO it's probably worth a `logger.warning()` or something.

netmiko/flexvnf/flexvnf_ssh.py
ktbyers: @wrgeorge1983 The difficult on the other side is `if I ever add an argument to the base class, then I have 50+ places I  also need to update it in the child classes`.

I am fine with either pattern though (i.e. either the below) or fully defining all of the arguments.

`def exit_enable_mode(self, *args: Any, **kwargs: Any) -> str:`

I am pretty sure the above will pass mypy.

Note, if people are passing any extra arguments here there code really is broken and we are doing a major release (i.e. netmiko3 to netmiko4) so they should be testing the new code before they upgrade.

ktbyers: Or in the case of `check_config_mode`, the form should be one of the following two:

```
def check_config_mode(self, check_string: str = "]", *args: Any, **kwargs: Any) -> bool:
```

Or

```
def check_config_mode(self, check_string: str = "*>", pattern: str = "") -> bool:
```

netmiko/fortinet/fortinet_ssh.py
ktbyers: `command` would have been broken as the first positional argument (in the old code) for the Fortinet driver...so no one could have actually used command as the first positional argument for that driver.

ktbyers:
I incorporated your updates here (with some minor changes):

https://github.com/ktbyers/netmiko/pull/2159

PR #2160, opened 2021-02-19
Fix yamaha enter issue on exit enable mode

ktbyers:
.

### DIFF
--- a/netmiko/yamaha/yamaha.py
+++ b/netmiko/yamaha/yamaha.py
@@ -29,7 +29,7 @@ class YamahaBase(BaseConnection):
             output = self.read_channel()
             if "(Y/N)" in output:
                 self.write_channel(f"N{self.default_enter}")
-            output += self.read_until_prompt()
+                output += self.read_until_prompt()
             if self.check_enable_mode():
                 raise ValueError("Failed to exit enable mode.")
         return output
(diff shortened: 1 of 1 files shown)


ktbyers:
@Tachashi FYI

ktbyers:
@Tachashi Just let me know if this is still broken for some reason...I think it should work based on the information you provided in the other PR.

Tachashi:
@ktbyers Thanks! I confirmed that the enter issue has been resolved. But the code below stops with error:

https://github.com/ktbyers/netmiko/blob/2d716523f4d7481154c5c30c00a69972bbac38f6/netmiko/yamaha/yamaha.py#L31

```
  File "send_config_telnet.py", line 27, in <module>
    output = net_connect.exit_enable_mode()
  File "/home/centos/venv_netmiko3.3dev2/netmiko/netmiko/yamaha/yamaha.py", line 31, in exit_enable_mode
    self.write_channel(f"N{self.default_enter}")
AttributeError: 'YamahaTelnet' object has no attribute 'default_enter'
```

So, I think it probably needs to be changed like (1) or (2).

(1) Without default_enter

```python
            if "(Y/N)" in output:
                self.write_channel("N")
```

(1) Result

```
---------------exit enable---------------
exit 
Save new configuration ? (Y/N)N
yamaha1> 
```

(2) With RETURN

```python
            if "(Y/N)" in output:
                self.write_channel(f"N{self.RETURN}")
```

(2) Result

```
---------------exit enable---------------
exit 
Save new configuration ? (Y/N)N
yamaha1> 
yamaha1> 
```

ktbyers:
@Tachashi Okay, sorry about that...that was just my screw-up.

Fix is here:

https://github.com/ktbyers/netmiko/pull/2163

PR #2166, opened 2021-02-25
make send_config_set available for generic devices

Trasmontinho:
Hi @ktbyers , 

I would like to propose a change on terminal_server. Here is the thing : 
I'm making some changes on Avocent Equipments (console stuff), I don't think that a plugin for it should be necessary and terminal_server is enough to do it. I've been able to pass send_commands with the terminal_server but send_config_set is failing the exit_config_mode. I basically would like to add methods that will do nothing (as for the f5 driver) to allow the user to send config set.

What do you think about this change ?


Trasmontinho:
I'm ok with that and can propose the pull request. 

Which Name would be good to you ? I've seen that generic_ssh is already defined as a platform 

Let me know, I'll prepare it.

ktbyers:
@Trasmontinho Are you still interested in working on this?

Trasmontinho:
Hi @ktbyers,

I’ve been quite busy with other subjects. I’ll may reach you next week to define a male for the platform to take this specific case. If you are ok with that.

Sam

Trasmontinho:
Hi @ktbyers, 

If you're ok with it, I propose you "generic_device" as platform name. Is it ok for you ?

Thanks,

Regards, 
Sam

ktbyers:
I am going to close this as it doesn't seem like we are actively working on it (my fault as it took me so long to respond).

If anyone still wants to work on this, just re-open or create a new pull request.

PR #2167, opened 2021-02-25
mypy updates for extreme, endace, enterasys, eltex, coriant, cloudgenix

wrgeorge1983:
I believe this reflects guidance from last PR.  Of course let me know if not.

### DIFF
--- a/netmiko/cloudgenix/cloudgenix_ion.py
+++ b/netmiko/cloudgenix/cloudgenix_ion.py
@@ -44,7 +44,7 @@ class CloudGenixIonSSH(CiscoSSHConnection):
 
     def save_config(self, *args: Any, **kwargs: Any) -> str:
         """No save method on ION SSH"""
-        pass
+        return ""
 
     def send_config_set(
         self,
--- a/netmiko/cloudgenix/cloudgenix_ion.py
+++ b/netmiko/cloudgenix/cloudgenix_ion.py
@@ -1,49 +1,58 @@
+import io
+from typing import Any, Union, Iterable
+
 from netmiko.cisco_base_connection import CiscoSSHConnection
 
 
 class CloudGenixIonSSH(CiscoSSHConnection):
-    def establish_connection(self):
-        super().establish_connection(width=100, height=1000)
+    def establish_connection(self) -> None:
+        self.channel.establish_connection(width=100, height=1000)
 
-    def session_preparation(self, *args, **kwargs):
+    def session_preparation(self, *args: Any, **kwargs: Any) -> None:
         """Prepare the session after the connection has been established."""
         self.ansi_escape_codes = True
         self._test_channel_read()
         self.write_channel(self.RETURN)
         self.set_base_prompt(delay_factor=5)
 
-    def disable_paging(self, *args, **kwargs):
+    def disable_paging(self, *args: Any, **kwargs: Any) -> str:
         """Cloud Genix ION sets terminal height in establish_connection"""
         return ""
 
-    def find_prompt(self, delay_factor=1):
+    def find_prompt(self, delay_factor: float = 1.0) -> str:
         prompt = super().find_prompt(delay_factor=delay_factor)
         prompt = self.strip_backspaces(prompt).strip()
         return prompt
 
-    def strip_command(self, command_string, output):
+    def strip_command(self, command_string: Any, output: Any) -> str:
         output = super().strip_command(command_string, output)
         # command_string gets repainted potentially multiple times (grab everything after last one)
         output = output.split(command_string)[-1]
         return output
 
-    def check_config_mode(self):
+    def check_config_mode(self, *args: Any, **kwargs: Any) -> bool:
         """Devices do not have a config mode."""
         return False
 
-    def config_mode(self):
+    def config_mode(self, *args: Any, **kwargs: Any) -> str:
         """Devices do not have a config mode."""
         return ""
 
-    def exit_config_mode(self):
+    def exit_config_mode(self, *args: Any, **kwargs: Any) -> str:
         """Devices do not have a config mode."""
         return ""
 
-    def save_config(self, *args, **kwargs):
+    def save_config(self, *args: Any, **kwargs: Any) -> str:
         """No save method on ION SSH"""
         pass
 
-    def send_config_set(self, config_commands=None, exit_config_mode=False, **kwargs):
+    def send_config_set(
+        self,
+        config_commands: Union[str, Iterable[str], io.TextIOWrapper, None] = None,
+        exit_config_mode: bool = False,
+        *args: Any,
+        **kwargs: Any
+    ) -> str:
         return super().send_config_set(
             config_commands=config_commands, exit_config_mode=exit_config_mode, **kwargs
         )
--- a/netmiko/coriant/coriant_ssh.py
+++ b/netmiko/coriant/coriant_ssh.py
@@ -1,35 +1,40 @@
+from typing import Any
+
 from netmiko.cisco_base_connection import CiscoSSHConnection
 
 
 class CoriantSSH(CiscoSSHConnection):
-    def session_preparation(self):
+    def session_preparation(self) -> None:
         self._test_channel_read()
         self.set_base_prompt()
 
-    def check_enable_mode(self, *args, **kwargs):
+    def check_enable_mode(self, *args: Any, **kwargs: Any) -> bool:
         raise AttributeError("Coriant devices do not support enable mode!")
 
-    def enable(self, *args, **kwargs):
+    def enable(self, *args: Any, **kwargs: Any) -> str:
         raise AttributeError("Coriant devices do not support enable mode!")
 
-    def exit_enable_mode(self, *args, **kwargs):
+    def exit_enable_mode(self, *args: Any, **kwargs: Any) -> str:
         raise AttributeError("Coriant devices do not support enable mode!")
 
-    def check_config_mode(self):
+    def check_config_mode(self, *args: Any, **kwargs: Any) -> bool:
         """Coriant devices do not have a config mode."""
         return False
 
-    def config_mode(self):
+    def config_mode(self, *args: Any, **kwargs: Any) -> str:
         """Coriant devices do not have a config mode."""
         return ""
 
-    def exit_config_mode(self):
+    def exit_config_mode(self, *args: Any, **kwargs: Any) -> str:
         """Coriant devices do not have a config mode."""
         return ""
 
     def set_base_prompt(
-        self, pri_prompt_terminator=":", alt_prompt_terminator=">", delay_factor=2
-    ):
+        self,
+        pri_prompt_terminator: str = ":",
+        alt_prompt_terminator: str = ">",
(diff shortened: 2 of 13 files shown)


netmiko/cloudgenix/cloudgenix_ion.py
```diff
@@ -1,49 +1,58 @@
...
+from typing import Any, Union, Iterable
+
 from netmiko.cisco_base_connection import CiscoSSHConnection
 
 
 class CloudGenixIonSSH(CiscoSSHConnection):
-    def establish_connection(self):
-        super().establish_connection(width=100, height=1000)
+    def establish_connection(self) -> None:
+        self.channel.establish_connection(width=100, height=1000)
```
wrgeorge1983, line 9: So this is "fixed" but I don't think this code will ever be called.  If we need to ensure width/height are set to 100, what's the right hook to use?  Do we need a custom channel?

ktbyers: Okay, thanks...yeah, I will have to figure out how to handle it. I am not sure if it is even needed any longer. 

I have created a Netmiko issue to track the problem.

netmiko/cloudgenix/cloudgenix_ion.py
```diff
@@ -1,49 +1,58 @@
...
         return ""
 
-    def find_prompt(self, delay_factor=1):
+    def find_prompt(self, delay_factor: float = 1.0) -> str:
         prompt = super().find_prompt(delay_factor=delay_factor)
         prompt = self.strip_backspaces(prompt).strip()
         return prompt
 
-    def strip_command(self, command_string, output):
+    def strip_command(self, command_string: Any, output: Any) -> str:
```
ktbyers, line 27: Should be:

```python
     def strip_command(self, command_string: str, output: str) -> str:
```

netmiko/cloudgenix/cloudgenix_ion.py
```diff
@@ -1,49 +1,58 @@
...
         output = super().strip_command(command_string, output)
         # command_string gets repainted potentially multiple times (grab everything after last one)
         output = output.split(command_string)[-1]
         return output
 
-    def check_config_mode(self):
+    def check_config_mode(self, *args: Any, **kwargs: Any) -> bool:
         """Devices do not have a config mode."""
         return False
```
ktbyers, line 35: We should switch this to `return True` that is the standard thing to do for devices that do not have a config mode.

netmiko/coriant/coriant_ssh.py
```diff
@@ -1,35 +1,40 @@
...
 
 class CoriantSSH(CiscoSSHConnection):
-    def session_preparation(self):
+    def session_preparation(self) -> None:
         self._test_channel_read()
         self.set_base_prompt()
 
-    def check_enable_mode(self, *args, **kwargs):
+    def check_enable_mode(self, *args: Any, **kwargs: Any) -> bool:
         raise AttributeError("Coriant devices do not support enable mode!")
```
ktbyers, line 12: Should switch this to `return True`

netmiko/coriant/coriant_ssh.py
```diff
@@ -1,35 +1,40 @@
...
 
-    def enable(self, *args, **kwargs):
+    def enable(self, *args: Any, **kwargs: Any) -> str:
         raise AttributeError("Coriant devices do not support enable mode!")
```
ktbyers, line 15: Should switch this to `return ""`

netmiko/coriant/coriant_ssh.py
```diff
@@ -1,35 +1,40 @@
...
 
-    def exit_enable_mode(self, *args, **kwargs):
+    def exit_enable_mode(self, *args: Any, **kwargs: Any) -> str:
         raise AttributeError("Coriant devices do not support enable mode!")
```
ktbyers, line 18: should switch this to `return ""`

netmiko/coriant/coriant_ssh.py
```diff
@@ -1,35 +1,40 @@
...
 
-    def check_config_mode(self):
+    def check_config_mode(self, *args: Any, **kwargs: Any) -> bool:
         """Coriant devices do not have a config mode."""
         return False
```
ktbyers, line 22: This should `return True`

netmiko/eltex/eltex_esr_ssh.py
```diff
@@ -16,19 +18,27 @@ def session_preparation(self):
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()
 
-    def config_mode(self, config_command="configure", pattern=r")#"):
+    def config_mode(
+        self,
+        config_command: str = "configure",
+        pattern: str = r")#",
```
ktbyers, line 24: We should escape this pattern so it should become:

`r"\)\#"`

netmiko/eltex/eltex_esr_ssh.py
```diff
@@ -16,19 +18,27 @@ def session_preparation(self):
...
+        *args: Any,
+        **kwargs: Any
+    ) -> str:
```
ktbyers, line 27: Instead of passing *args/**kwargs, I would pass this one additional argument: `re_flags: int = 0` and then use that in the super call i.e. add `re_flags=re_flags` as part of the super call.

netmiko/endace/endace_ssh.py
```diff
@@ -1,18 +1,36 @@
...
 
 class EndaceSSH(CiscoSSHConnection):
-    def disable_paging(self, command="no cli session paging enable", delay_factor=1):
+    def disable_paging(
+        self,
+        command: str = "no cli session paging enable",
+        delay_factor: float = 1.0,
+        *args: Any,
+        **kwargs: Any
+    ) -> str:
```
ktbyers, line 14: You should pass *args and **kwargs in the super call so that arguments are not dropped.

netmiko/endace/endace_ssh.py
```diff
@@ -1,18 +1,36 @@
...
-    def enable(self, cmd="enable", pattern="", re_flags=re.IGNORECASE):
+    def enable(
+        self, cmd: str = "enable", pattern: str = "", re_flags: int = re.IGNORECASE
+    ) -> str:
         return super().enable(cmd=cmd, pattern=pattern, re_flags=re_flags)
 
-    def check_config_mode(self, check_string="(config) #"):
+    def check_config_mode(
+        self, check_string: str = "(config) #", *args: Any, **kwargs: Any
+    ) -> bool:
```
ktbyers, line 24: Probably just add the pattern argument with the same default value as base_connection.py

```python
    def check_config_mode(self, check_string: str="(config) #", pattern: str = "") -> bool:
```

And then pass that in the `super()` call

netmiko/endace/endace_ssh.py
```diff
@@ -1,18 +1,36 @@
...
         return super().check_config_mode(check_string=check_string)
 
-    def config_mode(self, config_command="conf t", pattern=""):
+    def config_mode(
+        self,
+        config_command: str = "conf t",
+        pattern: str = "",
+        *args: Any,
+        **kwargs: Any
```
ktbyers, line 32: Probably just add the one additional argument `re_flags: int = 0` (with the standard default value instead of *args/**kwargs). This argument will not be used.

netmiko/extreme/extreme_exos.py
```diff
@@ -19,7 +21,7 @@ def session_preparation(self):
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()
 
-    def set_base_prompt(self, *args, **kwargs):
+    def set_base_prompt(self, *args, **kwargs) -> str:
```
ktbyers, line 24: Missing `: Any` on *args and **kwargs.

netmiko/extreme/extreme_exos.py
```diff
@@ -43,7 +45,7 @@ def set_base_prompt(self, *args, **kwargs):
         else:
             return self.base_prompt
 
-    def send_command(self, *args, **kwargs):
+    def send_command(self, *args, **kwargs) -> str:
```
ktbyers, line 48: Missing `Any` on *args and **kwargs.

netmiko/extreme/extreme_exos.py
```diff
@@ -53,21 +55,26 @@ def send_command(self, *args, **kwargs):
...
 
-    def config_mode(self, config_command=""):
+    def config_mode(self, config_command: str = "", *args: Any, **kwargs: Any) -> str:
         """No configuration mode on Extreme Exos."""
         return ""
 
-    def check_config_mode(self, check_string="#"):
+    def check_config_mode(
+        self, check_string: str = "#", *args: Any, **kwargs: Any
+    ) -> bool:
```
ktbyers, line 64: Probably just define it as:

```python
def check_config_mode(self, check_string: str = "#", pattern: str = "") -> bool:
```

And then add `pattern=pattern` in the super() call

netmiko/extreme/extreme_exos.py
```diff
@@ -53,21 +55,26 @@ def send_command(self, *args, **kwargs):
...
         """Checks whether in configuration mode. Returns a boolean."""
         return super().check_config_mode(check_string=check_string)
 
-    def exit_config_mode(self, exit_config=""):
+    def exit_config_mode(self, exit_config: str = "", *args: Any, **kwargs: Any) -> str:
```
ktbyers, line 68: Probably just definite it as:

```python
    def exit_config_mode(self, exit_config: str = "", pattern: str = "") -> str:
```

netmiko/extreme/extreme_exos.py
```diff
@@ -79,7 +86,7 @@ class ExtremeExosSSH(ExtremeExosBase):
 
 
 class ExtremeExosTelnet(ExtremeExosBase):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
```
ktbyers, line 89: Missing `: Any` for *args and **kwargs

netmiko/extreme/extreme_netiron.py
```diff
@@ -14,7 +19,7 @@ class ExtremeNetironSSH(ExtremeNetironBase):
 
 
 class ExtremeNetironTelnet(ExtremeNetironBase):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
```
ktbyers, line 22: Missing `: Any` for *args and **kwargs

ktbyers (commented):
I will try to provide some more comments/clarification tomorrow as I think some of this is potentially inconsistent with what I stated earlier.

ktbyers:
@wrgeorge1983 I tried to clarify the proper behaviors of this here:

https://github.com/ktbyers/netmiko/discussions/2188

I will have to retrofit some of the Netmiko4 code to match this.

Let me know if that doesn't make sense or is not consistent with previous statements.


Once again, thanks for your help.


Regards, Kirk

ktbyers:
Superseded by https://github.com/ktbyers/netmiko/pull/2189

Issue #2139, opened 2021-02-05
Cisco ASA connect failure when device is already in enable mode after login


mrkcmo:
Works in version 3.1.0 and breaks in version 3.3.0.

If the device is not in enable mode at login then it works correctly and goes into enable mode allowing the script to continue.

If the device drops to enable mode at login (aaa authorization exec LOCAL auto-enable) then we get the below traceback at login using 3.3.0 but not with 3.1.0.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/paramiko/channel.py", line 699, in recv
    out = self.in_buffer.read(nbytes, self.timeout)
  File "/usr/local/lib/python3.6/site-packages/paramiko/buffered_pipe.py", line 164, in read
    raise PipeTimeout()
paramiko.buffered_pipe.PipeTimeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/netmiko/base_connection.py", line 569, in _read_channel_expect
    new_data = self.remote_conn.recv(MAX_BUFFER)
  File "/usr/local/lib/python3.6/site-packages/paramiko/channel.py", line 701, in recv
    raise socket.timeout()
socket.timeout

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/netmiko/base_connection.py", line 1680, in enable
    pattern=pattern, re_flags=re_flags
  File "/usr/local/lib/python3.6/site-packages/netmiko/base_connection.py", line 669, in read_until_prompt_or_pattern
    return self._read_channel_expect(combined_pattern, re_flags=re_flags)
  File "/usr/local/lib/python3.6/site-packages/netmiko/base_connection.py", line 580, in _read_channel_expect
    "Timed-out reading channel, data not available."
netmiko.ssh_exception.NetmikoTimeoutException: Timed-out reading channel, data not available.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "vpn_stats.py", line 313, in <module>
    vpnstats()
  File "vpn_stats.py", line 308, in vpnstats
    vpn_sess_obj.updateVpnSessionStats()
  File "vpn_stats.py", line 235, in updateVpnSessionStats
    cmd_out = self.__ciscoSessions(fw[X].strip())
  File "vpn_stats.py", line 138, in __ciscoSessions
    secret=self.password)
[...]

PR #2182, opened 2021-03-09
Fix ASA enable() issue in session_preparation

ktbyers:
In cases where end-users were already in enable-mode, session_preparation would incorrectly believe you were not in enable mode (in other words check_enable_mode() would fail due to the base_prompt not being set).

This would generate an exception for the end-user.


ktbyers:
Hopefully fixes https://github.com/ktbyers/netmiko/issues/2139

PR #2207, opened 2021-03-29
ubuntu sshd integration test

alexcb:
- this creates an integration test that runs sshd in a dockerized ubuntu
container, and validates netmiko can connect using a username/password
and can issue a cat command.
- this makes use of earthly to run the test, this allows it to work
under github actions (or any other CI), as well as supports running it
locally via "earthly +integration-test"

Signed-off-by: Alex Couture-Beil <alex@earthly.dev>


alexcb:
Hi @ktbyers , any chance you've been able to look at this?

PR #2208, opened 2021-04-01
Add run_ttp method to base connection

dmulyalin:
# Why?

To ease structured data extraction from network devices show commands output.

# How?

TTP templates can contain inputs tags, these tags can have various parameters defined, these parameters can be used by parent program to build a logic around adding data to TTP parser object.

# What?

Using new `run_ttp` method we can instruct Netmiko to collect output from devices, place it in appropriate inputs and parse it with TTP. For instance consider this template:

```
<input name="ntp_cfg" load="yaml">
method: send_command
kwargs: {}
commands:
  - show run | inc ntp
</input>
    
<input name="interfaces_cfg">
commands = [
    "show run | sec interface"
]
</input>

<group name="misc" input="ntp_cfg">
ntp server {{ ntp_servers }}
</group>

<group name="interfaces" input="interfaces_cfg">
interface {{ interface }}
 description {{ description | re(".+") }}
 encapsulation dot1Q {{ dot1q }}
 ip address {{ ip }} {{ mask }}
</group>
```

Sample code:

```
import pprint

lab = {
    "device_type": "cisco_ios",
    "host": "1.2.3.4",
    "username": "cisco",
    "password": "cisco",
}

connection = ConnectHandler(**lab)

res = connection.run_ttp(template)

pprint.pprint(res)
```

# Main features

- Can source templates from strings, files or TTP templates collection
- Any Netmiko connection object method can be called to collect devices output together with required arguments
- Multiple commands can be collected per input
- Additional arguments can be supplied to TTP object instantiation and TTP result method
- Allows to provide data to groups for certain commands output only, lowering probability of false matches

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -31,6 +31,7 @@ from netmiko.utilities import (
     get_structured_data,
     get_structured_data_genie,
     get_structured_data_ttp,
+    run_ttp_template,
     select_cmd_verify,
 )
 from netmiko.utilities import m_exec_time  # noqa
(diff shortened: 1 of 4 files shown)


ktbyers:
Sounds good and makes sense.

It looks like you have a few linting/black issues, can you fix these?


Regards, Kirk

dmulyalin:
@ktbyers black and pylama looks good now

PR #2218, opened 2021-04-13
Fix XR exit_config_mode issue

ktbyers:


ktbyers:
@Marshall700 FYI, I made some minor updates to your original PR.

ktbyers:
@Marshall700 Just let me know if you see any issues with the final change, but it looks like it should work properly.

PR #2231, opened 2021-04-22
Supermicro Driver

ktbyers:


ktbyers:
@sandyw777 @puffc Sorry for the very long delay on this.

I switched the device_type to `supermicro_smis` (was `supermicro_nos` previously). Looking online the name of the operating system is SMIS--is this correct?

Just let me know if this is wrong and I can fix it.

I am going to go ahead and merge this as it looks correct and we can correct the naming issue after the fact (if needed).

sandyw777:
SMIS is good, thank you!

PR #2233, opened 2021-04-23
Fix hp_comware.py support h3c S5820V2-52QF config_set match error.

ColinSix:
support S5820V2-52QF version send_config_set match prompt error.

change:
pattern = f"(?:{re.escape(self.base_prompt)}|#)"
to:
pattern = f"(?:{re.escape(self.base_prompt)}|\])"

### DIFF
--- a/netmiko/hp/hp_comware.py
+++ b/netmiko/hp/hp_comware.py
@@ -91,7 +91,70 @@ class HPComwareBase(CiscoSSHConnection):
 
 
 class HPComwareSSH(HPComwareBase):
-    pass
+    def send_config_set(
+            self,
+            config_commands=None,
+            exit_config_mode=True,
+            delay_factor=1,
+            max_loops=150,
+            strip_prompt=False,
+            strip_command=False,
+            config_mode_command=None,
+            cmd_verify=True,
+            enter_config_mode=True,
+    ):
+        delay_factor = self.select_delay_factor(delay_factor)
+        if config_commands is None:
+            return ""
+        elif isinstance(config_commands, str):
+            config_commands = (config_commands,)
+
+        if not hasattr(config_commands, "__iter__"):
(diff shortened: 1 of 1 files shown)


ktbyers:
What is the underlying issue you are trying to fix? Can you show me an example exception/error.

Please include the full stack trace.

ColinSix:
Test model:

```
H3C Comware Software, Version 7.1.045, Release 2432P03
S5820V2-52QF with 2 Processors
BOARD TYPE:         S5820V2-52QF
```

Test code:
```
def test_confset_h3c():
    netcfg = {
            'host': host,
            'username': username,
            'password': password,
            'port': port,
            'device_type': 'hp_comware'
        }

    command = ('interface Ten-GigabitEthernet 1/0/1', 'dis this')

    with ConnectHandler(**netcfg) as net_connect:
        output = `net_connect.send_config_set(command)`
        return output
```

If it should be output according to the normal situation：
```
#
interface Ten-GigabitEthernet1/0/1
 port link-mode bridge
 port access vlan 10
 stp edged-port
#
return
```
But due to the wrong match in "base_connection.py"：`pattern = f"(?:{re.escape(self.base_prompt)}|#)"` 
Resulting in incomplete output, like this:
```
#
interface Ten-GigabitEthernet1/0/1
 port link-mode bridge
 port access vlan 10

return
```
The reason is because the output contains `#` special characters, which cleverly matches `pattern = f"(?:{re.escape(self.base_prompt)}|#)"`

So I will rewrite this def `HPComwareSSH.send_config_set` in hp_comware.py, just replace `pattern = f"(?:{re.escape(self.base_prompt)}|#)"` to `pattern = f"(?:{re.escape(self.base_prompt)}|\])"`

ktbyers:
Fix superseded by:

https://github.com/ktbyers/netmiko/pull/2245

PR #2243, opened 2021-04-28
Updated Viptela driver

ktbyers:


ktbyers:
I switched to an explicit `commit()` method and made `save_config()` yield a NotImplementedError. This is more consistent with what Netmiko expects for devices that support `commit()`.

Additionally, I converted `send_config_set()` to not exit config mode. So that `commit()` could be executed after send_config_set(). So for configuration changes the user should do the following:

```
net_connect.send_config_set(config_list)
net_connect.commit()
net_connect.exit_config_mode()
```

Additionally, I added to `exit_config_mode()` an explicit send of `no` if this message is encountered:

```
Uncommitted changes found, commit them? [yes/no/CANCEL]
```

In other words, it is the user's responsibility to `commit()` prior to executing config_mode.

ktbyers:
@benmcbenben

I made the changes referenced above. Just let me know if you see any issues with them. Most of these items are to make the drivers behavior more consistent with standard Netmiko behavior.

Thanks for the contribution.

benmcbenben:
I see no issues here at all, thanks for merging :)

PR #2245, opened 2021-04-29
Make the alternate terminator a variable in send_config_set

ktbyers:
Also fix the HP comware issue that the wrong terminator character is causing.


ktbyers:
@ColinSix FYI, I put in a different fix for this HP comware issue here. Just let me know if it doesn't work for some reason.


Regards, Kirk

ColinSix:
Good, has been checked and passed!

PR #2247, opened 2021-04-30
Allow sending command to WLC with 'y' character

anirudhkamath:
On Cisco wireless controllers, a command like `show interface summary` may ask a prompt for showing more entries, such as:

```
(wlc) >show interface summary

 Number of Interfaces.......................... 17
Interface Name                   Port Vlan Id  IP Address      Type    Ap Mgr Guest
-------------------------------- ---- -------- --------------- ------- ------ -----
test       LAG  xxx      test-ip-addr  Dynamic No     No
test       LAG  xxx      test-ip-addr  Dynamic No     No
test       LAG  xxx      test-ip-addr  Dynamic No     No
...
test            LAG  xxx      test-ip-addr  Static  Yes    No
test            LAG  xxx      test-ip-addr  Static  No     No
test            -    untagged test-ip-addr  Static  No     No
Would you like to display the next 15 entries? (y/n) y
```

This prompt comes even after paging is disabled.
In such cases, `send_command` does not do the job as it does not get back the WLC prompt back. Hence, a method for this in the `CiscoWlcSSH` class would be needed which sends the `y` character to interact with the WLC as required.

### DIFF
--- a/netmiko/cisco/cisco_wlc_ssh.py
+++ b/netmiko/cisco/cisco_wlc_ssh.py
@@ -97,6 +97,23 @@ class CiscoWlcSSH(BaseConnection):
             output = self.strip_prompt(output)
         return output
 
+    def send_command_w_yes(self, *args, **kwargs):
+        """
+        For 'show interface summary' Cisco WLC adds a
+        'Would you like to display the next 15 entries?' message
+        Even though pagination is disabled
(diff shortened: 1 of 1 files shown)


ktbyers:
The WLC CLI is a mess... :-(

PR #2255, opened 2021-05-04
Strip y/n prompt line in send_command_w_yes for Cisco WLC

anirudhkamath:
Related to PR #2247 

- The method `send_command_w_yes` would allow interacting with a command such as `show interface summary` appropriately, but the output would contain the "Would you like to display the next 15 entries? (y/n)" line in the output. This change strips that prompt line from the output so that it does not show up in the returned output.
- This change also allows for positional arguments and keyword arguments to be applied to the second call of the `send_command_timing` method in `send_command_w_yes`.


anirudhkamath:
Will update this once #2523 and #2524 are addressed and resolved

PR #2265, opened 2021-05-07
add self.current_prompt attribute

Shrif88:
This is an improvement suggestion that I currently use in my derived class but I realized that this attribute is better off added as a new instance attribute "self.current_prompt" to the BaseConnection class. This attribute gets updated every time the find_prompt() is called. Sometimes we need to know the current prompt before, as well as without, invoking the find_prompt() when checking the current prompt. This attribute can be very handy when trying to print to stdout the current prompt without invoking the find_prompt().

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -312,6 +312,7 @@ class BaseConnection(object):
             self.global_delay_factor = 0.1
 
         # set in set_base_prompt method
+        self.current_prompt = ""
         self.base_prompt = ""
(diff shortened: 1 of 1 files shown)


ktbyers:
Probably better to track that in your particular program than in the object (or if you want it in your objects to bind it yourself to the object in your particular progeram). 

The main issue/concern I have is that `netmiko_object.current_prompt` would potentially be out of date/invalid at any point in time after you execute find_prompt().

Issue #2073, opened 2020-12-14
loading ntc templates installed via pip fails with "IsADirectoryError"


991jo:
Hello

when I am using textfsm installed via pip I get this error:

```
Traceback (most recent call last):
  File "/home/jo/foo/netmiko-bug/test.py", line 4, in <module>
    print(router.send_command("show int desc", use_textfsm=True))
  File "/home/jo/foo/netmiko-bug/lib/python3.9/site-packages/netmiko/utilities.py", line 429, in wrapper_decorator
    return func(self, *args, **kwargs)
  File "/home/jo/foo/netmiko-bug/lib/python3.9/site-packages/netmiko/base_connection.py", line 1538, in send_command
    structured_output = get_structured_data(
  File "/home/jo/foo/netmiko-bug/lib/python3.9/site-packages/netmiko/utilities.py", line 340, in get_structured_data
    template_dir = get_template_dir()
  File "/home/jo/foo/netmiko-bug/lib/python3.9/site-packages/netmiko/utilities.py", line 277, in get_template_dir
    with importresources_path(
  File "/usr/lib/foo/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/usr/lib/python3.9/importlib/resources.py", line 175, in _path_from_reader
    opener_reader = reader.open_resource(norm_resource)
  File "<frozen importlib._bootstrap_external>", line 995, in open_resource
IsADirectoryError: [Errno 21] Is a directory: '/home/jo/foo/netmiko-bug/lib/python3.9/site-packages/ntc_templates/templates'
```

Environment is:
Python 3.9.0
netmiko 3.3.2
ntc-templates 1.6.0
textfsm 1.1.0
venv installed via
```
python3 -m venv .
source bin/activate
pip install netmiko ntc_templates
```

The templates and the index file are in `/home/jo/foo/netmiko-bug/lib/python3.9/site-packages/ntc_templates/templates` as expected. Setting the `NET_TEXTFSM` environment variable to this path works as expected.

the script I run is this:

```
from netmiko import Netmiko

router = Netmiko(host="router.example.com", username="admin", password="somepass", secret="somesecret", device_type="cisco_xr")
print(router.send_command("show int desc", use_textfsm=True))
```

ktbyers:
Are you still seeing this issue?

991jo:
I will try find some time to try it again over the weekend.

ktbyers:
[...]

PR #2274, opened 2021-05-13
Fix path for pip installed ntc_templates on 3.9+

jgsawyers:
Workaround for the change in behavior of importlib.resources.path on Python 3.9, with minimal modifications. (Also 3.8, though it apparently still worked there). Since the docs say it will only return a file within a package, and no directories, changed to lookup the parse.py file in the root of the package. Gets the parent dir of that file and appends the `templates` directory. In testing, fixes the errors seen on #2048 and #2073 on my local systems.

### DIFF
--- a/netmiko/utilities.py
+++ b/netmiko/utilities.py
@@ -279,10 +279,10 @@ Alternatively, `pip install ntc-templates` (if using ntc-templates).
         # Try 'pip installed' ntc-templates
         try:
             with importresources_path(
-                package="ntc_templates", resource="templates"
+                package="ntc_templates", resource="parse.py"
             ) as posix_path:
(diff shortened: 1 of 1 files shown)


ktbyers:
@jgsawyers Thanks for this contribution :-)

vvevans:
Hey everyone, I'm running into this original issue still, after having upgraded netmiko and ntc-templates. I was looking over the "utilities.py" changes implemented by @jgsawyers and noticed those new lines were not present in my "utilities.py" file after the upgrade of netmiko. I added those lines and ran into a different issue ("package="ntc_templates", resource="parse.py" - invalid syntax error).

I'm new to python, so I don't really know what I'm doing, but I'm trying to work with TextFSM in a script and I was running into the "IsADirectoryError" issue that started this thread.

PR #2283, opened 2021-05-27
fixing cleanup doesn't close paramiko session when log in as root

l3-ojisan:
I've found that LinuxSSH doesn't close paramiko session if user is set to root.
Currently, cleanup() checks the prompt and if it detects the user is root then, "exit" is sent.
And the following "exit" fails, but since disconnect() catches Exception and does nothing, paramiko session is left open.

```python
def cleanup(self, command="exit"):
        """Gracefully exit the SSH session."""
        try:
            # The pattern="" forces use of send_command_timing
            if self.check_config_mode(pattern=""):
                self.exit_config_mode()
        except Exception:
            pass
        # Always try to send final 'exit' (command)
        self._session_log_fin = True
        self.write_channel(command + self.RETURN)
```

I updated cleanup() to check if the username is root or not, and I confirmed it worked in my local environment.
I suspect that this also affects F5LinuxSSH since it just inherits LinuxSSH and doesn't override anything.

I'm pretty new to contributing, so I'm missing something, please kindly let me know.

Thanks in advance


ktbyers:
I think this issue was fixed in this PR:

https://github.com/ktbyers/netmiko/pull/2846/files


Thanks for reporting it and sorry for the delayed response.

PR #2290, opened 2021-06-01
Create decorator for lock_channel

ktbyers:
Also add some of the glue for mypy and type hints.


ktbyers:
Also elimiated cmd_echo argument from send_command_timing (which I would think is generally not used).

PR #2307, opened 2021-06-14
Add telnet support for Cisco SG300

Joseph9883:
This is to add support for Telnet to the SG300 series of switches. 
this can be used with the Sx200 series of switches too, if you enable the telnet server via modifying the running config.


ktbyers:
Sorry I haven't got to this...I will try to look at this some time this week.

PR #2352, opened 2021-07-02
Add Escape Sequences for cursor forward and wrap around

tgoncuoglu:
Currently netmiko fails to connect to Nokia SRLinux devices with a "ValueError cannot find prompt" exception, because there there are two ANSI Escape codes missing from stripped codes list that are sent by device on connect: cursor forward and wrap around. This patch adds these two ANSI escape sequences.

Once added, netmiko successfully connects to Nokia SRLinux devices using the "linux" device_type.

In case you are not aware: SRLinux is the new OS that replaces SROS/TiMOS in a new line of Nokia routers.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -2054,8 +2054,10 @@ You can also look at the Netmiko session_log or debug log for more information.
         code_attrs_off = chr(27) + r"\[0m"
         code_reverse = chr(27) + r"\[7m"
         code_cursor_left = chr(27) + r"\[\d+D"
-        code_cursor_forward = chr(27) + r"\[\d+C"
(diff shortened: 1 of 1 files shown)


ktbyers:
We need to run python black on this as the linting is failing.

tgoncuoglu:
The reason black fails is because the file has a lot of long lines (>80). Many of those seem to be comments, but not all. None of those long lines is related to the few lines I added, so it would be failing before my changes. Having said that I could make changes to shorten those lines, but looking at the code I feel it might be better to increase line length limit to something like 120 or 150.

ktbyers:
https://github.com/ktbyers/netmiko/pull/2357

tgoncuoglu:
Thank you.

PR #2356, opened 2021-07-04
Invoke /bin/ls instead of ls to avoid custom alias or command path is…

jafingerhut:
…sues

### DIFF
--- a/netmiko/scp_handler.py
+++ b/netmiko/scp_handler.py
@@ -246,7 +246,7 @@ class BaseFileTransfer(object):
         """Check if the dest_file already exists on the file system (return boolean)."""
         if self.direction == "put":
             self.ssh_ctl_chan._enter_shell()
-            remote_cmd = f"ls {self.file_system}"
+            remote_cmd = f"/bin/ls {self.file_system}"
             remote_out = self.ssh_ctl_chan.send_command(
                 remote_cmd, expect_string=r"[\$#]"
             )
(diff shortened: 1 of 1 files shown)


jafingerhut:
Thanks for developing and publishing netmiko.  Very handy open source tool similar to Cisco's pyATS.

I was trying it out to connect to and control some Linux systems via SSH, and found that because my .bashrc defined some aliases for the 'ls' command that used additional command line options that I like to have during interactive CLI use, this was interfering with netmiko's ability to parse the output of these 'ls' commands.  As a temporary local workaround, I have issued an 'unalias ls' command just after connecting in my test scripts, but it seems like the changes proposed in this PR, or something similar, would help make netmiko more robust to other developers who might do similar customizations to their 'ls' command.

BenjaminHoegh:
Please make this customizable so you can change this after need, sometimes you will not have permission to use `/bin/ls` but can still use the `ls` command. This is currently giving me trouble and is quite annoying to have to change netmikos files directly

ktbyers:
@BenjaminHoegh Can you provide a bit more detail on why `/bin/ls` doesn't work and your use-case/situation that is failing?

BenjaminHoegh:
> @BenjaminHoegh Can you provide a bit more detail on why `/bin/ls` doesn't work and your use-case/situation that is failing?

Thanks for the quick response ❤️ 

Its the permissions settings on the target server I try there only allow me to use the `$ ls` command but not as `$ /bin/ls` this is someting I can't change as I do not have access to change the permissions on the server to my need.

ktbyers:
What netmiko device_type are you using? I am assuming `linux`?

BenjaminHoegh:
> What netmiko device_type are you using? I am assuming `linux`?

Linux yes :)

PR #2376, opened 2021-07-16
Type hints for f* drivers

verailina:
Added type hints for `f5` and `flexvnf` drivers.

### DIFF
--- a/netmiko/f5/f5_tmsh_ssh.py
+++ b/netmiko/f5/f5_tmsh_ssh.py
@@ -1,5 +1,3 @@
-from typing import Union, List, Any, Dict
-
 import time
 from netmiko.no_config import NoConfig
 from netmiko.base_connection import BaseConnection
--- a/netmiko/f5/f5_tmsh_ssh.py
+++ b/netmiko/f5/f5_tmsh_ssh.py
@@ -20,7 +20,7 @@ class F5TmshSSH(NoConfig, BaseConnection):
         )
         self.clear_buffer()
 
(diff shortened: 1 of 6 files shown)


netmiko/f5/f5_tmsh_ssh.py
```diff
@@ -28,12 +30,12 @@ def tmsh_mode(self, delay_factor=1):
         self.clear_buffer()
         return None
 
-    def exit_tmsh(self):
+    def exit_tmsh(self) -> Union[str, List[Any], Dict[str, Any]]:
```
ktbyers, line 33: Should always return a string (given the context/way send_command is used)...so just do a type check that output is a str and return a str

ktbyers:
@verailina A few comments inline. Also looks like you need to run `black .` on the code (from the base of the repository). I am using Python `black` for auto-code formatting. This is what is causing the CI-CD to fail.

Also, instead of doing assert_str, let's make the following methods in base_connection.py:

```python
def _send_command_str(self, *args, **kwargs) -> str:
```

And

```python
def _send_command_timing_str(self, *args, **kwargs) -> str:
```

Little wrapper methods for send_command and send_command_timing that always return a string. You can safely assume that any internal call of send_command and send_command_timing inside of Netmiko will return a str.

I put some more details about this in one of my comments.

Everything else looks good.

ktbyers:
Superseded by (one minor additional change):

https://github.com/ktbyers/netmiko/pull/2380

PR #2377, opened 2021-07-17
Type hints for h* drivers

verailina:
Added type hints for `hp` and `huawei` drivers.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1140,7 +1140,7 @@ Device settings: {self.device_type} {self.host}:{self.port}
             else:
                 return self.global_delay_factor
 
-    def special_login_handler(self, delay_factor: float = 1.0) -> None:
+    def special_login_handler(self, delay_factor: float = 1.0) -> Optional[str]:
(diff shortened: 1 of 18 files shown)


netmiko/base_connection.py
```diff
@@ -1140,7 +1140,7 @@ def select_delay_factor(self, delay_factor: float) -> float:
             else:
                 return self.global_delay_factor
 
-    def special_login_handler(self, delay_factor: float = 1.0) -> None:
+    def special_login_handler(self, delay_factor: float = 1.0) -> Optional[str]:
```
ktbyers, line 1143: I wouldn't expect this to change was there something that made you switch this? I would expect this to always return None.

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/2381


Minor update to it.

netmiko/base_connection.py
verailina: That was for `HuaweiSSH` class. `HuaweiSSH.special_login_handler` returns string.

PR #2385, opened 2021-07-18
K* l* type hints

verailina:
Added type hints for `keymile` and `linux` packages.


ktbyers:
Some minor updates here: https://github.com/ktbyers/netmiko/pull/2394

PR #2387, opened 2021-07-18
N* type hints

verailina:
Added type hints in `netapp_cdot_ssh`, `netgear_prosafe_ssh`, `nokia_sros` modules.

### DIFF
--- a/netmiko/nokia/nokia_sros.py
+++ b/netmiko/nokia/nokia_sros.py
@@ -9,6 +9,7 @@
 import re
 import os
 import time
+from typing import Any, Optional, Union, Sequence, TextIO
 
 from netmiko import log
 from netmiko.base_connection import BaseConnection
@@ -36,7 +37,7 @@ class NokiaSros(BaseConnection):
         - check_enable_mode()
     """
 
-    def session_preparation(self):
+    def session_preparation(self) -> None:
         self._test_channel_read()
         self.set_base_prompt()
         # "@" indicates model-driven CLI (vs Classical CLI)
@@ -58,7 +59,7 @@ class NokiaSros(BaseConnection):
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()
 
-    def set_base_prompt(self, *args, **kwargs):
+    def set_base_prompt(self, *args: Any, **kwargs: Any) -> str:
         """Remove the > when navigating into the different config level."""
         cur_base_prompt = super().set_base_prompt(*args, **kwargs)
         match = re.search(r"\*?(.*?)(>.*)*#", cur_base_prompt)
@@ -67,7 +68,7 @@ class NokiaSros(BaseConnection):
             self.base_prompt = match.group(1)
             return self.base_prompt
 
-    def _disable_complete_on_space(self):
+    def _disable_complete_on_space(self) -> str:
         """
         SR-OS tries to auto complete commands when you type a "space" character.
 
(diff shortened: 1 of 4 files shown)


netmiko/nokia/nokia_sros.py
```diff
@@ -152,11 +162,16 @@ def check_config_mode(self, check_string=r"(ex)[", pattern=r"@"):
...
-        return self.send_command(command_string="/admin save", expect_string=r"#")
+        return self._send_command_str(command_string="/admin save", expect_string=r"#")
 
-    def send_config_set(self, config_commands=None, exit_config_mode=None, **kwargs):
+    def send_config_set(  # type: ignore
+        self,
+        config_commands: Union[str, Sequence[str], TextIO, None] = None,
+        exit_config_mode: bool = None,
+        **kwargs: Any,
+    ):
```
ktbyers, line 174: No return value? It should define its return value for the function.

netmiko/nokia/nokia_sros.py
```diff
@@ -244,31 +259,51 @@ class NokiaSrosTelnet(NokiaSros):
...
+        self,
+        ssh_conn: BaseConnection,
+        source_file: str,
+        dest_file: str,
+        file_system: Optional[str] = None,
+        direction: str = "put",
+        socket_timeout: float = 10.0,
+        progress: Optional[Callable[..., Any]] = None,
+        progress4: Optional[Callable[..., Any]] = None,
+        hash_supported: bool = True,
```
ktbyers, line 271: You changed the argument here...needs to be `hash_supported: bool = False`

netmiko/nokia/nokia_sros.py
```diff
@@ -244,31 +259,51 @@ class NokiaSrosTelnet(NokiaSros):
...
-            ssh_conn, source_file, dest_file, hash_supported=hash_supported, **kwargs
+            ssh_conn,
+            source_file,
+            dest_file,
+            file_system,
+            direction,
+            socket_timeout,
+            progress,
+            progress4,
+            hash_supported,
```
ktbyers, line 282: Needs to be named arguments as I wouldn't want to count on the order of arguments here (for anything after the first three arguments).

netmiko/nokia/nokia_sros.py
```diff
@@ -286,8 +321,12 @@ def check_file_exists(self, remote_cmd=""):
                 raise ValueError("Unexpected output from check_file_exists")
         elif self.direction == "get":
             return os.path.exists(self.dest_file)
+        else:
+            raise ValueError("Unexpected value for self.direction")
 
-    def remote_file_size(self, remote_cmd=None, remote_file=None):
+    def remote_file_size(
+        self, remote_cmd: Optional[str] = None, remote_file: Optional[str] = None
```
ktbyers, line 328: I would switch this to be `remote_cmd: str = ""` to make it consistent with what is in `scp_handler` for remote file size.

netmiko/nokia/nokia_sros.py
```diff
@@ -299,11 +338,12 @@ def remote_file_size(self, remote_cmd=None, remote_file=None):
             remote_cmd = self._file_cmd_prefix() + "file dir {}/{}".format(
                 self.file_system, remote_file
             )
-        remote_out = self.ssh_ctl_chan.send_command(remote_cmd)
+        remote_out = self.ssh_ctl_chan._send_command_str(remote_cmd)
 
         if "File Not Found" in remote_out:
             raise IOError("Unable to find file on remote system")
 
+        assert remote_file is not None
```
ktbyers, line 346: I would just put an `else` clause up in the self.direction check where if self.direction is neither "put" nor "get" you raise an exception, that should ensure that remote_file is always set to a string and mypy should be able to work it out from that (hopefully).

ktbyers (changes requested):
See comments I posted.

netmiko/nokia/nokia_sros.py
verailina: Yes, mypy is able to process it :)

PR #2389, opened 2021-07-18
R* type hints

verailina:
Added type hints for `rad`, `raisecom`,  `ruckus` and `ruijie` packages.

### DIFF
--- a/netmiko/rad/rad_etx.py
+++ b/netmiko/rad/rad_etx.py
@@ -1,4 +1,6 @@
 import time
+from typing import Any
+
 from netmiko.no_enable import NoEnable
 from netmiko.base_connection import BaseConnection
 
@@ -6,7 +8,7 @@ from netmiko.base_connection import BaseConnection
 class RadETXBase(NoEnable, BaseConnection):
     """RAD ETX Support, Tested on RAD 203AX, 205A and 220A."""
 
-    def session_preparation(self):
+    def session_preparation(self) -> None:
         self._test_channel_read()
         self.set_base_prompt()
         self.disable_paging(command="config term length 0")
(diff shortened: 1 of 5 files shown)


netmiko/ruckus/ruckus_fastiron.py
```diff
@@ -50,7 +57,7 @@ def enable(
                     if not re.search(
                         r"error.*incorrect.*password", new_data, flags=re.I
                     ):
-                        return output
+                        break
```
verailina, line 60: @ktbyers Is it OK to break here and return the `output` at the end of this method? Will the following check  `if not self.check_enable_mode():` be false in case of correct output?

ktbyers: Yes, that is fine. You will need to add that final return statement though (just making sure we are on the same page here). Yes, that final exception shouldn't happen (i.e. that means something unexpected went wrong...for example, the while loop completed with never seeing the username/password.

ktbyers: Ohh, I see the `return output` now...it was just really hard to see in the GitHub diff.

ktbyers (changes requested):
See comments.

netmiko/rad/rad_etx.py
```diff
@@ -65,8 +73,14 @@ class RadETXTelnet(RadETXBase):
     """RAD ETX Telnet Support."""
 
     def telnet_login(
-        self, username_pattern=r"(?:user>)", alt_prompt_term=r"#\s*$", **kwargs
-    ):
+        self,
+        pri_prompt_terminator: str = r"#\s*$",
```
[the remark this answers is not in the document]
verailina: In `BaseConnection.telnet_login`, `pri_prompt_terminator` has the same value: `r"#\s*$"`. So, it seems to me, that it will work as before.

netmiko/ruckus/ruckus_fastiron.py
verailina: Should I change something here? :) This final `return output` should do the job, right?

ktbyers: Yes, this should be good.

PR #2393, opened 2021-07-18
W* Y* Z* type hints

verailina:
Added type hints in `fireware_ssh.py`,  `yamaha.py`  and  `zte_zxros.py` modules.


ktbyers:
Tests will fail until u-v type hints are also merged.

PR #2408, opened 2021-07-21
Resolve type: ignore - 2

verailina:
Fixed issues with `_auth` and  `send_config_set` methods.


ktbyers:
I consolidated the `no_auth` here:

https://github.com/ktbyers/netmiko/pull/2422


I decided not to do the `send_config_set` as it is a trade-off between maintainability (i.e. using **kwargs makes maintenance easier).

PR #2409, opened 2021-07-21
Resolve type: ignore - 3

verailina:
Resolved issues with `send_config_set` methods.


ktbyers:
Kept the Nokia one here:

https://github.com/ktbyers/netmiko/pull/2423


Dropped the others...since harder to maintain without the **kwargs :-(

PR #2427, opened 2021-07-28
Replace `netmiko._textfsm` with `textfsm` dependency

verailina:


setup.py
```diff
@@ -51,6 +51,7 @@ def find_version(*file_paths):
         "paramiko>=2.6.0",
         "scp>=0.13.2",
         "tenacity",
+        "textfsm>=1.1.1",
```
ktbyers, line 54: Probably better to make it be >= 1.1.2 (I don't think there is any textfsm `1.1.1`)

ktbyers:
Everything else looks good besides the one minor item in setup.py.

PR #2453, opened 2021-08-19
Update extreme_ers_ssh.py

Microbuse:
If SSH use Menu on login, this will take you to Command Line Interface.


ktbyers:
Wouldn't this cause issues if there was no `Menu`?

The code no longer breaks out of a successful send of the username/password?

Can you show me what the CLI session looks like when the `Menu` is present?

romanstech:
@ktbyers 
Look at my diff here (from Netmiko v.4.1.1):
https://gist.github.com/romanstech/c52bdc7b87885f51429feb9797a05ee4

I've tested this on switch with Ctrl-Y, without Ctrl-Y, with Menu, without Menu and this fix worked flawlessly.

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3169

PR #2460, opened 2021-08-20
Update PLATFORMS.md

tamaszl:
Added supported platform device_type:


ktbyers:
I wanted to keep the old form and also add the device_type so I did this instead:

https://github.com/ktbyers/netmiko/pull/2466/files

PR #2470, opened 2021-09-01
Script to test performance and generate graphs

verailina:
`gen_graph.py`:
* Automatically generate graphs for all platforms listed in `test_devices.yml`
* Generate a markdown report containing all graphs
* Filter patch versions to display on the graph

`netmiko_performance.csv` has been regenerated in order to get performance data for all required platforms.


tests/performance/test_performance.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+export PYTHONPATH=/home/kbyers/netmiko/tests
+cd /home/vera/netmiko/tests/performance
```
ktbyers, line 4: Need to get rid of hard-coding to a home directory. We can use ~ for the PYTHONPATH.

I would probably just get rid of the `cd` command and then if either of the test_netmiko.py or gen_graph.py fail add into the failure message that you need to be in {NETMIKO_REPO}/tests/performance directory.

tests/performance/test_performance.sh
```diff
@@ -0,0 +1,37 @@
...
+
+pip install pyaml pygal jinja2
```
ktbyers, line 6: Let's not do the pip install and instead do a "pip show" on each of the packages and print a message if they are not installed. I say this as the most likely case is probably me trying to run this using the wrong virtual environment and I would rather know that I was in the wrong venv than installing the packages.

ktbyers:
@verailina How hard would it be to remove duplicate entries from the network_performance.csv?

What I mean by this is--if I run re-run the script now using Netmiko 4.0.0a4 (current develop branch) then I will get two sets of data for 4.0.0a4. It would be better to remove the old 4.0.0a4 data and add the new data to network_performance.csv.

ktbyers:
This is probably a separate Pull Request, but what would be the process of adding a new platform? Say for example, I wanted to add "hp_procurve" and to generate all of the data for Netmiko 2.4.2 through current develop branch. It would be nice to have a relatively easy way of generating all of that including cycling through all of the Netmiko versions.

This probably should be a separate script though as it is not the standard process i.e. the standard process is what you did here.

verailina:
> @verailina How hard would it be to remove duplicate entries from the network_performance.csv?
> 
> What I mean by this is--if I run re-run the script now using Netmiko 4.0.0a4 (current develop branch) then I will get two sets of data for 4.0.0a4. It would be better to remove the old 4.0.0a4 data and add the new data to network_performance.csv.

Yes, good idea! I added a function `remove_old_data` called at the beginning of  `test_netmiko.py` script.

PR #2614, opened 2021-12-21
Update Platforms.md to include NVIDIA-Mellanox

Yuval-Mellanox:


PLATFORMS.md
```diff
@@ -185,6 +184,7 @@
 - netgear_prosafe
 - netscaler
 - nokia_sros
+- NVIDIA-Mellanox
```
ktbyers, line 187: This list here is actually the driver name i.e. it needs to actually match the Netmiko code for the driver name.

You could put this change further up on the page where it just refers to the name (in a general sense). In other words, you could change line 41, but not this line.

ktbyers:
See comment that I added.

ktbyers:
Superseded by https://github.com/ktbyers/netmiko/pull/2629

PR #2616, opened 2021-12-28
AutoDetect support for Dell OS6

Dev01024:
Add AutoDetect support for Dell OS6


ktbyers:
You need to Git Rebase your code. It looks like your code is using a very old starting point which is why you are getting all of these unrelated commits.

ktbyers:
Closing as the PR needs rebased. If you rebase please feel free to resubmit.

PR #2630, opened 2022-01-22
Cisco UC CLI find_prompt fix

llama85:
Using 3.4.0 I was unable to get a prompt when connecting to CUCM 12.5

the connection would time out saying that the prompt was never seen, when enabling debugging, the prompt was returned.

by adding a tertiary prompt terminator into set_base_prompt I have been able to connect and send commands without issue

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1229,6 +1229,7 @@ Device settings: {self.device_type} {self.host}:{self.port}
         self,
         pri_prompt_terminator: str = "#",
         alt_prompt_terminator: str = ">",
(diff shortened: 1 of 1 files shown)


ktbyers:
Which Netmiko device_type are you using? I don't think there is any Cisco UCM driver so I think the solution would be to add a UCM driver.

The process for adding a driver is here:

https://github.com/ktbyers/netmiko/blob/develop/VENDOR.md

I wouldn't be willing to add the tertiary prompt so I am going to close this PR.

llama85:
that's totally fair.
I tried with cisco_ios and linux.
I'll have a look at adding a driver for it.

Issue #2639, opened 2022-02-06
Add support for file transfer on Extreme EXOS


teixemf:
This is a FR to add support for file transfer on Extreme EXOS.

Notes:
- the OS has no support for file hashing;
- the OS has no support for disable/enable SCP

ktbyers:
I am going to close this as we already have the Pull Request open to track it.

PR #2640, opened 2022-02-06
#2639 Add support for file transfer on Extreme EXOS

teixemf:
Issue: #2639
This PR adds support for for file transfer on Extreme EXOS


ktbyers:
Will try to look at this in the next week.

ktbyers:
@teixemf Can you fix the Black/linting issue?

teixemf:
Thanks, @ktbyers 
I believe the issue is gone.

ktbyers:
@teixemf Is there any way on EXOS to verify the file transferred is correct (md5/sha1)?

This is typically a requirement for Netmiko SCP.

teixemf:
> @teixemf Is there any way on EXOS to verify the file transferred is correct (md5/sha1)?
> 
> This is typically a requirement for Netmiko SCP.

@ktbyers 
up until the latest EXOS release (31.6) it does not provide any built-in option to verify if the file was transferred correctly.
That's the reason why I disable the option on the code I'm submitting.

If file verification is typically a requirement for Netmiko SCP, could it be accepted whiteout the verification since it is a device limitation? Maybe with a disclaimer in the code?

Technically, there is a way to do it but... _these are not the droids you are looking for...._
It requires writing a custom made python script an run it directly from the EXOS switch.
The script would import the python hashlib, compute the file hash and print it out to the CLI. 
I successfully did it as a prof-of-concept but IMO this approach is not suitable to incorporate into Netmiko. We can not expect all EXOS devices to have such script.

ktbyers:
@teixemf Is there a way you can verify the file size?

This is what we did for Nokia SR-OS, see code example here:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/nokia/nokia_sros.py#L361-L386

teixemf:
> @teixemf Is there a way you can verify the file size?
> 
> This is what we did for Nokia SR-OS, see code example here:
> 
> https://github.com/ktbyers/netmiko/blob/develop/netmiko/nokia/nokia_sros.py#L361-L386

Yes, I already had the function _remote_file_size()_.
All I needed was to override the function _verify_file()_ on the class _ExtremeExosFileTransfer_ with a copy-paste from the function _verify_file()_ on the class _NokiaSrosFileTransfer_  

I committed the change but I don't feel confortable with the copy-paste.
I believe it might be better if the function _verify_file()_ located at the base class _BaseFileTransfer_ was refactored to allow the transfer to be verified based on file hash or based on file size if  _hash_supported=False_

https://github.com/ktbyers/netmiko/blob/49f073f967a1e2198d47e4be03783d2d5b13f08e/netmiko/scp_handler.py#L411-L413

ktbyers:
Yeah, the verify_file you implemented should be fine. The class hierarchy will use your verify_file method.

I assume you ran the Netmiko secure copy tests against this code?

Note, I merged a PR with this change where I fixed the linting. See the PR linked to above.  @teixemf 

I am going to go ahead and close this PR.

PR #2657, opened 2022-03-01
Enable ANSI code handling for zyxel_os devices

ajveitch:
ZyNOS generates ANSI escape codes in output.  This change enables ANSI escape code handling.


ktbyers:
You need to also call `super()` on the parent class for `session_preparation` method (or some other needed things won't get done).

Can you adjust the PR to do this?

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/2660

PR #2660, opened 2022-03-01
Zyxel enable ANSI escape code stripping

ktbyers:


ktbyers:
@ajveitch Just let me know if this doesn't work for some reason. I think it should.

PR #2670, opened 2022-03-09
Fix send_command() read issue where data spanned the deque

ktbyers:
This resulted in the terminating pattern never being detected

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1111,31 +1111,22 @@ Device settings: {self.device_type} {self.host}:{self.port}
         delay_factor = self.select_delay_factor(delay_factor=0)
 
         if pattern:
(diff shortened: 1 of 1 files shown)


ktbyers:
Issue was on A10 platform using threads and AAA (both authentication and authorization) and large number of devices.

The deque would only come back with part of terminating pattern in other words, you would never get the full trailing prompt in the deque (I assume you must have gotten multiple empty reads back for this to happen).

This is compounded in Netmiko 4.x code that read loop was very fast (i.e. made it a lot more likely).

Fix is:
```
Increase the loop sleep from 10mS to 25mS
Increase the size of the deque from 3 to 20
Check the total size of the output and if it is below some threshold like 2_000_000 characters, then check the whole output.
```

PR #2671, opened 2022-03-09
Add support for VyOS operator users in prompt detection

### LINKED ISSUE #1591: Support for VyOS operator mode: need 3 different prompt_terminator?
In operator mode, the prompt terminator is ``>`` instead of ``$`` or ``#``. 

The current framework only supports a ``pri_prompt_terminator`` and an ``alt_prompt_terminator``. Maybe ``alt_prompt_terminator`` can be a list?

blkmajik:
Fixes #1591

I just ran into this bug today, and this is a minimal fix to address the need for a 3rd prompt terminator for VyOS when using an "operator" user instead of an "admin" user.   It should be generic and apply for 3+ terminators (adds a list as a defaulted keyword argument) should future routers also need this functionality.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1252,6 +1252,7 @@ Device settings: {self.device_type} {self.host}:{self.port}
         pri_prompt_terminator: str = "#",
         alt_prompt_terminator: str = ">",
         delay_factor: float = 1.0,
+        additional_terminators: List[str] = [],
     ) -> str:
         """Sets self.base_prompt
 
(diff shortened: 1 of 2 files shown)


ktbyers:
@blkmajik Thanks for this PR.

I happened to be working on the set_base_prompt code in Netmiko for a different reason today so I fixed this issue there. Basically I allowed for `set_base_prompt()` to take/use a regular expression pattern and in the case of VYoS I set that regex pattern to be any of the three possible prompt terminators.

See PR here:

https://github.com/ktbyers/netmiko/pull/2673

Let me know if that doesn't work for some reason (as I don't have a way to test on VyOS).

I am going to close this PR as it should not be needed now.

ktbyers:
@blkmajik Thanks for this PR.

I happened to be working on the set_base_prompt code in Netmiko for a different reason today so I fixed this issue there. Basically I allowed for `set_base_prompt()` to take/use a regular expression pattern and in the case of VYoS I set that regex pattern to be any of the three possible prompt terminators.

See PR here:

https://github.com/ktbyers/netmiko/pull/2673

Let me know if that doesn't work for some reason (as I don't have a way to test on VyOS).

I am going to close this PR as it should not be needed now.

PR #2714, opened 2022-03-31
Fix aruba controller paging issue

ktbyers:


EarlRamirez:
After this change was implemented "no paging" broke Aruba AOS-CX at least on version 10.09.0010 where paging was removed in favor of page. Therefore, when you send a command, for example, 'show running-config' the output is stuck at the pager because it was not executed. As a dirty fix, I updated the code and added 'no page'.

ktbyers:
@EarlRamirez `aruba_os` device_type is for Aruba Wireless Controllers only. 

You should try `aruba_osswitch` for the Netmiko device_type for Aruba CX (though this is technically for Aruba Switches that are rebranded ProCurves so Aruba CX may or may not work). 

Let me know if it works or not.

EarlRamirez:
@ktbyers, thanks, it works.

Issue #2711, opened 2022-03-28
Huawei special_login_handler is not logging in successfully


quinntwohao:
The system information is as follows:
1. netmiko version: 4.0
2. python 3.10
3. window 11

error_print:
```shell
Traceback (most recent call last):
  File "E:\web_API\test.py", line 11, in <module>
    app.net_ssh_proxy(switch_json = switch_json, commands=commands)
  File "E:\web_API\app.py", line 25, in net_ssh_proxy
    with ConnectHandler(**device_info, sock=sock) as net_connect:
  File "E:\venv_02\lib\site-packages\netmiko\ssh_dispatcher.py", line 344, in ConnectHandler
    return ConnectionClass(*args, **kwargs)
  File "E:\venv_02\lib\site-packages\netmiko\base_connection.py", line 434, in __init__
    self._open()
  File "E:\venv_02\lib\site-packages\netmiko\base_connection.py", line 439, in _open
    self.establish_connection()
  File "E:\venv_02\lib\site-packages\netmiko\base_connection.py", line 1092, in establish_connection
    self.special_login_handler()
  File "E:\venv_02\lib\site-packages\netmiko\huawei\huawei.py", line 105, in special_login_handler
    output = self.read_until_pattern(password_change_prompt)
  File "E:\venv_02\lib\site-packages\netmiko\base_connection.py", line 631, in read_until_pattern
    raise ReadException(msg)
netmiko.exceptions.ReadException: Unable to successfully split output based on pattern:
pattern=((Change now|Please choose))|([\]>]\s*$)
output='\nInfo: The max number of VTY users is 21, the number of current VTY users online is 2, and total number of terminal users online is 2.\n      The current login time is 2022-03-28 15:55:30+08:00.\n<xxxx_hostname>'
results=['\nInfo: The max number of VTY users is 21, the number of current VTY users online is 2, and total number of terminal users online is 2.\n      The current login time is 2022-03-28 15:55:30+08:00.\n<xxxx_hostname', None, None, '>', '']

```

test instanse
2. python 3.10
3. window 11
[Netmiko 3.4.0 Release](https://github.com/ktbyers/netmiko/releases/tag/v3.4.0)

out_print

no problem

migmorales22:
are you trying to do it with Huawei device_type?

Cause I have the same problem.
Tes instance
1. python 3.9.7
2. Mac OS
3. Netlike 4.0.0

**from netmiko import ConnectHandler
from getpass import getpass
[...]

PR #2719, opened 2022-04-04
Fix Huawei special_login_handler (#2711)

fharbe:
This is a proposal to fix the `special_login_handler()` issue (#2711).

TL;DR: I removed the `special_login_handler()` completely and moved the handling for the password change prompt into `session_preparation()`, because I did not see a better way of fixing the issues altogether (read the whole story below).

After spending some time investigating the issue I found out that the pattern matching logic in `read_until_pattern` expects the pattern to split exactly in 3 parts (using re.split), see [base_connection.py#L579](https://github.com/ktbyers/netmiko/blob/c76d09fad12c98b1e70176906a4129a53d9be06c/netmiko/base_connection.py#L624) for reference. Seems this change was introduced during some fundamental changes to the channel reading logic in in https://github.com/ktbyers/netmiko/commit/793100dcfd76d3eb5089ba4cd27c99e3c3c0f886.

At first I tried to simply fix the `password_change_prompt` regex (see [huawei.py#L104](https://github.com/ktbyers/netmiko/blob/develop/netmiko/huawei/huawei.py#L104)) by replacing the inner groups (parenthesis) with `(?:)` like so:

```python
        password_change_prompt = r"((?:Change now|Please choose))|(?:[\]>]\s*$)"
```

However, when looking more closely at the code, I stumbled upon the `if re.search(password_change_prompt, output):` condition, which seemed wrong (would always evaluate to `True`). So I modified the if-condition to actually search for the password_change_prompt only, like so:

```python
password_change_prompt = r"(?:Change now|Please choose).+"
combined_pattern = r"({}|[>\]])".format(password_change_prompt)
output = self.read_until_pattern(combined_pattern, read_timeout=3.0)
if re.search(password_change_prompt, output):
    self.write_channel("N\n")
```

At first it seemed to work, but it turns out that connecting to the device now only works if a password change prompt is present. When no password change prompt is found, the following exception is raised:

```python
netmiko.exceptions.ReadTimeout: 
Pattern not detected: '[>\\]]' in output.
```

Huh? Inspecting the debug logs, it turns out that  `self.write_channel("N\n")` (which was always called before changing the if-condition) was responsible for the following debug logs: 

```python
DEBUG:netmiko:read_channel: N
DEBUG:netmiko:read_channel: 
        ^
Error: Unrecognized command found at '^' position.
<AC6508>
```

This means, the `self.write_channel("N\n")` did not only produce an error, but also a new prompt!

To me, it looks like it was only this new prompt (which was triggered by the  "Unrecognized command" error) that actually made `_test_channel_read` in `session_preparation` work at all.

Therefore, I think it would only be possible to keep the handling logic for the password change prompt in `special_login_handler` if either `_test_channel_read` in `session_preparation` would be removed completely or `self.write_channel(self.RETURN)` would be needed to to trigger a new prompt for  `_test_channel_read` to read it. Both of which are not really elegant IMHO. Maybe it would also be possible to read until the prompt by using a lookahead assertion, but I don't know if that's possible with the way `read_until_pattern` is currently implemented.

Let me know what you think.

Kind regards
  Flo

### DIFF
--- a/netmiko/huawei/huawei.py
+++ b/netmiko/huawei/huawei.py
@@ -6,7 +6,7 @@ import warnings
 from netmiko.no_enable import NoEnable
 from netmiko.base_connection import DELAY_FACTOR_DEPR_SIMPLE_MSG
 from netmiko.cisco_base_connection import CiscoBaseConnection
-from netmiko.exceptions import NetmikoAuthenticationException
+from netmiko.exceptions import NetmikoAuthenticationException, ReadTimeout
 from netmiko import log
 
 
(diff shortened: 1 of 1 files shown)


fharbe:
@ktbyers Hopefully fixed the issues. Tests should pass now :slightly_smiling_face:

nivaldoinacios (approved):
work here

ktbyers:
Superseded by: 

https://github.com/ktbyers/netmiko/pull/2728

PR #2728, opened 2022-04-12
Updates to Huawei Login issue fix

ktbyers:


ktbyers:
@fharbe First thanks for the fix.

I read through your comments on the PR and I believe the original issue must have been this code here

```python
        password_change_prompt = r"((Change now|Please choose))|([\]>]\s*$)"
        output = self.read_until_pattern(password_change_prompt)
        if re.search(password_change_prompt, output):
            self.write_channel("N\n")
            self.clear_buffer()
```
 
coupled with the _test_channel_read() call.

In other words, since that password_change_prompt pattern was incorrectly defined and would cause the initial prompt to be read (so the above would ONLY work if the special password change prompting was present).

I made a few updates to your PR.

1. I moved the code back to special_login_handler() as that is the standard location in Netmiko for that type of code and I didn't see any reason it would work there.
2. I moved this special_login_handler() method to be in HuaweiBase as that code should apply to both SSH and telnet (as far as I could see).
3. I got rid of your try / except ReadTimeout and of the 3 second read_timeout. Basically, one of those conditions should always apply (either "Change now|Please choose" or the standard prompt) and Netmiko should keep reading until one of those happens. 

From a Netmiko perspective it is bad for the program to continue onward if the prompt is not successfully read. So if this doesn't happen it should raise an exception.

Basically, Netmiko would fail later in a hard to diagnose way if the prompt isn't read (i.e. we are fully logged in and the device should be responsive).

Do you see any issues with the updated PR?

Can you test that it still works properly?


Regards, Kirk

Issue #2628, opened 2022-01-21
Ubiquity Edge Switch is failing on save_config due to the wrong search pattern.


DarkPhily:
In the method save_config of the class `UbiquitiEdgeSSH` are you waiting for the wrong search pattern.
At the moment you are expecting `\(UBNT\ EdgeSwitch\)\ \#`.

After the command `write memory` the cli print following out:
```
This operation may take a few minutes.
Management interfaces will not be available during this time.

Are you sure you want to save? (y/n) 
```

It's expecting an additional input.

This is how i handle it, maybe you can do it the same way:
```
    def save_config(self):
        try:
            output = self.send_command(command_string="write memory", expect_string=r"(y/n)", strip_command=False, strip_prompt=False)
            output += self.send_command(command_string="y", expect_string=r"#", strip_command=False, strip_prompt=False)
            return output
        except Exception as e:
            raise e
```

ktbyers:
@DarkPhily Is this some special subcase that prompts for this? Is there a more global way to disable this (so that this prompting does not happen).

Very probably this would have been tested for some class of UbuitiEdge Switch (i.e. there is some class of Edge Switch that don't do this).

Also can you put more of the CLI output (before and after what you put above)?

Finally, just to verify what model and OS version are you on?

DarkPhily:
@ktbyers It doesn't appear to have a way to disable this.


First example:
```
(UBNT EdgeSwitch) #boot host dhcp 

(UBNT EdgeSwitch) #write memory

This operation may take a few minutes.
Management interfaces will not be available during this time.

Are you sure you want to save? (y/n) y

Config file 'startup-config' created successfully .


Configuration Saved!

(UBNT EdgeSwitch) #
```

Second example:
```
(UBNT EdgeSwitch) #no ip telnet server enable 

(UBNT EdgeSwitch) #write memory 

This operation may take a few minutes.
Management interfaces will not be available during this time.

Are you sure you want to save? (y/n) y

Config file 'startup-config' created successfully .


Configuration Saved!

(UBNT EdgeSwitch) #
```

Model: `Ubiquiti ES-16-XG`
Version: `1.9.2`

ktbyers:
Okay, this looks like a bug that needs fixed.

ktbyers:
Fixed in referenced PR

PR #2731, opened 2022-04-13
Fix issue with 'write memory' on ubiquity edge switches

ktbyers:
Fixes https://github.com/ktbyers/netmiko/issues/2628


ktbyers:
@DarkPhily Ubiquity edge switch write memory / save_config() issue is fixed here.

If you could test it, that would be great.

I am going to go ahead and merge it, but let me know if you run into any issues with it.

DarkPhily:
@ktbyers Works great. I tested it alone with different firmware versions and with my own functions which use save_config().

ktbyers:
@DarkPhily Great, thanks for testing.

PR #2737, opened 2022-04-15
Huawei fix regular expression issue on login

ktbyers:


ktbyers:
Had to move special_login_handler back to Huawei SSH class as telnet already had a solution for this problem (in telnet_login method).

Also fix issues with regular expression patterns not working properly on the initial login.

Issue #2744, opened 2022-04-22
120 second teardown after command with excessively long output


zohassadar:
The condition is a command that takes an excessively long time to complete while providing a continuous stream of data during execution.  In this example, the command is `ping 192.168.42.1 size 18024 repeat 2147483647`.  The read_timeout, adjusted to 2 seconds, works as expected.   During the teardown the call to `check_config_mode`, which in turn calls `read_channel_timing`, introduces a 120 second delay.  This is due to read_channel_timing's default read_timeout [value of 120.0](https://github.com/ktbyers/netmiko/blob/v4.0.0/netmiko/base_connection.py#L656).  

```
import os
import time

import netmiko

def send_command_with_timeout_context(command: str, read_timeout: int):
    pre_start = time.time()
    with netmiko.ConnectHandler(
        host="192.168.42.254",
        device_type="cisco_ios",
        username=os.getenv("USERNAME"),
        password=os.getenv("PASSWORD"),
    ) as session:
        try:
            error = None
            start = time.time()
            session.send_command(command, read_timeout=read_timeout)
        except netmiko.ReadTimeout:
            error = 'timeout'
        finally:
            pre_finish = time.time()
    finish = time.time()
    print(f"{command=} {read_timeout=} setup={start-pre_start:.02f} execution={pre_finish-start:.02f} teardown={finish-pre_finish:.02f} {error=}")

send_command_with_timeout_context("ping 192.168.42.1 size 18024 repeat 2147483647", 2)
```

The result:

```
command='ping 192.168.42.1 size 18024 repeat 2147483647' read_timeout=2 setup=0.78 execution=2.05 teardown=120.01 error='timeout'
```

A lower read_timeout value of 2.0 (chosen to match the default value of last_read) when called from check_config_mode would solve this issue, and in normal conditions the results from the carriage return sent shouldn't require more than 2 seconds.   Attaching pull request.  Willing to adjust if there's more that needs to be taken into consideration.  

The result after adjusting the read_timeout:

```
command='ping 192.168.42.1 size 18024 repeat 2147483647' read_timeout=2 setup=0.73 execution=2.05 teardown=2.00 error='timeout'
```

ktbyers:
@zohassadar  Is there any broader purpose here besides testing? I.E. I can't really see any reason to do the above besides to create an artificial test case that gums up the CLI with a continuously running command and output (obviously you can unintentionally gum up the CLI by accidentally executing a command that you shouldn't).

[...]

PR #2745, opened 2022-04-22
call read_channel_timeout from check_config_mode with read_timeout value of 2.0

zohassadar:
Reference https://github.com/ktbyers/netmiko/issues/2744


zohassadar:
This will not work as `check_config_mode` is called outside of cleanup mode when 2 seconds wouldn't be enough.  Will submit something different.

ktbyers:
I re-opened this so I could track that I need to fix that underlying failure i.e. I didn't want to lose track of this.

ktbyers:
Superseded by https://github.com/ktbyers/netmiko/pull/2751

PR #2747, opened 2022-04-22
Update vyos_ssh.py

showipintbri:
Implemented `save_config()` function.

Tested & Validated on VyOS 1.3 and 1.4.

Used as code reference: https://github.com/ktbyers/netmiko/blob/89b45fb1977aa04d685fc240c83496c37442353c/netmiko/ubiquiti/edgerouter_ssh.py#L18

### DIFF
--- a/netmiko/vyos/vyos_ssh.py
+++ b/netmiko/vyos/vyos_ssh.py
@@ -132,9 +132,12 @@ class VyOSSSH(NoEnable, CiscoSSHConnection):
 
     def save_config(
         self,
-        cmd: str = "copy running-config startup-config",
+        cmd: str = "save",
         confirm: bool = False,
-        confirm_response: str = "",
+        confirm_response: str = ""
     ) -> str:
-        """Not Implemented"""
-        raise NotImplementedError
+        """Saves Config."""
+        output = self._send_command_str(command_string=cmd)
+        if "Done" not in output:
+            raise ValueError(f"Save failed with following errors:\n\n{output}")
+        return output
--- a/netmiko/vyos/vyos_ssh.py
+++ b/netmiko/vyos/vyos_ssh.py
@@ -131,13 +131,12 @@ class VyOSSSH(NoEnable, CiscoSSHConnection):
         )
 
     def save_config(
-        self,
-        cmd: str = "save",
-        confirm: bool = False,
-        confirm_response: str = ""
+        self, cmd: str = "save", confirm: bool = False, confirm_response: str = ""
     ) -> str:
         """Saves Config."""
         output = self._send_command_str(command_string=cmd)
         if "Done" not in output:
             raise ValueError(f"Save failed with following errors:\n\n{output}")
-        return output
+        return super().save_config(
+            cmd=cmd, confirm=confirm, confirm_response=confirm_response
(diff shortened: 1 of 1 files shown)


netmiko/vyos/vyos_ssh.py
```diff
@@ -132,9 +132,12 @@ def send_config_set(
 
     def save_config(
         self,
-        cmd: str = "copy running-config startup-config",
+        cmd: str = "save",
```
ktbyers, line 135: Did this change in vyos at some point? I would have thought the original contributor tested it so I am wondering why it used to work and now doesn't (though maybe it was never tested, I would have to look through some history to figure that out).

ktbyers:
I deleted a comment as I was only looking at the arguments and missed the NotImplemented in the body of the old code.

ktbyers:
It looks like you have a linting issue (i.e. just need to run `black` on that one file you changed).

Also it would be better if you called `super()` with all the arguments. In this way, if there is ever any case where vyos prompts you for confirmation on save that someone could feed in arguments to handle that confirmation.

Thanks for the contribution.

showipintbri:
Kirk, thanks for the feedback. I've made the linting and code changes as you recommended. 

Can you please read my latest commit comment, I wanted to get your opinion on why this appears to be 'double-saving' now.

netmiko/vyos/vyos_ssh.py
```diff
@@ -131,10 +131,12 @@ def send_config_set(
...
-        self,
-        cmd: str = "copy running-config startup-config",
-        confirm: bool = False,
-        confirm_response: str = "",
+        self, cmd: str = "save", confirm: bool = False, confirm_response: str = ""
     ) -> str:
-        """Not Implemented"""
-        raise NotImplementedError
+        """Saves Config."""
+        output = self._send_command_str(command_string=cmd)
```
ktbyers, line 137: You should be able to just do the following:

```python
        output = return super().save_config(
            cmd=cmd, confirm=confirm, confirm_response=confirm_response
        )
        if "Done" not in output:
            raise ValueError(f"Save failed with following errors:\n\n{output}")
        return output

The `save` command will be sent to the remote device in the parent method (i.e. as part of that super() call).

Comment (showipintbri) [reply netmiko/vyos/vyos_ssh.py:137]: Your recommendation caused an error:
```
```
    def save_config(
        self, cmd: str = "save", confirm: bool = False, confirm_response: str = ""
    ) -> str:
        """Saves Config."""
        output = self._send_command_str(command_string=cmd)
        return super().save_config(
            cmd=cmd, confirm=confirm, confirm_response=confirm_response
        )
        if "Done" not in output:
            raise ValueError(f"Save failed with following errors:\n\n{output}")
        return output
```

ktbyers: Ooops I screwed up the syntax above :-)

ktbyers:
PR superseded by:

https://github.com/ktbyers/netmiko/pull/2752

showipintbri:
Works, perfect. No more double saves.

...and after reviewing my code I can see why it was double-saving: I was returning `output` twice. This was a simple error I overlooked.

PR #2752, opened 2022-04-25
Update vyos save_config()

ktbyers:


ktbyers:
@showipintbri Can you test that this works properly? It should be very similar to what you had (except eliminates the double save issue that we were discussing).

ktbyers:
I am going to go ahead and merge and if it is still broken, then we can fix it.

showipintbri:
Works, perfect. No more double saves.

...and after reviewing my code I can see why it was double-saving: I was returning output twice. This was a simple error I overlooked.

PR #2788, opened 2022-05-18
Replace typing.re import

srittau:
`typing.re` has been deprecated since Python 3.8 and will be removed in Python 3.12.
`Pattern` is also available in the top-level `typing` namespace.


srittau:
The reason for that is that `typing.re` was never known to type checkers as it's not part of typeshed. (This is not flagged due to --ignore-missing-imports being passed to mypy.) Therefore, `Pattern` was `Any` before. Anyway, I added a type argument, could you verify that `str` is correct?

ktbyers:
mypy in ci-cd is still broken.

ktbyers:
I am going to back this change out as it is still breaking CI-CD.

PR #2793, opened 2022-05-19
HP ProCurve improvements; fix enable() bug

ktbyers:


ktbyers:
- Fixed regex missing parenthesis issue.
- Defaulted enable() username to self.username
- Incorporated disabled_algoritms to auto-disable ["rsa-sha2-256", "rsa-sha2-512"]
- Netmiko auto-calls enable() in session_preparation() as HP ProCurve requires this to disable output paging.

Issue #2669, opened 2022-03-09
Mikrotik hEX - Prompt not truncated but output randomly truncated


elavaud:
Hi @ktbyers 

We are seeing a weird behavior on mikrotik hEX router. 
First ,the prompt is not removed, and the output is randomly truncated.
See 4 consecutive tries below. The last one has the output intended. The third one even cuts the output in the middle of a line...

(`driver.client` corresponds to netmiko. We are on the HEAD of develop, with the few PR I made on top of it.)

```
>>> print(driver.client.send_command('interface ethernet print detail terse'))
[admin@MikroTik] > interface ethernet print detail terse

 0 R  name=ether1 default-name=ether1 mtu=1500 l2mtu=1596 mac-address=2C:C8:1B:04:46:FB orig-mac-address=2C:C8:1B:04:46:FB arp=enabled arp-timeout=auto loop-protect=default loop-protect-status=off loop-protect-send-interval=5s loop-protect-disable-time=5m auto-negotiation=yes advertise=10M-half,10M-full,100M-half,100M-full,1000M-half,1000M-full full-duplex=yes tx-flow-control=off rx-flow-control=off speed=1Gbps bandwidth=unlimited/unlimited switch=switch1 
 1 RS name=ether2 default-name=ether2 mtu=1500 l2mtu=1596 mac-address=2C:C8:1B:04:46:FC orig-mac-address=2C:C8:1B:04:46:FC arp=enabled arp-timeout=auto loop-protect=default loop-protect-status=off loop-protect-send-interval=5s loop-protect-disable-time=5m auto-negotiation=yes advertise=10M-half,10M-full,100M-half,100M-full,1000M-half,1000M-full full-duplex=yes tx-flow-control=off rx-flow-control=off speed=1Gbps bandwidth=unlimited/unlimited switch=switch1 
 2  S name=ether3 default-name=ether3 mtu=1500 l2mtu=1596 mac-address=2C:C8:1B:04:46:FD orig-mac-address=2C:C8:1B:04:46:FD arp=enabled arp-timeout=auto loop-protect=default loop-protect-status=off loop-protect-send-interval=5s loop-protect-disable-time=5m auto-negotiation=yes advertise=10M-half,10M-full,100M-half,100M-full,1000M-half,1000M-full full-duplex=yes tx-flow-control=off rx-flow-control=off speed=1Gbps bandwidth=unlimited/unlimited switch=switch1 
[...]

PR #2797, opened 2022-05-25
Mikrotik strip command

elavaud:
Hello,

We noticed an issue with mikrotik having the command not stripped from output. It was then messing up the ntc template.
Half fixing #2669 

Before:

```
>>> print(net_connect.send_command("system routerboard print"))
[admin@MikroTik] > system routerboard print

       routerboard: yes
        board-name: hEX
             model: RB750Gr3
          revision: r4
     serial-number: CC210E592C82
     firmware-type: mt7621L
  factory-firmware: 6.48
  current-firmware: 6.49
  upgrade-firmware: 6.49
```

After:
```
 >>> print(net_connect.send_command("system routerboard print"))
routerboard: yes
        board-name: hEX
             model: RB750Gr3
          revision: r4
     serial-number: CC210E592C82
     firmware-type: mt7621L
  factory-firmware: 6.48
  current-firmware: 6.49
  upgrade-firmware: 6.49
```

### DIFF
--- a/netmiko/mikrotik/mikrotik_ssh.py
+++ b/netmiko/mikrotik/mikrotik_ssh.py
@@ -1,4 +1,5 @@
 from typing import Any, Union, List, Dict, Optional
+import re
 from netmiko.no_enable import NoEnable
 from netmiko.cisco_base_connection import CiscoSSHConnection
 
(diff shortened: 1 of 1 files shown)


ktbyers:
@elavaud Any idea why this is needed (i.e. why command stripping normally works on Mikrotik, but fails here)?

Can you print out a `repr(output)` in the case where it fails (preferably before any command stripping has happened as I would like to see what is actually there).

elavaud:
Without stripping I have a

```
system routerboard print
[admin@MikroTik] > system routerboard print

       routerboard: yes
        board-name: hEX
             model: RB750Gr3
          revision: r4
     serial-number: CC210E592C82
     firmware-type: mt7621L
  factory-firmware: 6.48
  current-firmware: 6.49
  upgrade-firmware: 6.49
```

The first line is well taken out by the command stripping, but not the second.
Let me check with the send_command_timing now that it is fixed

ktbyers:
I made some updates here...it should be similar to your original solution.

https://github.com/ktbyers/netmiko/pull/2810

Superseded by that PR.

PR #2799, opened 2022-05-27
Add Nokia SR Linux support in netmiko

sebastien-guay:
This is to add Nokia SR Linux support in netmiko. 

- A new class named NokiaSrlSSH has been added which support all the basic operations. There is no file transfer at the moment.
- Tests parameters have been added for Nokia SRL in commands.yml.example, responses.yml.example and test_devices.yml.example.
- Folder test_nokia_srl has been added in /tests with a README and a containerlab file. The README explains how to use containerlab to run test_srl.sh and execute test_netmiko_show.py and test_netmiko_config.py. There is also a txt file showing the output of a test run.

Tell me if something is missing to have this merged in the develop branch, thanks.


sebastien-guay:
Hi @ktbyers, 

I'm Sebastien and I created a pull request from my fork to you develop branch to add Nokia SR Linux support to netmiko. Since I'm a new contributor there is a workflow approval that is required.

hellt:
Hi @ktbyers 
I am part of this initiative as well and wanted to make sure that you are able to test this integration (or even integrate srlinux container in gh actions and make it all automated)

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,197 @@
...
+        supported first line prompt configuration:
+        {modified_flags}{mode_and_session} }}--[ {pwc} ]
+        """
+        if not pattern:
+            output = self.read_channel_timing(read_timeout=10.0)
+        else:
+            output = self.read_until_pattern(pattern=pattern)
+
+        matches = re.search(check_string, output)
+        return True if matches else False
```
ktbyers, line 108: Let's actually add some sort of argument to the parent method (as I had been thinking about doing this anyways). Something like,

force_regex = True

And in the `check_config_mode()` parent method in base_connection.py we will do:

```python
     def check_config_mode(
         self, 
         check_string: str = "", 
         pattern: str = "",
         force_regex: bool = False
) -> bool:
```

In other words, in the `check_config_mode()` parent method we will provide an option for this to be `regex-based` which we will default to False for backwards compatibility.

And then in your code we will just call `super()` on it with force_regex=True.

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,197 @@
...
+        confirm_response: str = "",
+    ) -> str:
+        """Save current running configuration as initial (startup) configuration"""
+        self.write_channel(self.normalize_cmd(cmd))
+        output = (
+            self._get_cmd_output_and_prompt(cmd)
+            if self.global_cmd_verify is not False
+            else ""
+        )
+        return output
```
ktbyers, line 134: Why the need of this private method?

I would have a pretty big bias to doing the following instead:

```python
return self._send_command_str(cmd)
```

I say this as I will forget about this private method and then accidentally break your driver. In other words, it is better to follow more standard Netmiko practices unless we can't make it work (and I am not seeing here why the standard practices wouldn't work).

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def exit_config_mode(self, exit_config: str = "", pattern: str = "") -> str:
+        """Exit the candidate private mode"""
+        output = ""
+        self.write_channel(self.RETURN)
+        prompt = self.read_until_pattern(pattern="]--")
+        matches = re.search(r"\n--{( | \* | \+ | \+\* | \!\+ | \!\* )candidate", prompt)
```
ktbyers, line 141: Can we abstract this * and + processing into a private method that we call and potentially reuse that we somehow return a `state` from indicating what state the device is in with respect to this?

I think I am not following what the `+` sign means--can you expand/clarify on this? I tried to look it up online, but was having difficulty finding references.

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,197 @@
...
+            config_commands=config_commands, exit_config_mode=exit_config_mode, **kwargs
+        )
+
+    def _discard(self) -> str:
+        """Discard changes made in candidate private mode"""
+        log.warning("Uncommitted changes will be discarted!")
+        cmd = "discard stay"
+        self.write_channel(self.normalize_cmd(cmd))
+        output = (
+            self._get_cmd_output_and_prompt(cmd)
```
ktbyers, line 177: Once again pretty strong preference just to use standard Netmiko practice of calling `self._send_command_str(cmd)`.

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,197 @@
...
+            else ""
+        )
+        return output
+
+    def _running_mode(self) -> str:
+        """Enter running mode"""
+        cmd = "enter running"
+        self.write_channel(self.normalize_cmd(cmd))
+        output = (
+            self._get_cmd_output_and_prompt(cmd)
```
ktbyers, line 188: Use `self._send_command_str(cmd)` instead.

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,197 @@
...
+            if self.global_cmd_verify is not False
+            else ""
+        )
+        return output
+
+    def _get_cmd_output_and_prompt(self, cmd: str = "") -> str:
```
ktbyers, line 194: Eliminate and replace any use with `self._send_command_str(cmd)` calls

tests/test_netmiko_config.py
```diff
@@ -20,6 +20,9 @@ def test_enable_mode(net_connect, commands, expected_responses):
     Catch exception for devices that don't support enable
     """
     try:
+        if net_connect.device_type in ["nokia_srl"]:
+            assert pytest.skip()
```
ktbyers, line 24: Shouldn't be necessary the `NoEnable` parent class should handle this properly such that the tests should work properly without any modifications.

ktbyers (commented):
See comments inline. I am happy to help work through these as well.

ktbyers:
@sebastien-guay @hellt Comments above. Let me know if they aren't clear or won't work. 

I wasn't testing against a device when looking at the code so I might have invalid assumptions in my statements.

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,197 @@
...
+        self._test_channel_read()
+        self.ansi_escape_codes = True
+        # Bottom toolbar text not required
+        command = "environment cli-engine type basic"
+        self.write_channel(self.normalize_cmd(command))
+        command = "environment complete-on-space false"
+        self.write_channel(self.normalize_cmd(command))
+        self.set_base_prompt()
+        time.sleep(10 * self.global_delay_factor)
+        self.clear_buffer()
```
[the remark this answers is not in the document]
sebastien-guay: Next thing tot appears is just the prompt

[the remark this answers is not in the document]
ktbyers: Okay, it would be better for the performance of this driver to do the following:

```python
        # Overloading disable_paging which is confusing
        command = "environment cli-engine type basic"
        self.disable_paging(
            command=command,
            pattern=r"#",
        )   
        command = "environment complete-on-space false"
        self.disable_paging(
            command=command, pattern=r"#"
        )  
```

And get rid of the sleep and clear_buffer call.

This is a bit of kludge using disable_paging, but it is fine, I have used it elsewhere (I am just using it to send command and search for a pattern and also for command echo verification).

Let me know if that doesn't work though.

netmiko/base_connection.py
```diff
@@ -1961,6 +1971,7 @@ def config_mode(
         """
         output = ""
         if not self.check_config_mode():
+            self.write_channel(self.RETURN)
```
sebastien-guay, line 1974: I'm proposing this change as an improvement. When I run test the first command never has the prompt in front of it. This is because the it has been read by the check_config_mode(). We see the same behavior in the cisco_ios example where the "configure terminal" has no prompt.

`$ python config_change.py 
Password: 

configure terminal
cisco1(config)#logging buffered 100000
cisco1(config)#end
cisco1#write mem
Building configuration...
[OK]
cisco1#
`

ktbyers:
@sebastien-guay Just let me know when you want to review this again.

sebastien-guay:
> @sebastien-guay Just let me know when you want to review this again.

Thanks @ktbyers! I'm not ready yet. I will tell you.

sebastien-guay:
> @sebastien-guay Just let me know when you want to review this again.

@ktbyers  I'm doing the modification to change the signature of check_config_mode:

`def check_config_mode(
        self, check_string: str = "", pattern: str = "", force_regex: bool = False
    ) -> bool:` 

This method is overridden in 45 classes. Is there a way in python to not change the child class method signature?

sebastien-guay:
@ktbyers Your comments have been addressed. Ready for another check. Thanks for your helpful suggestions!

ktbyers:
Sounds good. Let me review.

sebastien-guay:
> Sounds good. Let me review.

Hi @ktbyers, Would you have time to look at the review beginning of next week?

ktbyers:
@sebastien-guay Let me see what I can do.

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,180 @@
...
+    def session_preparation(self) -> None:
+        self._test_channel_read(pattern=r"#")
+        self.ansi_escape_codes = True
+        # Bottom toolbar text not required
+        commands = [
+            "environment cli-engine type basic",
+            "environment complete-on-space false",
+        ]
+        for command in commands:
+            self.disable_paging(command=command, cmd_verify=False, pattern=r"#")
```
ktbyers, line 65: Does it work properly here with `cmd_verify=True`? It is much better to have cmd_verify=True here unless there is some reason it doesn't work.

netmiko/nokia/nokia_srl.py
```diff
@@ -0,0 +1,180 @@
...
+        The plus sign (+) in the prompt indicates that the running configuration differs
+        from the startup configuration. After you enter the save startup command,
+        the running configuration is synchronized with the startup configuration,
+        and the plus sign is removed from the prompt.
+
+        The exclamation mark (!) in the prompt indicates that another user has commited
+        changes to the running datastore.
+        """
+        matches = re.search(r"\n--{( | \* | \+ | \+\* | \!\+ | \!\* )candidate", prompt)
+        return True if matches and "*" in matches.group() else False
```
ktbyers, line 180: So if you do the `commit stay` that synchronizes candidate to running. Does it also synchronize running to startup or do you separately need to execute `save startup` for that?

netmiko/nokia/nokia_srl.py
sebastien-guay: > Does it work properly here with `cmd_verify=True`? It is much better to have cmd_verify=True here unless there is some reason it doesn't work.

You can update with the following piece of code and it works. I tried to update the new branch but I don't have the right.

`commands = [
            r"environment complete-on-space false",
            r"environment cli-engine type basic",
        ]
        for command in commands:
            self.disable_paging(command=command, cmd_verify=True, pattern=r"#")`

ktbyers: Sounds good. I will update that now.

ktbyers:
Fixing strange GitHub contributor issue.

I will fix the outstanding issues after merge.

PR #2810, opened 2022-06-02
Mikrotik duplicate command echo

ktbyers:


ktbyers:
@elavaud I tested it pretty well with the output you provided in the previous solution so I think it works properly. Let me know if you see issues though.

elavaud:
It indeed works properly. Thanks!

PR #2820, opened 2022-06-20
Nokia SRL Driver

ktbyers:


ktbyers:
@sebastien-guay FYI, I made a few minor updates to the PR and just referenced the outstanding questions from the previous PR here.

sebastien-guay:
``> So there are two outstanding questions:
> 
> 1. cmd_verify=True question on session_preparation()
> 2. The question about commit and the running/startup config

My answers

1. You can update session_preparation() with the following piece of code and it works. I tried to update the new branch but I don't have the right.
```
commands = [
            r"environment complete-on-space false",
            r"environment cli-engine type basic",
        ]
        for command in commands:
            self.disable_paging(command=command, cmd_verify=True, pattern=r"#")
```

2. You are right. `save startup` is required to synch to startup. `commit save` exists to sync to startup right away but I chose to use commands performing only one action.

netmiko/nokia/nokia_srl.py
```diff
@@ -50,7 +50,7 @@ def session_preparation(self) -> None:
             "environment complete-on-space false",
```
sebastien-guay, line 50: Change with:
```
commands = [
            r"environment complete-on-space false",
            r"environment cli-engine type basic",
        ]
        for command in commands:
            self.disable_paging(command=command, cmd_verify=True, pattern=r"#")
```
Order of the command seems important because it was failing if environment complete-on-space false is executed last.

ktbyers: Okay fixing...

ktbyers:
@sebastien-guay Does everything look good and do the tests pass now?

sebastien-guay:
@

> @sebastien-guay Does everything look good and do the tests pass now?

All good!

sebastien-guay:
@ktbyers Could you add me as a contributor please? We will use the SR Linux driver with https://docs.openstack.org/networking-generic-switch/latest/ which is using netmiko. It would be easier if changes are required if I'm a contributor.

ktbyers:
@sebastien-guay 

Hmmm weird. I don't know what is going on with GitHub there.

Let me dig into it.

ktbyers:
I force removed this Pull Request from the `develop` branch and will correct in separate PR(s)

PR #2842, opened 2022-06-29
set self.global_cmd_verify=False only if None for Classic CLI in Nokia SROS

laimaretto:
We want to honor the cmd_verify definition under the Connecthandler. Only set to False, if unset.


ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/2856

For some reason, github would not let me rebase to fix the CI-CD issue so I had to do it from the command-line and then push up a new pull-request.

But your change is merged there...

PR #2865, opened 2022-07-20
Updating Audiocode driver with test results.

Gatorjosh14:


ktbyers:
@Gatorjosh14 Can you explain when you would use each of these three drivers:

```
	"audiocode_base" : AudiocodeBaseSSH,
	"audiocode_66" : Audiocode66SSH,
	"audiocode_shell" : AudiocodeShellSSH, 
```

Just want to make sure I understand the devices you would use each of them on.

Gatorjosh14:
> @Gatorjosh14 Can you explain when you would use each of these three drivers:
> 
> ```
> 	"audiocode_base" : AudiocodeBaseSSH,
> 	"audiocode_66" : Audiocode66SSH,
> 	"audiocode_shell" : AudiocodeShellSSH, 
> ```
> 
> Just want to make sure I understand the devices you would use each of them on.

Sure thing, it's dependent upon the Model and Firmware version.  For example, most Audiocode M500 devices would run the 7.2 firmware and therefore use the AudiocodeBaseSSH.  M800/M600 devices would likely be on the 6.6 firmware which would use Audiocode66SSH, and lastly some M600 and Media Pack devices would use the AudiocodeShellSSH.

netmiko/audiocode/audiocode_ssh.py
```diff
@@ -1,363 +1,735 @@
...
+		config_commands = None,
+		exit_config_mode: bool = True,
+		read_timeout: float = None,
+		delay_factor: float = 1.0,
+		max_loops: int = 150,
+		strip_prompt: bool = False,
+		strip_command: bool = False,
+		config_mode_command: str = None,
+		cmd_verify: bool = True,
+		enter_config_mode: bool = False,
```
ktbyers, line 174: Why is this `enter_config_mode=False`?

Gatorjosh14: Because there is no universal configuration mode on these devices.  I think we spoke about this before, but you can go into different config modes, such as:  config network, config data, config system, config voip, config troubleshooting.  Therefore, by default you would want that de-active so the user would choose the configuration mode they want to go into.  If you want to choose a default value, you could and the user could override the parameter, but not sure what that would serve.

ktbyers:
FYI, I am working on fixing the linting and type hint issues.

netmiko/audiocode/audiocode_ssh.py
ktbyers: Yes, understood...but you always need to pick one of those i.e. it always needs to enter config mode (using some command)--correct?

We can just put a check that `config_mode_command is not None` and raise and exception if it is.

Gatorjosh14:
> Sounds good, afterwards I was thinking I could have helped better there by importing typing so that most of it would have transitioned.  Thanks for doing that.

netmiko/audiocode/audiocode_ssh.py
Gatorjosh14: That is correct, you'd have to choose a config option to go into the appropriate mode.  I typically just include the "config system" in my base script, otherwise it would fail to enter config mode anyways.

ktbyers:
Closing as superseded by #2869

Issue #2084, opened 2020-12-23
Cisco WLC driver is leaving stale Paramiko threads in Netmiko >3.0.0


pwiddicombe:
Not sure exactly which version this cropped up.  At version 3.0.0 and lower, using cisco_wlc model, when the .disconnect() method is called (directly, or indirectly via the "with" method), the underlying Paramiko thread is correctly closed.

3.3.0 and 3.3.2 fail to close the Paramiko thread when cisco_wlc is used (directly via disconnect() or implicitly by a completed "with" block); although works as expected with model cisco_ios.
(This may be related to issue #2080).  If connecting to multiple WLC's, you will get one stranded underlying Paramiko thread for each WLC you have connected to.

pwiddicombe:
Tested with Netmiko 3.3.3.  The issue still persists with device type cisco_wlc - Paramiko thread left open after connection closure.

ktbyers:
@pwiddicombe You need to post your code particularly how you are ending your SSH session.

pwiddicombe:
from netmiko import Netmiko
import threading
import getpass
import time

userid = getpass.getuser()
userpwd = getpass.getpass(f"Enter password for {userid}: ")

conn = Netmiko(
    device_type="cisco_wlc",
    ip="172.22.122.247",
    username=userid,
    password=userpwd,
    secret="cisco",
)
resp = conn.send_command("show certificate summary")
print("Before disconnect")
print(threading.enumerate())
print(resp)
print("After disconnect")
conn.disconnect()
print(threading.enumerate())
time.sleep(5)
print(threading.enumerate())

=========================== Output
Enter password for pw32115: 
Before disconnect
[<_MainThread(MainThread, started 267572)>, <Thread(SockThread, started daemon 261988)>, <paramiko.Transport at 0x32de3f0 (cipher aes128-ctr, 128 bits) (active; 1 open channel(s))>]
Web Administration Certificate................... 3rd Party
Web Authentication Certificate................... Locally Generated
Certificate compatibility mode:.................. off
Lifetime Check Ignore for MIC ................... Disable
Lifetime Check Ignore for SSC ................... Disable

After disconnect
[<_MainThread(MainThread, started 267572)>, <Thread(SockThread, started daemon 261988)>, <paramiko.Transport at 0x32de3f0 (cipher aes128-ctr, 128 bits) (active; 0 open channel(s))>]
[...]

PR #2875, opened 2022-07-28
Added sleep to wait for session closure on old wlc

mfr-6:
This fix solving issue with Netmiko, where Paramiko thread is not closed for WLCs with older software version.

More details here:
https://github.com/ktbyers/netmiko/issues/2084


ktbyers:
Linting issue is due to problems that already have been fixed (should be fixed by rebase).

PR #2876, opened 2022-07-28
Updating Fortinet Driver for 4.1.1 and Multivdom compatibility

Gatorjosh14:


ktbyers:
@Gatorjosh14 You should probably rebase your pull request with Netmiko's develop branch as it looks like you have a conflict with some of the files in Netmiko (related to your audiocode driver).

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		output = self.send_config_set(
+			config_commands = disable_paging_commands,
+			exit_config_mode = True,
+			read_timeout = None,
+			delay_factor = 1.0,
+			max_loops = 150,
+			enter_config_mode = False,
+		)
+		log.debug("***Window Paging Disabled***")
+		return output
```
ktbyers, line 232: So I assume this depends on VDOM support? Also what happens here on failure, does an exception get generated?

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		) -> str:
+		"""Disable paging is only available with specific roles so it may fail.
+
+		:param command: Device command to disable pagination of output
+
+		:param delay_factor: Deprecated in Netmiko 4.x. Will be eliminated in Netmiko 5.
+		"""
+
+		disable_paging_commands = [
+			"config global",
```
ktbyers, line 218: We probably need to discuss in the context of send_config_set() config_mode behavior.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+			"end"
+			]
+		output = self.send_config_set(
+			config_commands = disable_paging_commands,
+			exit_config_mode = True,
+			read_timeout = None,
+			delay_factor = 1.0,
+			max_loops = 150,
+			enter_config_mode = False,
+		)
```
ktbyers, line 230: Some of these arguments shouldn't be set (so we should discuss these when we discuss send_config_set)

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		# Return paging state
+		enable_paging_commands = ["config global", "config system console", "set output more", "end"]
+		# Should test output is valid
+		output = self.send_config_set(
+			config_commands = enable_paging_commands,
+			exit_config_mode = True,
+			read_timeout = None,
+			delay_factor = 1.0,
+			max_loops = 150,
+			enter_config_mode = False,
```
ktbyers, line 245: Same send_config_set argument issue.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+				self.send_command(command_string="config global", expect_string=r"\(global\) # ")
+			except:
+				raise ValueError("Device Failed to Enter Global Mode.")
+		else:
+			log.debug("***Entering VDOM Mode***")
+			try:
+				self.send_command(command_string="config vdom", expect_string=r"\(vdom\) # ")
+				self.send_command(command_string=f"edit {mode_name:}", expect_string=fr"\({mode_name}\) # ")
+			except:
+				raise ValueError(f"Device Failed to Enter '{mode_name}' VDOM Mode.")
```
ktbyers, line 281: When is this used? We should probably decouple these two items and create two separate methods (or just one if only one is needed).

netmiko/nokia/nokia_sros.py
```diff
@@ -53,9 +53,7 @@ def session_preparation(self) -> None:
         else:
             # Classical CLI has no method to set the terminal width nor to disable command
             # complete on space; consequently, cmd_verify needs disabled.
-            # Only disabled if not set under the ConnectHandler.
-            if self.global_cmd_verify is None:
-                self.global_cmd_verify = False
+            self.global_cmd_verify = False
```
ktbyers, line 56: Looks like a rebasing issue (i.e. undoing a change that was previously merged into the develop branch.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		while mode_active:
+			output = self.find_prompt()
+
+			if ") #" in output:
+				self.write_channel("end")
+			elif ") #" not in output or loops == 5:
+				break
+
+			time.sleep(1)
+			loops += 1
```
ktbyers, line 298: Once again probably should be two separate methods.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+			pattern=pattern,
+		)
+
+	def send_config_set(
+		self,
+		config_commands = None,
+		exit_config_mode: bool = True,
+		read_timeout: float = None,
+		delay_factor: float = 1.0,
+		max_loops: int = 150,
```
ktbyers, line 97: `delay_factor` and `max_loops` should both default to None. They are largely deprecated.

ktbyers (commented):
Also there are some mypy type hint issues.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		data = self._test_channel_read(pattern="to accept|[#$]")
+		# If "set post-login-banner enable" is set it will require you to press 'a'
+		# to accept the banner before you login. This will accept if it occurs
+		if "to accept" in data:
+			self.write_channel("a\r")
+			self._test_channel_read(pattern=r"[#$]")
+
+		self.set_base_prompt(alt_prompt_terminator="$")
+		self.disable_paging()
+		self.clear_buffer()
```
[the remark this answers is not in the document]
Gatorjosh14: I don't if it's really "needed", but I just like having a clear buffer before proceeding.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		"""Prepare the session after the connection has been established."""
+
+		data = self._test_channel_read(pattern="to accept|[#$]")
+		# If "set post-login-banner enable" is set it will require you to press 'a'
+		# to accept the banner before you login. This will accept if it occurs
+		if "to accept" in data:
+			self.write_channel("a\r")
+			self._test_channel_read(pattern=r"[#$]")
+
+		self.set_base_prompt(alt_prompt_terminator="$")
```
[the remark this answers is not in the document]
Gatorjosh14: Agreed, can remove that portion.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		self.disable_paging()
+		self.clear_buffer()
+
+	def set_base_prompt(self, 
+		pri_prompt_terminator="#", 
+		alt_prompt_terminator="$", 
+		delay_factor=1.0,
+		pattern = r"(#|\$)"
```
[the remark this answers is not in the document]
Gatorjosh14: I'd be fine with that, why do you think it would cause an issue?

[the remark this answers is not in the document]
Gatorjosh14: Done

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		:type delay_factor: int
+
+		:param pattern: Regular expression pattern to determine whether prompt is valid
+
+		:param confirm_pattern: Regular expression pattern to confirm prompt was found by auto discovery.
+		"""
+		return super().find_prompt(
+			delay_factor=delay_factor, 
+			pattern=pattern,
+		)
```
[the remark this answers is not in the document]
Gatorjosh14: Done

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: We'll probably have to do something similar like we did with Audiocodes since Fortigates don't have a traditional configuration mode.  You basically type something like "config system global" or "config system snmp user" then that takes into a hierarchy to make configs there.  There's no "config mode" that can be define as a constant.  Also, if multi-vdom mode is enabled, you first have to go into the vdom you want to change, like "config vdom" > "edit root" > then "config router static".  So it's best to leave flexibility like we did on Audiocodes.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: Well, they're need for multi-vdom mode if you don't want the user to define it in the "send_config_set" commands as I explained above.  If the device is single-vdom mode, then these methods aren't needed, it's contextual.  I've personally always handle entering and exiting the vdom like a configuration mode and do it in my "send_config_set" script.  I only developed this so it could pass the pytest scripts.

Gatorjosh14: Getting more maybe at your intended point of separating them.  The global mode becomes it's own thing like a vdom, it's difficult to explain to someone if they haven't used a Fortigate, but here's my attempt.

If you wanted to change something for snmp for single-vdom, everything is contained with root, by default so it's pretty simple:
"config system snmp user"

However, if multi-vdom mode is active, then you then sets certain configuration under "global" and some under "root" and then some under whatever additional vdom's you created like "management".  Global configs apply to all vdoms and are mostly system related configs.  Root is also device management related, but nothing in root has a relationship to another vdom.  There's almost like a VRF, but more like a Virtual Domain (hence vdom), it has it's own routing table and firewall policies.  Like different customers for example.

Now, I included them in the same Method because they are similar as you can see, but it makes it easier in my opinion to have 1 dedicated Method to enter into global or vdom, less commands to type overall.  so you could type _enter_global_or_vdom("global") or you could type _enter_global_or_vdom("root").   If you want to separate them, I'm fine with that too, was just my original thought process.

netmiko/nokia/nokia_sros.py
Gatorjosh14: Probably.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: If you separate them, it would be kind of redundant here as it's the same code, but if adds ease of use, I'm fine with that.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: Done

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		config_commands = None,
+		exit_config_mode: bool = True,
+		read_timeout: float = None,
+		delay_factor: float = 1.0,
+		max_loops: int = 150,
+		strip_prompt: bool = False,
+		strip_command: bool = False,
+		config_mode_command: str = None,
+		cmd_verify: bool = True,
+		enter_config_mode: bool = False,
```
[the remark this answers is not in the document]
Gatorjosh14: See my further comments below where I talk about how it's like an Audiocode device in some respects.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		error_pattern: str = "",
+		terminator: str = r"",
```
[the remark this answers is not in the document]
Gatorjosh14: I think I was having issues with it, trying to remember.  Would we apply the "self.prompt_pattern" here?

if not terminator:
     terminator = self.prompt_pattern

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+			delay_factor=delay_factor,
+			max_loops=max_loops,
+			strip_prompt=strip_prompt,
+			strip_command=strip_command,
+			config_mode_command=config_mode_command,
+			cmd_verify=cmd_verify,
+			enter_config_mode=enter_config_mode,
+			error_pattern=error_pattern,
+			terminator=terminator,
+			bypass_commands=bypass_commands
```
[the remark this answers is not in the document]
Gatorjosh14: As mentioned elsewhere, we likely require an Audiocode type solution.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+		)
+
+	def enable(self, cmd="", pattern="", re_flags=re.IGNORECASE):
+		"""Not Supported for this platform"""
+		pass
```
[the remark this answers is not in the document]
Gatorjosh14: There is no enable mode on Fortigates.  You pretty much get the privilege level you authenticate at based on your profile, you can't elevate it.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+
+		:param check_string: Identification of configuration mode from the device
+		:type check_string: str
+
+		:param pattern: Pattern to terminate reading of channel
+		:type pattern: str
+		"""
+		return super().check_config_mode(
+			check_string=check_string, pattern=pattern, force_regex=force_regex
+		)
```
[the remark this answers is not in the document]
Gatorjosh14: Discussed elsewhere about an Audiocode type solution.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: The command "config global" would fail if the device was not in multi-vdom mode, but the remaining commands are ran with no issue.  It doesn't cause any problem so saw no issue to do any kind of predetermination of what vdom mode it was in.  If you don't like that solution, we can do like the former and discover if the device is in multi-vdom mode and have different commands to run depending on that result.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: Yeah, those are discussed in other threads.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: Fixed that already.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -1,104 +1,298 @@
...
+			"diffie-hellman-group14-sha1",
+			"diffie-hellman-group-exchange-sha1",
+			"diffie-hellman-group-exchange-sha256",
+			"diffie-hellman-group1-sha1",
+		)
+
+	def session_preparation(self) -> None:
+		"""Prepare the session after the connection has been established."""
+
+		data = self._test_channel_read(pattern="to accept|[#$]")
```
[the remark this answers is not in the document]
Gatorjosh14: Done

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
ktbyers: The issue is that I later re-wrap `pattern` in another set of parenthesis (in another logical-or) and doing that causes problems in the regex behavior. One workaround to this is to use the square-bracket so that there is only one set of parenthesis.

netmiko/fortinet/fortinet_ssh.py
ktbyers: Okay, we can do what we did in audiocode (i.e. require a `config mode` command, and raise an error if the user doesn't provide it). So `enter_config_mode = True` would be the default.

netmiko/fortinet/fortinet_ssh.py
ktbyers: I will propose an alternate solution for this. I am pretty sure we can create an alternate solution for the test code problem which will be more maintainable in the long run.

ktbyers:
Fortinet issue that we make sure we address in this PR:

https://github.com/ktbyers/netmiko/issues/2946

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
ktbyers: I am not seeing why you would need to override the default `#` so I would just revert it back to that (and then see what issue we run into).

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
ktbyers: If there is no enable mode, then we should use the standard Netmiko `NoEnable` helper class as follows:

```python
class FortinetSSH(NoEnable, BaseConnection):
```

And then remove all references to the `enable` and `exit_enable_mode` code here.

netmiko/fortinet/fortinet_ssh.py
ktbyers: So for audiocode we required:

```
enter_config_mode: bool = True,
```

And then left:

```
config_mode_command: Optional[str] = None,
```

And then in the code checked that `config_mode_command` MUST be set to something or Netmiko we will raise an exception, we can do that again here...that is fine with me.

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
Gatorjosh14: Well, you have to remember that the prompt for a configuration could either contain # or $, so if we want to retain something there, it either needs to be self.prompt_pattern or a [$#] solution.

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
Gatorjosh14: Okay, will do.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: Sounds good, I'll do a draft with the changes and submit the PR.

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: Have you come up with a solution for this yet or does my proposed solution work?

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: Are okay with this approach, letting the command fail for single-vdom devices since it doesn't break anything?

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
ktbyers: Okay, makes sense...so yes then probably `self.prompt_pattern` which would translate to:

```
terminatoror: str = r"",
```

In the method arguments and then adding:

```python
if not terminator:
    terminator = self.prompt_pattern
```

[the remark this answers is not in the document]
ktbyers: In other words, set it to null-value in the arguments and then check for the null-value inside the method...and if null, set it to self.prompt_pattern.

netmiko/fortinet/fortinet_ssh.py
ktbyers: Yes, I will fix. I haven't done it yet, but I will.

netmiko/fortinet/fortinet_ssh.py
ktbyers: So just verifying this--with VDOM support, the below commands all work?

```
			"config global",
			"config system console",
			"set output standard",
			"end"
```

With no VDOM support, the first command fails with an error, but then the remaining three commands work?

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
Gatorjosh14: Okay, removed.

netmiko/fortinet/fortinet_ssh.py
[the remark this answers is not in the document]
Gatorjosh14: Does this work?

```
def send_config_set(
		self,
		config_commands = None,
		exit_config_mode: bool = True,
		read_timeout: float = None,
		delay_factor: float = None,
		max_loops: int = None,
		strip_prompt: bool = False,
		strip_command: bool = False,
		config_mode_command: str = None,
		cmd_verify: bool = True,
		enter_config_mode: bool = True,
		error_pattern: str = "",
		terminator: str = r"",
		bypass_commands: str = None,
		) -> str:

		if not terminator:
			terminator = self.prompt_pattern

		if enter_config_mode and config_mode_command is None:
			msg = """
send_config_set() for the Fortinet drivers require that you specify the
config_mode_command. For example, config_mode_command="config system global".
			"""
			raise ValueError(msg)
			
		return super().send_config_set(
			config_commands=config_commands,
			exit_config_mode=exit_config_mode,
			read_timeout=read_timeout,
			delay_factor=delay_factor,
			max_loops=max_loops,
			strip_prompt=strip_prompt,
			strip_command=strip_command,
			config_mode_command=config_mode_command,
			cmd_verify=cmd_verify,
			enter_config_mode=enter_config_mode,
			error_pattern=error_pattern,
			terminator=terminator,
			bypass_commands=bypass_commands
		)
```

netmiko/fortinet/fortinet_ssh.py
Gatorjosh14: See the terminator thread, posted the updated config there.

ktbyers:
Mostly implemented in a separate PR (still need to do send_config_set work).

PR #2883, opened 2022-08-03
updating to new audiocode driver with updated tests

Gatorjosh14:


Gatorjosh14:
pulled incorrect driver

PR #2914, opened 2022-08-23
Add Cisco SG200 device type

hackvanru:
This connection class writes for low cost switches SG200 series, in which there is no command
    ip ssh password-auth
    and username promted twice

    The switchs does not support any SSH authentication methods.


ktbyers:
@hackvanru Did you look into SSHClient_noauth?

See:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/ssh_auth.py

Also see the hp_procurve driver for an example of using this:

https://github.com/ktbyers/netmiko/blob/develop/netmiko/hp/hp_procurve.py


Using that I think we can avoid overriding the establish_connection() method (and this is the standard way of implementing no_auth in Netmiko/Paramiko for devices that force you into a non-standard SSH login). There are other examples in Netmiko besides HP Procurves.

hackvanru:
Thank you for advice. I replaced establish_connection() method by _build_ssh_client() with class SSHClient_noauth.

ktbyers:
Code is broken here:

```
Run pylama .
netmiko/cisco/cisco_s200.py:29:35 E0602 undefined name 'path' [pyflakes]
Error: Process completed with exit code 1.
```

Also should make sure we run Netmiko's standard test suite on this:

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3171

PR #2931, opened 2022-09-02
nokia_sros: Disable paging in both CLI modes, independent of initial …

melewitz:
The previous driver disabled paging in both the Classic and MD CLI modes when the initial login mode is model driven (MD), but only for the Classic CLI when it's the initial mode.  This patch disables paging in both modes no matter which is the initial mode.  

This is necessary because "//" switches modes, and "//cmd" executes the command in the 'other' mode.  Users commonly require commands in both modes and therefore need paging disabled in both modes.


ktbyers:
LGTM

PR #2957, opened 2022-09-21
Mikrotik remove config mode

k-ribot:
Remove useless code to manage config mode to MikroTik router

### DIFF
--- a/netmiko/mikrotik/mikrotik_ssh.py
+++ b/netmiko/mikrotik/mikrotik_ssh.py
@@ -3,19 +3,18 @@ import re
 import os
 
 from netmiko.no_enable import NoEnable
+from netmiko.no_config import NoConfig
 from netmiko.cisco_base_connection import CiscoSSHConnection
 from netmiko.base_connection import BaseConnection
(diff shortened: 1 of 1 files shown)


netmiko/mikrotik/mikrotik_ssh.py
```diff
@@ -124,6 +101,21 @@ def set_base_prompt(
...
+        **kwargs: Any
+    ) -> str:
+        """No config mode on MikroTik"""
+        return super().send_config_set(
+            config_commands=config_commands,
+            exit_config_mode=exit_config_mode,
+            enter_config_mode=enter_config_mode,
+            **kwargs
+        )
+
```
ktbyers, line 118: This shouldn't be necessary...the `NoConfig` class should automatically make the default `send_config_set()` parent method work (without changing any of the send_config_set arguments).

ktbyers:
@kribotUnyc Good idea.

ktbyers:
Looks like Python black is failing.

You should be able to fix that by just executing `black ./netmiko` at the base of the Netmiko repository (and then saving the changes that black makes).

PR #2958, opened 2022-09-23
Added support for Arris CER and Casa CMTS platforms.

vangesseld:
Greetings, 

Looking to add support for Arris and Casa CMTS platforms. Please let me know if you have any feedback. 

Test results (arris_cer):
`(venv) PS C:\Users\VangesselD\PycharmProjects\netmiko\tests> py.test -v test_netmiko_show.py --test_device arris_cer
========================================================================================================= test session starts =========================================================================================================
platform win32 -- Python 3.9.13, pytest-7.1.2, pluggy-1.0.0 -- C:\Users\VangesselD\PycharmProjects\netmiko\venv\Scripts\python.exe
cachedir: .pytest_cache
rootdir: C:\Users\VangesselD\PycharmProjects\netmiko, configfile: setup.cfg
plugins: pylama-8.3.8
collected 25 items

test_netmiko_show.py::test_failed_key SKIPPED (Not using SSH-keys)                                                                                                                                                               [  4%]
test_netmiko_show.py::test_disable_paging PASSED                                                                                                                                                                                 [  8%]
test_netmiko_show.py::test_terminal_width PASSED                                                                                                                                                                                 [ 12%] 
test_netmiko_show.py::test_ssh_connect PASSED                                                                                                                                                                                    [ 16%]
test_netmiko_show.py::test_ssh_connect_cm PASSED                                                                                                                                                                                 [ 20%]
test_netmiko_show.py::test_send_command_timing PASSED                                                                                                                                                                            [ 24%]
test_netmiko_show.py::test_send_command_timing_no_cmd_verify SKIPPED                                                                                                                                                             [ 28%] 
test_netmiko_show.py::test_send_command PASSED                                                                                                                                                                                   [ 32%]
test_netmiko_show.py::test_send_command_no_cmd_verify SKIPPED                                                                                                                                                                    [ 36%] 
test_netmiko_show.py::test_complete_on_space_disabled SKIPPED                                                                                                                                                                    [ 40%]
test_netmiko_show.py::test_send_command_textfsm SKIPPED (TextFSM/ntc-templates not supported on this platform)                                                                                                                   [ 44%] 
test_netmiko_show.py::test_send_command_ttp SKIPPED (TTP template not existing for this platform)                                                                                                                                [ 48%] 
test_netmiko_show.py::test_send_command_genie SKIPPED (Genie not supported on this platform)                                                                                                                                     [ 52%] 
[...]

PR #2967, opened 2022-09-30
Remove Arris and Casa from SSH autodetect

ktbyers:
Temporary placeholder for this.

I think these two platforms need to be more specific in the SSH autodetect.


ktbyers:
@vangesseld FYI, temporary placeholder for this SSH autodetect issue (i.e. to make each of them more specific).

vangesseld:
@ktbyers 

I will take a look at this and see if I can work out something better. I agree that it's probably too risky as is now that you mention it.

vangesseld:
Hi @ktbyers,

I looked at this taking into account your feedback on #2958:

> Can we qualify these a bit more (in the case of CER add something that references Arris to the string and in the case of Casa add something to make it CMTS specific?

For Arris, I can not for the life of me find a command that actually mentions the vendor in the output. I think we might be able to be specific enough using the `show chassistype` command, which provides one line of output that reads `Chassis type: E6000`

The only reason I think this is potentially viable is that currently, the Arris CER software only runs on E6000 chassis, and an exact match like `'^Chassis type: E6000$'` will probably prevent false positives. 

They do have similar software/products in the form of the C4 and C4c chassis, but those are pretty old and all the ones I have had access to have been retired. The driver might work for those, but I couldn't test it on those. Additionally, limiting the auto detect to only the E6000 chassis doesn't strike me as particularly future proof. 

For Casa, that's another ball of wax. The show version output is very brief on these, and the items in the show version that are CMTS specific are also specific to the processor module installed (so we'd be matching on part numbers). I think that's probably a can of worms that no one is likely interested in opening. 

Given that these are admittedly pretty niche products (that you won't ever see outside of a DOCSIS network), and the platforms themselves really aren't giving us what we're needing here, it might be best to merge this PR. I personally don't make heavy use of the autodetect feature, I mostly included it to help out my colleagues, but we have other ways we can identify these outside of leveraging autodetect.

Thank you!

ktbyers:
Remove these from autodetect (as the patterns searched for are too broad, unfortunately).

PR #2977, opened 2022-10-07
Updates to maipu driver

alizee2020:
Hi Ktbyers,
    Thanks so much for your advice. I have made some change in maipu dirver. Please review again.


ktbyers:
Replaced by #2992

Issue #2987, opened 2022-10-12
NetmikoAutheticationError seen due to lower delay in Cisco devices


viholla:
Due to less delay between loops, read_channel is not capturing the buffer. hence delay need to be introduced between loops.

Also cisco's new security mandate has made sure that enable must be configured before saving the config and returning to prompt. Hence this needs to be handled in cisco_base_connection.py

viholla:
Trying to commit fix for this but i'm denied to raise any PR:

(base) VIHOLLA-M-350K:netmiko viholla$ git push origin cisco_netmiko
ERROR: Permission to ktbyers/netmiko.git denied to viholla.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

Toxic-Waste-:
> Trying to commit fix for this but i'm denied to raise any PR:
> 
> (base) VIHOLLA-M-350K:netmiko viholla$ git push origin cisco_netmiko ERROR: Permission to ktbyers/netmiko.git denied to viholla. fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights and the repository exists.

I am not an expert on the matter, but if you want to submit fixes to the repository, you need to make a new branch and ask for a merge request, the netmiko endboss needs to sign off on it :)

ktbyers:
@viholla Have you tried setting `fast_cli=False` as an argument to ConnectHandler?

Yes, in order to push a change in Git/GitHub, you need to fork the Netmiko repository in GitHub and then push the change to your copy of Netmiko.

Then from GitHub you should be able to make a pull-request (so that I see your proposed change).

viholla:
Thanks Kirk. I have raised the PR for the changes. Kindly look into this.

PR #2990, opened 2022-10-13
Fixed delay issue in Cisco devices and enable password prompt handling 

viholla:
https://github.com/ktbyers/netmiko/issues/2987

### DIFF
--- a/netmiko/cisco_base_connection.py
+++ b/netmiko/cisco_base_connection.py
@@ -145,6 +145,25 @@ class CiscoBaseConnection(BaseConnection):
                             time.sleep(2 * delay_factor)
                             count += 1
 
+                    # Support setting up secret from terminal server (Qwerty12345 is not saved)
+                    if re.search(r"Enter enable secret: ", output):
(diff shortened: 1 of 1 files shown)


viholla:
@ktbyers Please help me here. My team is waiting for this fix.

ktbyers:
@viholla You can incorporate any fix you want in your code. In other words, you already have the Netmiko code and your fix (so you can incorporate or not incorporate, but it doesn't depend on me approving it).

Can you explain to me the underlying issue you are trying to fix? If you are trying to fix an initial configuration problem that requires accessing the device via a console/terminal server, then this is probably not the location to fix the issue.

viholla:
@ktbyers So cisco devices have implemented a new rule where configuring enable secret is mandatory for any new device bring up, we can't leave it empty. Once configured, it can be saved or discarded and return to the prompt. To handle this ,I have introduced this new fix. If there is any other way to handle the prompt, please let me know.

ktbyers:
I am just going to close this as this doesn't look like the right place for this fix. Though if you have more information on the output and situation where it happens, then I can possibly see how you could potentially handle it.

PR #3011, opened 2022-10-29
Mikrotik license handling

ktbyers:


ktbyers:
@k-ribot Do you have a way to test this (with "software license" message).

I believe this should behave the same as the PR that you created.

I removed the `test_channel_read`  from `session_preparation` as that has already happened in `special_login_handler()`. Also now that has been read on some platforms you would have an issue since the prompt does not appear again (i.e. it already has been read so the `test_channel_read` would fail).

I am going to go ahead and merge this, but if you see issues with it (or run into problems), just let me know.

PR #3021, opened 2022-11-06
Add driver for Bintec/Funkwerk BOSS

mnhauke:


netmiko/bintec/bintec_boss.py
```diff
@@ -0,0 +1,40 @@
...
+        self,
+        cmd: str = "cmd=save",
+        confirm: bool = False,
+        confirm_response: str = "",
+    ) -> str:
+        """Save Config to flash."""
+        return super().save_config(
+            cmd=cmd, confirm=confirm, confirm_response=confirm_response
+        )
+        return self._send_command_timing_str(command_string=cmd)
```
ktbyers, line 23: @mnhauke Why do you need to send the `cmd=save` command twice? Once in the super() call and once in the subsequent send_command_timing() call?

ktbyers:
I reviewed this PR and it is all good to go except for the one question to @mnhauke regarding the duplicate command.

netmiko/bintec/bintec_boss.py
ktbyers: Never mind...I am just going to remove the duplicate `return` statement since it would never be executed anyways.

ktbyers:
Superseded by https://github.com/ktbyers/netmiko/pull/3576

Issue #3017, opened 2022-11-02
ssh_dispatcher.py ConnLogOnly() throwing exception


windmite:
ssh_dispatcher.py lines 414-421

```
except NetmikoTimeoutException as e:
        if "DNS failure" in str(e):
            msg = f"Device failed due to a DNS failure, hostname {hostname}"
        elif "TCP connection to device failed" in str(e):
            msg = f"Netmiko was unable to reach the provided host and port: {hostname}:{port}"
            msg += f"\n\n{str(e)}"
        logger.error(msg)
        return None
```

Unfortunately I wasn't able to capture the string that brought me to this bug (and it was a timeout error I've only gotten a couple times so difficult to reproduce, I've added some code to grab it if it every happens again) but I'm assuming this needs to be something like:

```
except NetmikoTimeoutException as e:
        if "DNS failure" in str(e):
            msg = f"Device failed due to a DNS failure, hostname {hostname}"
        elif "TCP connection to device failed" in str(e):
            msg = f"Netmiko was unable to reach the provided host and port: {hostname}:{port}"
            msg += f"\n\n{str(e)}"
        else:
            msg = f"Unknown Timeout Error device timed-out: {hostname}:{port}"
            msg += f"\n\n{str(e)}"
        logger.error(msg)
```

I was getting 'msg' referenced before assignment because presumably neither of those strings was in str(e)  Big fan of ConnLogOnly() by the way.

windmite:
Captured the exception string:
Paramiko: 'No existing session' error: try increasing 'conn_timeout' to 15 seconds or larger.

ktbyers:
@windmite Yeah, that msg referenced before assignment is a bug and needs fixed like you described.

Do you want to submit a PR on this?

windmite:
I'm up for it if you'll excuse any mistakes I make in the process. I'll get some tutorials in tonight.

ktbyers:
Fixed here:

https://github.com/ktbyers/netmiko/pull/3055

PR #3025, opened 2022-11-10
Fix for Issue #3017

windmite:
Added an elif to catch no existing session and and else to catch any others.


ktbyers:
@windmite Thanks for doing this.

I did this fix slightly different here. Basically I thought in these cases that it is better to just raise the original exception string.

Issue #3027, opened 2022-11-10
IPV6 support for SNMPDetect


kacem-expereo:
Hi Team!
The snmp_autodetect.py providing netmiko's SNMPDetect do not seem to support ipv6 addresses passed to the hostname parameter.

snmp_autodetect uses UdpTransportTarget by default , which works for ipv4 only.
We can use Udp6TransportTarget for ipv6 instead of UdpTransportTarget.
I propose to identify the host ip version using ipaddress.ip_address().version from ipaddress library and to switch between UdpTransportTarget and Udp6TransportTarget accordingly.
I tested it successfully.
If the idea looks ok for you let me know and I will submit a PR.

Kindly,

ktbyers:
@kacem-expereo Yeah, that sounds fine to me.

kacem-expereo:
Thanks a lot :)

PR #3028, opened 2022-11-14
 IPV6 support for SNMPDetect(#3027) 

kacem-expereo:
Hi Team!
Adding usage of Udp6TransportTarget when the device hostname is IPv6 as mentioned in issue #3027.
It felt cleaner to do it in the init rather than in the _get_snmpv2c and _get_snmpv3 functions.
Tested on a few hundred of devices without issue.
I don't see any test for SNMPDetect, should I prepare any?
Kindly,


ktbyers:
Looks good to me.

PR #3034, opened 2022-11-22
fix: Use conn_timeout for telnet connection

elavaud:
Hi @ktbyers,

Doesn't it make more sense to use the `conn_timeout` rather than `timeout` for the telnet connection?
On SSH connection it is already the `conn_timeout` which is used.


elavaud:
Hi @ktbyers , any news on this one? :)

ktbyers:
@k-ribot I need to review...let me see if I can do this by the end of the month.

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3577

To correct rebasing issues.

PR #3038, opened 2022-11-23
Snmpdetect fix cisco_ios detection and add routeros

kacem-expereo:
Hi Team,
I've found some Cisco ios devices that were not matching the regexp due to the "," not being present
Example `Cisco IOS Software C881 Software` will not match `".*Cisco IOS Software,.*"`
So I propose `".*Cisco IOS Software.*,.*"` instead.
I've run it successfully on many devices.
Additionally we can detect routeros via a very simple match.
Example of device output to snmprequest : 
`iso.3.6.1.2.1.1.1.0 = STRING: "RouterOS RB2011UiAS"`

Let me know what you think!
Kindly,

### DIFF
--- a/netmiko/snmp_autodetect.py
+++ b/netmiko/snmp_autodetect.py
@@ -56,7 +56,7 @@ SNMP_MAPPER_BASE = {
     },
     "cisco_ios": {
         "oid": ".1.3.6.1.2.1.1.1.0",
-        "expr": re.compile(r".*Cisco IOS Software,.*", re.IGNORECASE),
+        "expr": re.compile(r".*Cisco IOS Software.*,.*", re.IGNORECASE),
         "priority": 60,
     },
     "cisco_xe": {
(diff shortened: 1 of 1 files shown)


netmiko/snmp_autodetect.py
```diff
@@ -114,6 +114,11 @@
         "expr": re.compile(r"PowerConnect.*", re.IGNORECASE),
         "priority": 50,
     },
+    "mikrotik_routeros": {
+        "oid": ".1.3.6.1.2.1.1.1.0",
+        "expr": re.compile(r"^RouterOS.*", re.IGNORECASE),
```
ktbyers, line 119: I would just make it be .`*RouterOS.*` (i.e. make the beginning a little more flexible). Do you want to update/change that?

ktbyers (changes requested):

ktbyers (approved):

Issue #3033, opened 2022-11-17
Expose parameter expect_string in PaloAltoPanosBase commit() method


hkam40k:
Hello,

Currently the `commit` method has a hardcoded `expect_string="100%"` when calling `self.send_command_expect()`. If the response of the commit is anything else than a full success, the commit will fail with an error `Search pattern never detected in send_command: 100%".

It would be good to expose the `expect_string` parameter in the `commit` method parameters, so that users can customize the search parameter to account for common errors, such as validation errors and lock errors.

The following method signature for `commit` would catch validation errors, commit-lock errors (somebody else holds the commit-lock), and also successful commits:
```
def commit(
        self,
        comment=None,
        force=False,
        partial=False,
        device_and_network=False,
        policy_and_objects=False,
        vsys="",
        no_vsys=False,
        expect_string="(Validation [Ee]rror|Server [Ee]rror|100%)",
        delay_factor=0.1,
    ):
```
(Unfortunately, at least some PA platforms, seem to return the string "error" in lower or upper case depending on type of error, so it does not hurt to be prepared in case PA makes it more uniform one day...)

If you agree with the above, I can submit a PR. Even if the default value for `expect_string` is kept `"100%"` to be more backwards compatible, at least exposing the parameter would still be a good idea.

ktbyers:
We could just make the regex search not care about capitalization.

What about just changing the one existing line to:

```python
expect_string=r"error|100%"
```

Formatted correctly so that capitalization doesn't matter i.e. re.I flag (essentially).

I am not very keen on adding an argument to the method for this (but let me know if I am missing something).

hkam40k:
> What about just changing the one existing line to:
> 
> ```python
> expect_string=r"error|100%"
> ```
> 
> Formatted correctly so that capitalization doesn't matter i.e. re.I flag (essentially).

While catching the string "error" would maybe catch more types of errors that I can't currently think of, it may be too generic, since currently it would also detect a commit with a message with the word "error" in it to be a hit, resulting in prematurely exiting the commit progress polling, and eventually raising an error when `commit_marker` is not detected in the output.

Would a compromise of `expect_string=r"validation error|server error|100%"` be slightly more safe to use? Still not completely bulletproof but I guess it's not possible to get around it since the `command_string` gets echoed from the channel.

> I am not very keen on adding an argument to the method for this (but let me know if I am missing something).

[...]

PR #3053, opened 2022-12-14
Add more options to PA expect_string regex in commit() to also detect failure responses (#3033)

hkam40k:


ktbyers:
@hkam40k You should raise an error on a commit failure.

Also I probably wouldn't want to modify base_connection.py behavior since that is pretty fundamental. You can embed ignore capitalization inside the pattern instead.

See example here:

```python
re.search(r'(?i)test', 'TeSt').group()    ## returns 'TeSt'
re.match(r'(?i)test', 'TeSt').group()     ## returns 'TeSt'
```

https://stackoverflow.com/questions/500864/case-insensitive-regular-expression-without-re-compile

Do you want to modify this PR?

hkam40k:
> You should raise an error on a commit failure.

What do you mean by this? I think this already happens, since the `commit()` method already ends in error checking, and raises a `ValueError` if commit was unsuccessful due to `commit_marker` not being present in the command output, along with a more specific error message than the current `"Search pattern never detected in send_command: ..."`?

```
if commit_marker not in output.lower():
    raise ValueError(f"Commit failed with the following errors:\n\n{output}")
return output
```

> Also I probably wouldn't want to modify base_connection.py behavior since that is pretty fundamental. You can embed ignore capitalization inside the pattern instead.

This is a good idea, I had not thought of that. I will fix the PR, after your input for my question regarding the first point.

ktbyers:
I am going to close this PR since the referenced modifications were never made.

PR #3072, opened 2023-01-05
Allow specifying custom policy

taikedz:
I have use-cases where neither "Reject" nor "Accept and register host key" are suitable (testing suite which brings up fresh boxes with different fingerprints on a regular basis).

This change overrides `ssh_strict` to allow it to receive and use an explicit policy object.

My own purpose is to provide a customised policy instance to directly pass into our existing connection implementation.

I concede this is not the cleanest solution ; this PR may also stand as a discussion point for what better solution would be preferred. If a separate parameter (there are already many!) were preferable, what to do with `ssh_strict` ? Ignore it ? Raise exception for incompatible flag if `ssh_strict==True` ?

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -228,8 +228,9 @@ class BaseConnection:
 
         :param allow_agent: Enable use of SSH key-agent.
 
-        :param ssh_strict: Automatically reject unknown SSH host keys (default: False, which
-                means unknown SSH host keys will be accepted).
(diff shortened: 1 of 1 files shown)


ktbyers:
Couldn't you just set `self.key_policy` after object initialization?

You would have to set `auto_connect=False` in Netmiko arguments and then call `netmiko_object._open()` (i.e. basically don't have Netmiko automically connect so that you can change the key_policy before connecting).

taikedz:
Hadn't realised that was a possibility , but that makes total sense. Thank you.

This change therefore not required.

PR #3078, opened 2023-01-11
update terminal width and return char to work for all oneos versions

mwallraf:
Sets correct terminal width commands that will work for ONEOS5 + ONEOS6 versions. Also the return character will be changed for ONEOS6 devices, otherwise the prompt is repeated in the output results.

### ONEOS5 log  (runs hostname command) :

```
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: b'\r\n'
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
dops-lab-02#
DEBUG:netmiko:Pattern found: (\#|>) 
dops-lab-02#
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:[find_prompt()]: prompt is dops-lab-02#
DEBUG:netmiko:In disable_paging
DEBUG:netmiko:Command: term len 0

DEBUG:netmiko:write_channel: b'term len 0\r\n'
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: term len 0

DEBUG:netmiko:Pattern found: (term\ len\ 0) term len 0
DEBUG:netmiko:term len 0
DEBUG:netmiko:Exiting disable_paging
DEBUG:netmiko:write_channel: b'screen-width 512\r\n'
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: dops-lab-02#screen-width 512
            ^
Error: Invalid command
dops-lab-02#
DEBUG:netmiko:Pattern found: (dops\-lab\-02) 
dops-lab-02
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:write_channel: b'stty columns 255\r\n'
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: stty columns 255

DEBUG:netmiko:read_channel: 
DEBUG:netmiko:read_channel: dops-lab-02#
DEBUG:netmiko:Pattern found: (dops\-lab\-02) stty columns 255
dops-lab-02
DEBUG:netmiko:read_channel: 
DEBUG:netmiko:Clear buffer detects data in the channel
DEBUG:netmiko:read_channel: 
[...]

PR #3080, opened 2023-01-13
Fix oneos terminal width issue

ktbyers:


ktbyers:
Supersedes PR:

https://github.com/ktbyers/netmiko/pull/3078

ktbyers:
@mwallraf I made a few minor updates to your PR here.

ktbyers:
Proposed code for `session_preparation` looks as follows:

```python
    def session_preparation(self) -> None:
        """Prepare connection - disable paging"""

        self._test_channel_read(pattern=r"[>#]")
        self.disable_paging(command="term len 0")

        # try ONEOS6 command first - fallback to ONEOS5 if it fails
        self.set_terminal_width(command="screen-width 512", cmd_verify=True)
        output = self._test_channel_read(pattern=r"[>#]")
        if "error" in output.lower():
            self.set_terminal_width(command="stty columns 255", cmd_verify=True)
        else:
            # ONEOS6 uses different enter
            self.RETURN = "\n"

        self._test_channel_read(pattern=r"[>#]")
        self.set_base_prompt()
```

ktbyers:
@mwallraf Can you test this and still see if it works properly?

Basically I made a few pretty standard Netmiko updates that should improve the performance and reliability upon login (while keeping your original proposed patters to the terminal width).

mwallraf:
@ktbyers, I've just tested this and it works perfect, exactly as expected!  Thanks for the update.


FYI - the hostname command just returns the hostname of the device but without this fix the output always included the original prompt in the output.  (this happened for all commands, not only hostname)

for oneos5:
```
dops-lab-02
```

for oneos6:
```
uliege01-41che-01

uliege01-41che-01#
```

ktbyers:
Okay, so you were just using `hostname` as a test command then?

mwallraf:
> @mwallraf Okay, so you were just using `hostname` as a test command then?

@ktbyers Yes indeed, I should have mentioned it more clearly in the PR :-)
Thanks for the merge !!

PR #3087, opened 2023-01-18
Prevent channel reader from writing empty debug log lines

taikedz:
Some commands sent through to the SSH endpoint can have delays between outputs.

When a read action encounters the pause in emitted output, the debug statement causes numerous debug lines to appear in logs such as 

```
read_channel: earlier data
read_channel:
read_channel:
read_channel:
read_channel:
read_channel: some new data
```

This change prevents logging from happening when exactly no new data was received.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -592,7 +592,8 @@ key_file: {self.key_file}
         new_data = self.normalize_linefeeds(new_data)
         if self.ansi_escape_codes:
             new_data = self.strip_ansi_escape_codes(new_data)
-        log.debug(f"read_channel: {new_data}")
+        if len(new_data) > 0:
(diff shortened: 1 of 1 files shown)


ktbyers:
I find the empty reads useful as it gives better clues as to what is going on.

Have you looked at the Netmiko session_log that potentially gives you more what you are looking for.

taikedz:
We can use the session log indeed - but test frameworks often hijack the logging outputs and ingest them for producing reports (examples being `robotframework` and `pytest`) , and the extra lines cannot be prevented from entering the log from within our code - at leas that's what I was thinking.

Writing this now, I presume there's something that can be done by adding a handler to `netmiko.base_connection.log` from within our own code, without touching netmiko.

Apologies for the noise.

PR #3091, opened 2023-01-23
Improve Fortinet driver performance

ktbyers:


ktbyers:
@Gatorjosh14 FYI, updated Fortinet driver here.

There are still some missing parts from your original work namely send_config_set(), but this should be a significant improved to the original driver.

This includes some contributions from your original PR.

ktbyers:
I am going to go ahead and merge this as it is better than the current driver. We can then work on expanding/improving this code.

PR #3092, opened 2023-01-24
Update adva_aos_fsp_150_f3.py

guychilton77:
disable_paging command missing f string for the username. Not sure how this was missed during the regression testing, all tests passed, but I noticed this during some extended testing.


ktbyers:
@guychilton77 Yeah strange that the above issue didn't fail on the linter.

PR #3097, opened 2023-01-31
Docstring Updates

ktbyers:


ktbyers:
@MarkWine Thanks for your earlier PR on the docstrings.

I made a few additional updates here.

PR #3123, opened 2023-02-20
Allow check_string override in check_config_mode

okozachenko1203:
Aruba switch 8325 uses `(config)#` as the config mode check string without any whitespace.
To support both old versions (`(config) #`) and the new version (`(config)#`), I tried to make `config_mode` function configurable config mode check string.


okozachenko1203:
Can you guys review this PR please?

ktbyers:
@okozachenko1203 I am going to close this as it old and it looks like it is referring to Aruba AOS-CX switches which are a different driver than the PR is using.

If there is still an underlying issue that needs fixed, let me know.

PR #3127, opened 2023-02-23
Fortinet: Fix issue with privilege

FloLaco:
Issue #2833
I've added the ability to provide a new argument for Fortinet driver.

This new argument avoid to check the configuration for a certain command.
In some cases, this command is not on configuration, depending on user permission.


ktbyers:
Hopefully fixed by #3248 

If not, we should re-open an issue on the underlying problem.

PR #3131, opened 2023-03-02
Add buffer clean in Huawei session preparation

k-ribot:


ktbyers:
@k-ribot Why is this needed? Do you have more details?

I would have thought it was unnecessary and slows things down.

ktbyers:
Can you show what the CLI looks like when you see this message?

```
Pattern not detected: 'Info:\\ The\\ configuration\\ takes\\ effect\\ on\\ the\\ current\\ user\\ terminal\\ interface\\ only\\.' in output.
```

i.e. the message `The configuration takes effect on the current...`

You can probably capture this via the Netmiko session_log and just include the session_log.

I am reluctant to add the buffer clearing in the `session_preparation` method as it slows things down meaningfully (which is why I removed it previously). Also your error/issue doesn't exactly match what I would expect...so that is why I want to dig into it a bit more.


Thanks, Kirk

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3578

To fix rebasing issues.

PR #3140, opened 2023-03-15
Cisco WLC Read-Only user config paging disable fails

twistedroutes:
add send_command_w_paging function

For using a read-only user, 'config paging disable'will fail (config requires write). For results longer than a page Cisco WLC adds a '--More-- or (q)uit' message because pagination is still enabled.  This function sends a "space" if '--More-- or (q)uit' is identified


ktbyers:
@twistedroutes I am going to close and then roll into an issue for creating a general output pager (for the case you don't have specific permissions).

Issue #3012, opened 2022-11-01
send_command may return before the command output gets responsed


spengjie:
Hi all,

Please find the brief session log and my analysis below.
```# send '\n'

******************************************************************************
* Copyright (c) 2004-2021 New H3C Technologies Co., Ltd. All rights reserved.*
* Without the owner's prior written consent,                                 *
* no decompiling or reverse-engineering shall be allowed.                    *
******************************************************************************

<switch>
<switch>
# send '\n'

<switch>
# send '\nscreen-length disable\n'

<switch>screen-length disable
<switch>
# send_command starts here
# send '\n'
# send '\n'

# send '\n'
<switch>
<switch>
# send 'display version\n'

<switch>  # a prompt received before getting the response of 'display version', so send_command returns here. If the response of 'display version' is received before the prompt, send_command will work fine.
# send '\n'
display version
# send '\n'
# send 'quit\n'
```
H3C Comware version: 7.1.070, Release 6555P01
Set global_cmd_verify to True can work around this issue. This issue is related to #1763.

ktbyers:
Comware puts you in a bad position as they have no way to set the terminal width. This will cause cmd verify to fail for long commands.

Consequently, `global_cmd_verify` is disabled by default in Comware.

But then with cmd_verify disabled you can run into the above problems (i.e. slow devices that don't respond fast enough can lead Netmiko to match on the wrong prompt).

So we have two options:
1. Enable global_cmd_verify and live with the long command problem/errors (which users could potentially work around).
2. Disable global_cmd_verify and live with the slow device / match on wrong prompt problem (which users could potentially work around).

Fundamental issue is Comware should behave better with wide terminals.

ktbyers:
I am going to close this as I don't see a solution for Comware.

The workaround is to set `global_cmd_verify=True`, but then there are problems with long commands if you do this (due to lack of terminal width on Comware).

If you see a solution or better solution, let me know.

PR #3141, opened 2023-03-17
Update hp_comware.py and hp_procurve.py to fix bug of output end error in disable paging

jiujing:
In disable paging，it is  ok to just send command. If we add a RETURN in command ，it will  make a muddledness with output，Netmiko receive RETURN's prompt and    think it is ok. Disable paging command's output is still in channel.


ktbyers:
@jiujing Why does sending the extra `enter` matter? 

I believe there was a reason that I needed to do this and I am not seeing why it would matter (having the extra `enter` in the output)?

ktbyers:
Can you expand or provide more details on why the current code causes you an issue?

jiujing:
> @jiujing Why does sending the extra `enter` matter?
> 
> I believe there was a reason that I needed to do this and I am not seeing why it would matter (having the extra `enter` in the output)?

Let's assume this scenario, we log in to the device and send a command.Netmiko actually sends a carriage return, a command to cancel paging, a command that we actually want to send.The logic to cancel paging recognizes the echo of the first carriage return sent as the end. In fact, there is still an echo in the tunnel where we cancel pagination.
When we send the actual command to be executed, netmiko determines that the end is roughly determined by the pattern of the device name. The device gives us the echo of the corresponding command in the tunnel, but when we read the echo from the tunnel, we mistakenly end with the echo of the cancel paging command.
The real echo is still in the tunnel.
I'm a NetDevOps engineer doing some technical sharing in my spare time, and Netmiko is one of my favorite Python packages. My readers often encounter echo errors when using new versions of netmiko, and after giving them debugs, I locked in presumably because of this, and after modifying the source code, the error disappeared.
In netmiko 3.4.0 ，it use self.clear_buffer() to clear buffer to avoid this error，but it cancel in netmiko 4.
All of the above is machine translation, and I'm not sure if it's clear.

spengjie:
This PR seems related to issue #3012 that I filed in 2022. But this PR may not resolve the issue since there are still some other RETURNs may cause an early prompt response.

ktbyers:
Ended up merging both comware and procurve change (as I couldn't see a reason why the extra enter was really needed).

PR #3159, opened 2023-04-08
Implemented "try_alternative" in ConnectHandler

cla-azzarello:
This PR implements the new optional flag "try_alternative" in ConnectHandler.

It defaults to False.

Setting this flag to True, if a SSH connection fails, it will try using Telnet and vice versa.

I find it useful because in large network environments with mixed devices configurations, it could be boring to fix all device_type fields.

Sometimes when you operate on hundreds of devices, you don't know if SSH or Telnet is active. With this fix, you can save a lot of time in fixing.

I always deploy a similar behavior in my applications that use Netmiko, so I thought it might be useful to bring the functionality upstream.

Hope this community will appreciate this.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -141,6 +141,7 @@ class BaseConnection:
         secret: str = "",
         port: Optional[int] = None,
         device_type: str = "",
+        try_alternative: Optional[bool] = False,
         verbose: bool = False,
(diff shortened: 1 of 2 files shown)


cla-azzarello:
Closed because I noticed I had committed to the main branch, my mistake.
Opened a new one in #3160

PR #3160, opened 2023-04-09
Implemented new "AgnosticHandler" entrypoint

cla-azzarello:
~~This PR implements the new optional flag "try_alternative" in ConnectHandler.~~

~~It defaults to False.~~

~~Setting this flag to True, if a SSH connection fails, it will try using Telnet and vice versa.~~

This PR implements the new "AgnosticHandler" entrypoint.

By using it, if a SSH connection fails, it will try using Telnet and vice versa.

I find it useful because in large network environments with mixed devices configurations, it could be boring to fix all device_type fields.

Sometimes when you operate on hundreds of devices, you don't know if SSH or Telnet is active. With this ~~flag~~ entrypoint, you can save a lot of time in fixing.

I very often deploy a similar behavior in my applications that use Netmiko, so I thought it might be useful to bring the functionality upstream.

The default behavior of Netmiko doesn't change. It just adds a functionality that could be useful in some scenarios.

Hope this community will appreciate this.

### DIFF
--- a/netmiko/__init__.py
+++ b/netmiko/__init__.py
@@ -38,6 +38,7 @@ log.addHandler(logging.NullHandler())
 
 
 from netmiko.ssh_dispatcher import ConnectHandler  # noqa
+from netmiko.ssh_dispatcher import AgnosticHandler  # noqa
 from netmiko.ssh_dispatcher import ConnLogOnly  # noqa
 from netmiko.ssh_dispatcher import ConnUnify  # noqa
 from netmiko.ssh_dispatcher import ssh_dispatcher  # noqa
@@ -66,6 +67,7 @@ Netmiko = ConnectHandler
 
 __all__ = (
     "ConnectHandler",
+    "AgnosticHandler",
     "ConnLogOnly",
     "ConnUnify",
     "ssh_dispatcher",
--- a/netmiko/ssh_dispatcher.py
+++ b/netmiko/ssh_dispatcher.py
@@ -388,6 +388,25 @@ def ConnectHandler(*args: Any, **kwargs: Any) -> "BaseConnection":
     return ConnectionClass(*args, **kwargs)
 
 
+def AgnosticHandler(*args: Any, **kwargs: Any) -> "BaseConnection":
+    """Alternative entrypoint that is agnostic to connection method.
+    If an SSH connection fails it tries to use Telnet, if available
+    and vice versa."""
+    try:
+        return ConnectHandler(*args, **kwargs)
+    except (NetmikoTimeoutException, ConnectionRefusedError):
+        device_type = kwargs["device_type"]
+        if device_type in platforms_str:
+            alternative_device = device_type + "_telnet"
+        elif device_type in telnet_platforms_str:
+            alternative_device = device_type.replace("_telnet", "")
+
+        if alternative_device in platforms:
+            kwargs["device_type"] = alternative_device
+            return ConnectHandler(*args, **kwargs)
+        raise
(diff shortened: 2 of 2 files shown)


codekuu:
Hi! The formatting fails (black & pylama is complaining):
```
(.venv) codekuu:~/netmiko$ pylama netmiko/ssh_dispatcher.py 
netmiko/ssh_dispatcher.py:363:41 E252 missing whitespace around parameter equals [pycodestyle]
netmiko/ssh_dispatcher.py:363:42 E252 missing whitespace around parameter equals [pycodestyle]
netmiko/ssh_dispatcher.py:365:1 W293 blank line contains whitespace [pycodestyle]

(.venv) codekuu:~/netmiko$ black --check --diff netmiko/ssh_dispatcher.py 
--- netmiko/ssh_dispatcher.py   2023-04-21 05:57:16.537733 +0000
+++ netmiko/ssh_dispatcher.py   2023-04-21 06:13:35.866608 +0000
@@ -358,13 +358,15 @@
 telnet_platforms = [x for x in platforms if "telnet" in x]
 telnet_platforms_str = "\n".join(telnet_platforms)
 telnet_platforms_str = "\n" + telnet_platforms_str
 
 
-def ConnectHandler(try_alternative: bool=False, *args: Any, **kwargs: Any) -> "BaseConnection":
+def ConnectHandler(
+    try_alternative: bool = False, *args: Any, **kwargs: Any
+) -> "BaseConnection":
     """Factory function selects the proper class and creates object based on device_type.
-    
+
     :param try_alternative: If True, when an SSH connection fails,
                 try connecting using Telnet and vice versa.
     """
     device_type = kwargs["device_type"]
     if device_type not in platforms:
would reformat netmiko/ssh_dispatcher.py

Oh no! 💥 💔 💥
1 file would be reformatted.
```

And I think the variable should be moved to the BaseConnection class and the logic should be placed in the _open() or establish_connection methods instead.

I do however think this is a good addition, but the Device Inventory should have this information (ssh/telnet etc...)

cla-azzarello:
@codekuu thank you for your reply.

Now the formatting is fixed.

```
(.venv) > black --check --diff netmiko/ssh_dispatcher.py
All done! ✨ 🍰 ✨
1 file would be left unchanged.
(.venv) > pylama netmiko/ssh_dispatcher.py 
(.venv) > 
```


> And I think the variable should be moved to the BaseConnection class and the logic should be placed in the _open() or establish_connection methods instead.

Regarding this observation, the rationale I followed is that ConnectHandler should, indeed, handle the connection method by instantiating the correct Connection Class.
BaseConnection and the classes that inherit this, come into play only after the ssh_dispatcher function has selected the correct Connection Class.
We can discuss this and, if you think it is necessary, I can try to implement my logic inside BaseConnection.


> but the Device Inventory should have this information (ssh/telnet etc...)

Here I did not understand exactly what you are referring to.

cla-azzarello:
Force-pushed new commits to make the code more clean and readable.

ktbyers:
This probably should be implemented as its own entry point to Netmiko (i.e. similar to ConnLogOnly in the same file).

cla-azzarello:
> This probably should be implemented as its own entry point to Netmiko (i.e. similar to ConnLogOnly in the same file).

@ktbyers DONE!

cla-azzarello:
@ktbyers pls review if you have time, thanks!

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3232

PR #3168, opened 2023-04-17
Teldata CIT Driver

ktbyers:


ktbyers:
CC: [dysby](https://github.com/dysby)

ktbyers:
@dysby I think this is the same as what you implemented a while back. I added type hints and converted methods to be private methods.

I also made some other minor regression changes (though those were pretty minor).

I am going to go ahead and merge this. Let me know if you see any issues.

PR #3169, opened 2023-04-18
Extreme ERS handle menu

ktbyers:


ktbyers:
Related discussion:

https://github.com/ktbyers/netmiko/discussions/2374

ktbyers:
FYI, @romanstech  @Microbuse 

There were some other changes to the extreme ERS login so I think this incorporates the earlier menu fix in with that code.

Thanks for contributing this.

PR #3189, opened 2023-05-03
Added support for Ericsson MINI-LINK (MLTN) 66XX & 63XX (updated)

ktbyers:


ktbyers:
@codekuu FYI, I changed the special_login_handler to use read_until_pattern format as experience has shown me this is a more reliable pattern to use. 

Let me know if you see any issues with the changes.

ktbyers:
@codekuu I also added the standard Netmiko no-enable mode behavior. I assume this is correct i.e. these devices do not have any enable mode--correct?

ktbyers:
I am going to merge this. It would be a good idea if you regression tested the changes I made using the Netmiko test suite (if you can).

Thanks for all your work here.

If there are issues, just let me know.

codekuu:
Hi @ktbyers!

I have tested the code and ran it in our dev environment towards real machines, works like a charm.
Can't see any issues.

Thanks for really investing your time to make this project as great as it is!

ktbyers:
Thanks @codekuu

PR #3193, opened 2023-05-04
Improve behavior of normalize_linefeeds() method

ktbyers:


ktbyers:
Doing normalize_linefeeds only "on the complete buffer" is likely problematic. 

With screen scraping there is no good notion of the a command being done so there is no good definition of a complete buffer.

You could argue with a read_until_pattern() type of a behavior that the complete buffer is when trailing pattern is detected, but then this breaks down with a read_channel_timing() i.e. here Netmiko ends based on timers and there could be more data that we just didn't wait long enough for.

Although you can definitely argue that even with the above issues that potentially this somewhat complete buffer is better than the current extra whitespace problem.

I also recall there being issues with conversions on very large outputs (when they were done all at once on the entire output). Though it is possible that this was only an issue with stripping ANSI escape codes and not with newline normalization.

ktbyers:
This solutions has two parts:

1. A way to disable `normalize_linefeeds` if you want to just use what is returned by the device. This defaults to False i.e. Netmiko by default will still normalize_linefeeds (argument name: `disable_lf_normalization`).
2. Netmiko tries to somewhat intelligently avoid the block ending in "\r". Although there are some cases where Netmiko could/would still fail (in which case you would be forced to `disable_lf_normalization` (if the extra whitespace/newlines mattered to you).

ktbyers:
[tik-stbuehler](https://github.com/tik-stbuehler) 

Feel free to comment on this. I didn't super love the solution, but I like it better than the other options.

tik-stbuehler:
I asked for a flag to disable it, and you build one - should be good enough :)

Modifying existing behaviour is a difficult path. I think you found a good compromise:

* a small change (I could think of a complex state handling chunks with trailing `\r`... probably not worth it)
* catching the most obvious problem (large output, unlucky chunk split)
* flag to disable it

Issue #2168, opened 2021-02-27
cisco_ios get_remote_file_size does not properly handle remote files that are in subdirectories


anitrajpurohit28:
**what am I trying to achieve?**
>Copy a file from the cisco ios router to the Linux server, where I am running this script

I've hardcoded the source_file and dest_file to ease of understanding
the file does exist in the specified path and appropriate permission are given

**script file**
```
from netmiko import ConnectHandler, file_transfer

cisco_ios_details = {'device_type': 'cisco_ios',
                     'host': 'xx.xx.xx.xx',
                     'username': '*****',
                     'password': '*****',
                     'fast_cli': False}

def transfer_file_from_ios_current_directory():
    device_connector = ConnectHandler(**cisco_ios_details)
    try:
        return_dict = file_transfer(device_connector,
                                    source_file="guest-share/example_ios.json",
                                    dest_file='./example_linux.json',
                                    direction='get',
                                    file_system='flash:',
                                    overwrite_file=True)
        print(return_dict)
    except Exception as e:
        print('Unable to transfer.\nReason: {}'.format(e))

    device_connector.disconnect()


transfer_file_from_ios_current_directory()
```
**Output:**
```
Unable to transfer.
Reason: local variable 'file_size' referenced before assignment
```

**The file is present in router**
```
CSR#dir flash:guest-share/example_ios.json
Directory of bootflash:guest-share/example_ios.json

431424  -rw-             7115  Feb 27 2021 06:44:17 +00:00  example_ios.json
```

ktbyers:
Can you remove your try/except and post the full stack trace so I can see where the error is occurring?


Kirk

anitrajpurohit28:
After getting rid of try/except, following is the traceback


```
/Users/anbabula/.pyenv/shims/python "/Users/anbabula/Library/Application Support/JetBrains/PyCharmCE2020.3/scratches/scratch_10.py"
Traceback (most recent call last):
  File "/Users/anbabula/Library/Application Support/JetBrains/PyCharmCE2020.3/scratches/scratch_10.py", line 29, in <module>
    transfer_file_from_ios_current_directory()
  File "/Users/anbabula/Library/Application Support/JetBrains/PyCharmCE2020.3/scratches/scratch_10.py", line 23, in transfer_file_from_ios_current_directory
    overwrite_file=True)
[...]

PR #3199, opened 2023-05-05
Fix bug in handling of subdirectories in remote_file_size

dystland:
Fixes bug described in #2168  

The remote_file variable contains the path of the remote file, including parent directories. However, the output from dir only contains the filename without parent directories.

This fix utilizes os.path.basename to extract the filename from remote_file into remote_file_base and use that as the pattern when looking for the file size in the dir output.


ktbyers:
Thanks @dystland

PR #3215, opened 2023-05-19
Tplink read issue in session_preparation()

ktbyers:


ktbyers:
@drousseau FYI, let me know if you see any issues with this.

ktbyers:
It should fix the reported issue.

Issue #3144, opened 2023-03-22
SNMP autodetect is broken if you feed a hostname into it.


ktbyers:
Traceback (most recent call last):
  File "/home/kbyers/netmiko_course/class10/collateral/snmp_detect.py", line 9, in <module>
    my_snmp = SNMPDetect(
  File "/home/kbyers/VENV/py3_venv/src/netmiko/netmiko/snmp_autodetect.py", line 248, in __init__
    if ipaddress.ip_address(self.hostname).version == 6:
  File "/usr/local/lib/python3.9/ipaddress.py", line 53, in ip_address
    raise ValueError('%r does not appear to be an IPv4 or IPv6 address' %
ValueError: 'cisco1.lasthop.io' does not appear to be an IPv4 or IPv6 address

ipaddress.ip_address does not support a hostname here.

ktbyers:
Relevant change that broke this:

https://github.com/ktbyers/netmiko/pull/3028

ktbyers:
Need better testing for snmp_autodetect (would this have been caught in my standard release process)?

kacem-expereo:
Sorry about this, indeed i actually never use hostnames when using netmiko haha..
Having a look

kacem-expereo:
@ktbyers  sure, i will proceed with some testing on my side and move forward with a PR :)

kacem-expereo:
It's done @ktbyers ,sorry for the delay! waiting for your comments
Kindly,

ktbyers:
Should be fixed in this PR:

https://github.com/ktbyers/netmiko/pull/3230

PR #3219, opened 2023-05-22
Support ipv4 & ipv6 , static and fqdn in snmpdetect #3144

kacem-expereo:
Fixing [bug ](https://github.com/ktbyers/netmiko/issues/3144#)

- Removing ipaddress lib and using socket instead.
- adding a function identify_address_type which returns a list containing one or more entry, "IPv4" or "IPv6". Not a huge fan of it ...let me know what you think.

Tested scenarios : 

- IPV4 Static
- IPV6 static
- IPV4 FQDN ("A" record)
- IPV6 FQDN ("AAAA" record)
- IPV4+IPV6 FQDN ("A" & "AAAA" records for the same fqdn)

In the last scenario, IPV6 is priorized due to line 280...
Should we do it another way?

### DIFF
--- a/netmiko/snmp_autodetect.py
+++ b/netmiko/snmp_autodetect.py
@@ -20,10 +20,10 @@ SNMPDetect class defaults to SNMPv3
 Note, pysnmp is a required dependency for SNMPDetect and is intentionally not included in
 netmiko requirements. So installation of pysnmp might be required.
 """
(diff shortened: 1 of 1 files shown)


ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3230

PR #3231, opened 2023-06-09
Fiberstore s3200 driver with fixed type hints

ktbyers:


ktbyers:
@shrank FYI

ktbyers:
@shrank I think this should be essentially the same as what you had. I renamed some of the driver parts (as it looked like `FSOS` is the name of the underlying network operating system).

I also converted some things to be more consistent with the standard way of doing things in Netmiko.

Finally, I fixed a few items related to Python type hints.

Let me know if you see any issues. I will probably go ahead and merge this, but feel free to regression test it against your device(s) and let me know if there are issues.

PR #3232, opened 2023-06-09
Implement TelnetFallback entry point to Netmiko

ktbyers:
Will try to connect using telnet if SSH connection fails for the given device type.

### DIFF
--- a/netmiko/__init__.py
+++ b/netmiko/__init__.py
@@ -38,7 +38,7 @@ log.addHandler(logging.NullHandler())
 
 
 from netmiko.ssh_dispatcher import ConnectHandler  # noqa
-from netmiko.ssh_dispatcher import TelnetFallback # noqa
+from netmiko.ssh_dispatcher import TelnetFallback  # noqa
 from netmiko.ssh_dispatcher import ConnLogOnly  # noqa
 from netmiko.ssh_dispatcher import ConnUnify  # noqa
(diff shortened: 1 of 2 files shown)


ktbyers:
FYI @cla-azzarello

ktbyers:
@cla-azzarello This should be very similar to what you implemented. I renamed it `TelnetFallback` (people almost never agree on names).

I also made it only: `try SSH -> then try telnet` (i.e. there is no try telnet and then SSH).

Basically, if you have SSH working, you should not be using telnet.

I also added some minor logic to handle device_types of the form `cisco_ios_ssh` and have them get converted to `cisco_ios_telnet`.

Thanks for all your work on this.

ktbyers:
Let me know if you see any issues. @cla-azzarello 

I am going to go ahead and merge this, but if there are issues, then we can make changes/fixes.

cla-azzarello:
Good point @ktbyers !

Thank you!

netmiko/__init__.py
```diff
@@ -66,6 +67,7 @@ def check_python_version():  # type: ignore
 
 __all__ = (
     "ConnectHandler",
+    "AgnosticHandler",
```
cla-azzarello, line 70: @ktbyers this should be TelnetFallback

ktbyers: I will fix this.

PR #3233, opened 2023-06-09
Improve Extreme SLX-OS autodetect

ktbyers:


ktbyers:
@msimkins

I updated this to make the pull-request against `develop` and not `master`.

Also just made it search for the simpler string (since one string was a subset of the other string).

PR #3240, opened 2023-06-28
Nokia SR Linux: Modify check_config_mode regex

sebastien-guay:
Nokia SR Linux: Modify check_config_mode regex
Update test with use_keys: "True". Required with containerlab.


ktbyers:
@sebastien-guay Sorry for the very long delay on this.

PR #3245, opened 2023-07-04
Datacom DmOS driver

talesmm14:


ktbyers:
save_config() method probably should be replaced via `commit()` method.

It would be good to have both show, config, and commit tests included in the PR.

ktbyers:
Should update device_type to be `datacom_dmos`.

talesmm14:
> save_config() method probably should be replaced via `commit()` method.
> 
> It would be good to have both show, config, and commit tests included in the PR.

I no longer have contact with this equipment to perform tests, at the time I did it, it worked well on the system I was working on.

I apologize for the lack of dedication to the tests, at the time I had a lot of tasks to complete.

ktbyers:
@talesmm14 No worries at all...I let this PR sit for way too long (so it is pretty much on me).

PR #3247, opened 2023-07-07
Update scp_functions.py | Added support to display progress bar without cleaning the screen.

andrearaujo1988:
Added support to display progress bar without cleaning the screen.
Now it should only shows the percentange and not the progress bar char ">".
Also removed unused vars on the progress_bar function due to this update.
I thought this would be useful since clearing the screen makes the screen to scroll which is not often desired.

### DIFF
--- a/netmiko/scp_functions.py
+++ b/netmiko/scp_functions.py
@@ -18,18 +18,14 @@ if TYPE_CHECKING:
 def progress_bar(
     filename: AnyStr, size: int, sent: int, peername: Optional[str] = None
 ) -> None:
-    max_width = 50
     if isinstance(filename, bytes):
         filename_str = filename.decode()
     else:
         filename_str = filename
-    clear_screen = chr(27) + "[2J"
-    terminating_char = "|"
 
     # Percentage done
     percent_complete = sent / size
     percent_str = f"{percent_complete*100:.2f}%"
-
 
     if peername is None:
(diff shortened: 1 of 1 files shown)


ktbyers:
This was more an example progress bar that works reasonably well (i.e. if you don't like the provided one, you can just create your own).

I am not sure I understood this?

```
I thought this would be useful since clearing the screen makes the screen to scroll which is not often desired.
```

Can you expand/clarify the behavior that you observe?

andrearaujo1988:
> This was more an example progress bar that works reasonably well (i.e. if you don't like the provided one, you can just create your own).
> 
> I am not sure I understood this?
> 
> ```
> I thought this would be useful since clearing the screen makes the screen to scroll which is not often desired.
> ```
> 
> Can you expand/clarify the behavior that you observe?

When passing the progress bar imported from scp_functions I've noticed that the screen keeps getting cleared and new status being printed which makes the screen to scroll until the file transfer is done.
If you need to scroll up to find something you'll have hard times since the screen got populated with every updated on the progress.
I thought it would be 'better' if the progress gets written on the screen but withing the same like and only the percentage being refreshed.
If I can pass my progress it would be great, but I'm not sure how to do so and pass the arguments I want if you can refer me to any documentation with example it would be great.
Thank you.

ktbyers:
You can pass in your own progress_bar, you don't need to use the predefined one that is inside of Netmiko.

PR #3248, opened 2023-07-07
Updating to single command to parse 'set output mode'

msozer-dn:
Current method which sends `config system console` ,  `show full-configuration` , `end` commands in order to parse the set output mode has a limitation to run the commands with a user which has a global configuration permission. 

When a RO user tries to create a connection "show full-configuration" commands run without entering the console section and displays full configuration of the system. prompt_pattern which was used to find the end of the output is not suitable to parse the set output mode from full device configuration. In this situation netmiko raises below error.

`ValueError("Unable to determine the output mode on the Fortinet device.")`

`show full-configuration system console` command produces the expected result as the original commands as RO user can create a successful connection.

### DIFF
--- a/netmiko/fortinet/fortinet_ssh.py
+++ b/netmiko/fortinet/fortinet_ssh.py
@@ -148,13 +148,7 @@ Alternatively you can try configuring 'configure system console -> set output st
         if self._vdoms:
             self._config_global()
 
-        self._send_command_str(
-            "config system console", expect_string=self.prompt_pattern
-        )
-        output = self._send_command_str(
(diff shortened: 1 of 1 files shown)


ktbyers:
@msozer-dn Are there any restrictions on user permissions or Fortinet OS version on where these should work (or should this work generally).

Obviously the current solution as the referenced problems i.e. it doesn't work for RO users (as referenced in linked issue).

msozer-dn:
Current user is a RO user and having difficulties to collect output mode. Updated commands works in RW and RO users.
I am able to collect the output mode when I use a RW user.

ktbyers (approved):

PR #3249, opened 2023-07-08
Remove external dependency on grep from netmiko-show

ReK42:
This PR allows `netmiko-show` to run cross-platform via pure Python. Tested on Windows 11 in Python 3.11 and 3.7.

Also added `--write-output`, which will dump the output to files in the PWD rather than printing to stdout.

Minor, non-impacting edits to other files courtesy of running `black` and `ruff`, and adding a couple things to `.gitignore`. All substantial edits are in [`netmiko/cli_tools/netmiko-show.py`](https://github.com/ktbyers/netmiko/compare/develop...ReK42:netmiko:develop?expand=1#diff-84a4f764fde5b18be53867073cdd7196622a0cb5b1200c1f6d653e6cd1def21b).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ bin/.netmiko.cfg
 .idea/
 
 # Virtual Environment files
+.env/
 .venv/
 bin/
 lib/
@@ -37,6 +38,8 @@ docs/build/doctrees/
 .tox/*
 
 .pytest_cache/*
+.mypy_cache/
+.ruff_cache/
 
 # Mac stuff
 .DS_Store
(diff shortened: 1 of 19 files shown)


ktbyers:
@ReK42 Were you using the version of `black` specified in Netmiko's pyproject.toml. You should use exactly that version of black (otherwise the project will have too much unnecessary changes just do to minor `black version` changes)?

We should get rid of all of these linting changes.

I try to roll black version only at releases to minimize some of the impact of these unneeded linting changes.

ReK42:
Reverted that commit. I had used the latest version of black from pypi.

ktbyers:
This code (netmiko-show) has changed to much so we would need to reimplement this, if desired.

I apologize about that--this just sat for too long without me handling it.

Issue #3273, opened 2023-08-28
Feature Request: ability to register new drivers in ssh_dispatcher


alextremblay:
It would be nice for netmiko users to be able to locally develop drivers for new hardware or modify / extend existing drivers and dog-food them for a while before submitting PRs to merge them into netmiko itself.

It would be really nice if, for example, a user could implement a new driver like so:

```python
from netmiko.ssh_dispatcher import register_as
from netmiko.cisco_base_connection import CiscoSSHConnection

@register_as('aruba_aoscx')
class ArubaOSCXSSH(CiscoSSHConnection):
    # custom code and method overrides
    pass
```

Or make local modifications to netmiko's behaviour like so:

```python
from netmiko.ssh_dispatcher import register_as
from netmiko.arista import AristaSSH

@register_as('arista_eos')
class MyAristaSSH(AristaSSH):
    # custom code and method overrides
    pass
```

ktbyers:
@alextremblay Couldn't you just load the existing dictionary in ssh_dispatcher and add a new key and the associated class?

```python
from netmiko.ssh_dispatcher import CLASS_MAPPER

CLASS_MAPPER["my_driver"] = MyClass
```

alextremblay:
unfortuantely, no.

the `CLASS_MAPPER` global variable can be imported and manipulated, but the `platforms`, `platforms_base`, and `platforms_str` global variables from that same module cannot, and they also need to be modified in order for the `ssh_dispatcher` function to work properly

I'm working on a PR for this as we speak, and will submit it shortly

ktbyers:
Good point on `platforms`--you don't need `platforms_base` nor `platforms_str` though.

So this works:

```python
from netmiko.cisco.cisco_nxos_ssh import CiscoNxosSSH
from netmiko.ssh_dispatcher import CLASS_MAPPER, platforms
from netmiko import ConnectHandler

class MyClass(CiscoNxosSSH):
    pass

CLASS_MAPPER["my_driver"] = MyClass
platforms.append("my_driver")

device = {
    "device_type": "my_driver",
    "host": "nxos1.lasthop.io",
    "username": "admin",
    "password": "password",
}

with ConnectHandler(**device) as nc:
    d = nc.send_command("show lldp neighbors detail")
    print(d)
```

I am open to the PR though not really sure it is needed (since it is just three lines of code to do this with the existing code base).

alextremblay:
[...]

PR #3274, opened 2023-08-28
Class decorator to dynamically add new classes and supported device_type

alextremblay:
for https://github.com/ktbyers/netmiko/issues/3273


Noppes:
PR looks fine, also not 100% convinced this is something that is needed

ktbyers:
I am just going to close this as not needed.

PR #3296, opened 2023-09-20
Update vyos_ssh.py

ovikosta:
Raise ValueError without call config_mode


ktbyers:
Probably should `exit_config_mode()` also (if we are going to enter config mode).

Also you need `output +=` on the second `output` capture (or it will overwrite the first output).

ovikosta:
No problem and also add exit_config_mode() in function `def commit()`?
```
        output = self.config_mode()
        output += self._send_command_str(
            command_string,
            strip_prompt=False,
            strip_command=False,
            read_timeout=read_timeout,
        )
```

PR #3308, opened 2023-10-13
Pass SessionLog obj to session_log arg

### LINKED ISSUE #3307: security: allow SessionLog object to be passed to session_log arg
TL;DR: allow SessionLog object to be passed to session_log argument for BaseConnection

Netmiko logs all channel communication to a file via its own SessionLog for each device interaction. While SessionLog filters the username and password out automatically, there may be a case where you want more words/phrases sanitized in the session log. Currently, you can only pass a string or BufferedIOBase subclass to the session_log argument, which doesn't allow you to edit the SessionLog no_log argument. The no_log argument allows you to pass in a dict with words that you want sanitized from the session log.

By adding the ability to pass in a SessionLog object, a user can customize the no_log argument, which allows for an arbitrary number of words to be sanitized in the logs, not just the username and password.

dannywade:
Allows SessionLog object to be passed to session_log.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -280,7 +280,7 @@ class BaseConnection:
                 (default: False)
 
         :param encoding: Encoding to be used when writing bytes to the output channel.
-                (default: ascii)
+                (default: "utf-8")
 
         :param sock: An open socket or socket-like object (such as a `.Channel`) to use for
                 communication to the target host (default: None).
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -267,7 +267,8 @@ class BaseConnection:
                 to select smallest of global and specific. Sets default global_delay_factor to .1
                 (default: True)
 
-        :param session_log: File path, SessionLog object, or BufferedIOBase subclass object to write the session log to.
+        :param session_log: File path, SessionLog object, or BufferedIOBase subclass object
+                to write the session log to.
 
         :param session_log_record_writes: The session log generally only records channel reads due
                 to eliminate command duplication due to command echo. You can enable this if you
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -267,7 +267,7 @@ class BaseConnection:
                 to select smallest of global and specific. Sets default global_delay_factor to .1
                 (default: True)
 
-        :param session_log: File path or BufferedIOBase subclass object to write the session log to.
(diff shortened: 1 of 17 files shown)


netmiko/base_connection.py
```diff
@@ -386,10 +388,18 @@ def __init__(
                     no_log=no_log,
                     record_writes=session_log_record_writes,
                 )
+            elif isinstance(session_log, SessionLog):
+                # SessionLog object
+                self.session_log = session_log
+                # Add filter after self.session_log is constructed
+                log.addFilter(SecretsFilter(no_log=self.session_log.no_log))
```
ktbyers, line 395: I think we should decouple this, not have Netmiko does this automatically (i.e. the user should do this in their code if they want to set the Python logging the same way as the session_log).

End-user should be able to do this:

```python
>>> import netmiko
>>> netmiko.log
<Logger netmiko (WARNING)>
>>> from netmiko.base_connection import SecretsFilter
>>> netmiko.log.addFilter(SecretsFilter(no_log=log_dict))
```

netmiko/base_connection.py
```diff
@@ -386,10 +388,18 @@ def __init__(
...
+                # Check if filename was provided in session_log
+                if self.session_log.file_name is not None:
+                    self.session_log.open()
```
ktbyers, line 398: I probably wouldn't call `open()`, I would pass in an already open() session_log. I think this would give end-users more flexibility (for example, if they were trying to do something more complex like multiple-thread session_log.

netmiko/base_connection.py
dannywade: Agreed. I pushed a change to remove the filter.

netmiko/base_connection.py
dannywade: Okay. Would you mind suggesting how that should be accomplished? I'm unsure how to implement that... maybe I'm overthinking it.

ktbyers: Just remove lines `397` and `398`.

The SessionLog object would already have been `opened` when passed into Netmiko (i.e. so you pass in an already opened `SessionLog` object that you can just `write()` )

dannywade: Ah yeah, I figured I was overthinking it 😅. Changes pushed.

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3313

PR #3311, opened 2023-10-17
Huawei secure configuration message

ktbyers:


ktbyers:
Message:

```
 Warning: There are security risks in the configuration file. You are advised to save the configuration immediately. If you choose to save, the current configuration file will be unavailable after version downgrade. Are you sure to save now?[Y/N]: Y
 Warning: The current configuration will be written to the device. 
 Are you sure to continue?[Y/N]: Y
```

ktbyers:
@k-ribot FYI, I made some minor regex changes to make the regex easier to read.

I tested the changes with the output string that you provided.

PR #3326, opened 2023-10-24
Updated Adtran Driver - Fixed send_command sync and enable mode issues.

Gatorjosh14:
- Fixed `send_command()` sync issues by re-enabling `global_cmd_verify` via True.  My testing on the Adtran OS 1550-24p switch showed the problems were fixed once enabled.
- Fixed `enable()` mode issues that occur when AAA is configured but not syncing properly, which causes it to fail back to the local database.  When this happens, it rejects the first Secret Password attempt and prompts for a 2nd Secret Password after providing a message, which would be the locally defined enable password.

### DIFF
--- a/netmiko/adtran/adtran.py
+++ b/netmiko/adtran/adtran.py
@@ -5,114 +5,114 @@ from netmiko.exceptions import NetmikoTimeoutException
 
 
 class AdtranOSBase(CiscoBaseConnection):
-	prompt_pattern = r"[>#]"
+    prompt_pattern = r"[>#]"
 
-	def __init__(self, *args: Any, **kwargs: Any) -> None:
-		if kwargs.get("global_cmd_verify") is None:
-			kwargs["global_cmd_verify"] = True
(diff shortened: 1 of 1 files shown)


Gatorjosh14:
Sorry about the Tab formatting issues, it's unfortunately what I use lol.

ktbyers:
@Gatorjosh14 Looks like you ran into some linting issues (Executing `black` on the code will probably fix them).

ktbyers:
@Gatorjosh14 Yeah, you will need to fix it (or else it will just force me to fixing it).

Gatorjosh14:
It appears I was able to resolve the issues, thanks for the `black` recommendation.  It was also giving me issues with the `enable_pattern Optional[str]` vs the exected `str` argument for the newly added `self.read_until_prompt_or_pattern` change for the `pattern ` argument, I resolved by hard-coding the type as a string.

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3329

PR #3327, opened 2023-10-24
Added Telnet Proxy Support

Gatorjosh14:
- Added `telnet_proxy.py` module into main netmiko directory and amended `base_connection.py` to support the newly added feature.
- User must provide the `sock` parameter as a dictionary to the Connection Handler of all the `socks` parameters which will then be used to establish the SOCKS Proxy prior to attempting the connection to the device.


Gatorjosh14:
I successfully tested an Adtran OS 1550-24p switch via both the SSH and Telnet Proxy support.

ktbyers:
Fixed in #3334

PR #3329, opened 2023-10-26
Updated Adtran Driver - Fixed send_command sync and enable mode issues.

ktbyers:


ktbyers:
@Gatorjosh14 I made a few minor updates.

I broke out "Falling back" such that it was no longer coupled to the "enable_pattern" variable. That variable has a different use so I didn't want to overload it with a new use here.

Second, I removed the `def __init__()` method from the adtran.py file and just had it fall back to the default in `base_connection.py` for global_cmd_verify. 

This default value is None which means to use the method specific cmd_verify for send_command (True) and for send_command_timing (False). 

You can override both of these by adding the cmd_verify argument to the given method.

For example:

```
netmiko_obj.send_command_timing(command, cmd_verify=True)
```

ktbyers:
Feel free to test this and let me know if there are any issues (especially with the `enable()` call and the fallback authentication).

PR #3331, opened 2023-10-30
Fix session_log failure to hide "no_log" data.

ktbyers:


ktbyers:
Fixes:

https://github.com/ktbyers/netmiko/issues/3307

cc @dannywade 

Please, test this PR (@dannywade ) and verify it fixes what you saw in #3307

dannywade:
@ktbyers - I confirmed it works! Test output is below (using the same test from before - sanitizing the 10.1.1.1 IP address). Thank you for looking into this and the quick turnaround!

```
danslab-sw01#
danslab-sw01#terminal width 511
danslab-sw01#terminal length 0
danslab-sw01#
danslab-sw01#
danslab-sw01#configure terminal
danslab-sw01(config)#
danslab-sw01(config)#logging host ********
danslab-sw01(config)#logging host 10.1.1.2
danslab-sw01(config)#
danslab-sw01(config)#end
danslab-sw01#
danslab-sw01#
danslab-sw01#exit
```

ktbyers:
@dannywade FYI, there still are probably some edge cases where the session_log write operation is broken into two chunks and consequently misses the obfuscation. They should be very rare, but it is hard to cover all the cases (and it is a hard tradeoff between the use/value of the session log itself and the value of filtering/obscuring).

So I would always view the session log as containing sensitive data and act appropriately (of course, it really is this way filter or no filter as the router contains a lot of sensitive information that it will output and the vast majority of people will not filter it all).

dannywade:
@ktbyers - Agreed. Thanks again for working through this - I really appreciate it!

ktbyers:
@dannywade Yeah, that change ended up being bigger than anticipated :-)

PR #3334, opened 2023-11-09
Add Telnet Proxy Support

ktbyers:


ktbyers:
@Gatorjosh14 In your test code, what are argument are you passing in for Netmiko's `sock`?

Netmiko is expecting a `socket.socket` here, but it pretty clearly looks like you are passing in some form of a proxy_dict. What does proxy_dict look like?

Gatorjosh14:
Correct, in order to keep the TelnetLib changes minimal, I just repurposed the existing sock argument.  Here's how I use it in my code to accommodate for both SSH and Telnet arguments for the Connection Handler.

```
if self.socks_active:
	if self.proxy1_active or self.proxy2_active:
		raise ValueError("Cannot have both Proxy1 or Proxy2 active with SOCKS at the same time.")
	else:
		# Establish SOCKS proxy to established SOCKS Dynamic Tunnel.
		sock = socks.socksocket()

		if self.debug_detail:
			print(f'SOCKS Source IP: {self.socks_source_ip}')
			print(f'SOCKS Source Port: {self.socks_source_port}')

		if self.connection_type == 'ssh':
			sock.set_proxy(
				proxy_type=socks.SOCKS5,
				addr=self.socks_source_ip,
				port=self.socks_source_port,
			)
			sock.settimeout(self.max_auth_timeout)
			try:
				sock.connect((self.device_object['ip'], 22))
			except Exception as e:
				if self.debug_detail:
					raise e
				else:
					raise ValueError("Unable to SOCKS Proxy to device, Timeout")
		else:
		       # Telnet sock dictionary.
			sock = {
				'proxy_type': socks.SOCKS5,
				'proxy_addr': self.socks_source_ip,
				'proxy_port': self.socks_source_port,
				}

		# Inject VIP Access Code if Active.
		if self.device_inject_vip:
			vip_code = PYNDIT_support.get_vip_code(self)		
			print(f"VIP Access code required: {vip_code}")

			temp_password = f"{self.device_password}{vip_code}"
		else:
			temp_password = self.device_password

		if self.debug_detail:
			print(self.device_username)
			print(temp_password)

		# Startup Proxy1 SOCKS Connection to the Device.
		self.net_connect = netmiko.ConnectHandler(
			device_type=self.device_type,
			ip=self.device_object['ip'],
			username=self.device_username,
			password=temp_password,
			default_enter='\r',
			global_delay_factor=1,
			secret=self.device_secret,
			session_timeout=self.max_timeout,
			auth_timeout=self.max_auth_timeout,
			read_timeout_override=self.max_read_timeout,
			sock=sock,
		)
		PYNDIT_support.connection_banner(self, 
			protocol=self.connection_type.upper(),
			device=self.device_object['ip'],
			index_count=self.index_count,
			index_total=self.index_total
			)
		
		max_loops = self.max_timeout * 4
		loops = 1
		while loops <= max_loops:
			self.net_connect.write_channel('\r')
			time.sleep(.5)
			output = self.net_connect.read_channel()

			if '>' in output or '#' in output or '$' in output:
				self.device_main()
				break

			time.sleep(.25)
			loops += 1

		if self.connection_type == 'ssh':
			sock.close()
```

Gatorjosh14:
I'll probably end up moving the `sock = socks.socksocket()` object under the SSH exclusively, but this at least gives you an idea how it currently works when I was testing it.

ktbyers:
Here:

```
		       # Telnet sock dictionary.
			sock = {
				'proxy_type': socks.SOCKS5,
				'proxy_addr': self.socks_source_ip,
				'proxy_port': self.socks_source_port,
				}

```

proxy_type is always socks.SOCKS5--correct?

Okay, let me think about how to resolve making mypy happy (since we are passing in two very different things using that argument).

Gatorjosh14:
It doesn't have to be, but that's what I was using, I think I actually gave an example dictionary in the newly added telnet_proxy,py module comments.
```
def open(self, host, port=0, timeout=socket._GLOBAL_DEFAULT_TIMEOUT):
        """
        Connect to a host.
        The optional second argument is the port number, which
        defaults to the standard telnet port (23).

        Don't try to reopen an already connected instance.

        proxy_dict = {
                'proxy_type': socks.SOCKS5,
                'proxy_addr': hostname,
                'proxy_port': port,
                'proxy_username': username,
                'proxy_password': password
            }
        """
        self.eof = 0
        if not port:
            port = telnetlib.TELNET_PORT
        self.host = host
        self.port = port
        self.timeout = timeout

        sys.audit("telnetlib.Telnet.open", self, host, port)

        self.sock = socks.create_connection((host, port), timeout, **self.proxy_dict)
```
You can use the following for PySocks:

https://pypi.org/project/PySocks/
SocksiPy is compatible with three different types of proxies:

SOCKS Version 4 (SOCKS4), including the SOCKS4a extension.
SOCKS Version 5 (SOCKS5).
HTTP Proxies which support tunneling using the CONNECT method.

Gatorjosh14:
You could change the sock argument to `sock_ssh` which expects a `socket.socket` object and then `sock_telnet` which expects a dictionary, just a thought.  I was trying to not change much code in the `base_connection.py` module as much as possible, didn't really consider the Typing issues that would arise.

ktbyers:
I will probably just add a separate `sock_telnet` argument. I think that will be cleaner and leave the current argument as-is (and avoid overloading the argument with two very dissimilar things).

ktbyers:
@Gatorjosh14 Do you want to regression test this and verify it works properly?

ktbyers:
Note, I changed the argument that you pass in to ConnectHandler to `sock_telnet`.

Gatorjosh14:
Sure I'll try to test it next week as soon as I have time to set it up in my code.

Gatorjosh14:
I've tested out the new code and everything checks out, should be good to merge.

PR #3360, opened 2023-12-21
added oneaccess_tdre

Headknot:
Hi,

Using netmiko version 4.1.2 I have used the device type "oneaccess_oneos" when logging into OneAccess hardware models 1424, 1623, 1645 & 1646.

I recently noticed however that these hardware models are running the OS "TDRE", not "OneOS".
In version 4.2.0 the "oneaccess_oneos" code is different from 4.1.2 and so if i use "oneaccess_oneos" in 4.2.0 I get "Pattern not detected: '(\\#|>)' in output." when logging into OneAccess devices with TDRE.

I think it would be logical to create a separate device type for "oneaccess_tdre" that has pretty much the same code as in 4.1.2, but removing the commands from set_terminal_width and disable_paging as they do not work for TDRE. I also think it should use BaseConnection instead of CiscoBaseConnection as it doesn't have enable or config modes like Cisco.

Br,
Alex


Noppes:
Was this one fixed in #3080?

ktbyers:
@Headknot Do you still have access to one of these devices for testing?

If so, would you be able to run the `show` and `config` test suites against a QA/test device?

Testing process is here:

@Noppes I don't think this was fixed in in #3080, but a bit hard to tell (since documentation of TDRE is hard to come by).

ktbyers:
@Headknot I am going to close this--if you want to work on it again, just let me know and we can re-open it.

PR #3394, opened 2024-03-04
IBM Cloud fortigate driver for non-global commands

SaneiSaya:
Adding a driver that will work with IBM Cloud 1500D as users are restricted and not able to run global commands. This driver allows non-global commands to be run

### DIFF
--- a/netmiko/fortinet/fortinet_ssh_ibmcloud.py
+++ b/netmiko/fortinet/fortinet_ssh_ibmcloud.py
@@ -158,7 +158,7 @@ Alternatively you can try configuring 'configure system console -> set output st
         if result_mode_re:
             result_mode = result_mode_re.group("mode").strip()
             if result_mode in ["more", "standard"]:
-                return result_mode
+                return "more"
 
         raise ValueError("Unable to determine the output mode on the Fortinet device.")
 
@@ -180,11 +180,10 @@ Alternatively you can try configuring 'configure system console -> set output st
         pattern = r"output\s+:\s+(?P<mode>\S+)\s*$"
         result_mode_re = re.search(pattern, output, flags=re.M)
         if result_mode_re:
-            result_mode = result_mode_re.group("mode").strip()
+            result_mode = "more"
             if result_mode in ["more", "standard"]:
                 return result_mode
 
-        raise ValueError("Unable to determine the output mode on the Fortinet device.")
 
     def _get_output_mode(self) -> str:
         """Save the state of the output mode so it can be reset at the end of the session."""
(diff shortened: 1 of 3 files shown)


SaneiSaya:
@ktbyers Can you check this again after the changes I made?

ktbyers:
@SaneiSaya So you are trying to get the fortinet driver to work, but without needing any `global` commands--correct?

SaneiSaya:
Yes, we don't have access to run any global commands as we only have VDOM
access.

ktbyers:
@SaneiSaya Sorry for very delayed response--is this still an issue that you are looking to fix/work on?

SaneiSaya:
Yes, I would still like to get this in.

netmiko/fortinet/fortinet_ssh.py
```diff
@@ -166,9 +166,9 @@ def _get_output_mode_v6(self) -> str:
         pattern = r"^\s+set output (?P<mode>\S+)\s*$"
         result_mode_re = re.search(pattern, output, flags=re.M)
         if result_mode_re:
-            result_mode = result_mode_re.group("mode").strip()
-            if result_mode in ["more", "standard"]:
-                return result_mode
+            return "more"
+        else:
+            return "more"
```
ktbyers, line 171: I am confused as to why you made this change? Why did you make any changes to the non-IBM driver?

netmiko/fortinet/fortinet_ssh_ibmcloud.py
```diff
@@ -0,0 +1,220 @@
...
+    ) -> str:
+        """Disable paging is only available with specific roles so it may fail."""
+
+        output = ""
+        if self._output_mode == "standard":
+            # Do nothing - already correct.
+            return ""
+
+        if self._vdoms:
+            output += self._config_global()
```
ktbyers, line 116: I didn't understand this? I thought your purpose was to get rid of `config global`, but it still does `config global` (depending on VDoms)?

ktbyers:
How do you want this to behave?

So we can't execute `config global` and we are using `config global` to determine and configure the following:
1. Config Output Mode to "standard" in disable_paging() when VDOMs are in use.
2. Determine current output mode "more" or "standard" when VDOMs are in use for V6 or earlier (_get_output_mode_v6)
3. Determine current output mode "more" or "standard" when VDOMs are in use for V7 or later (_get_output_mode_v7)
4. Restore the original output mode in cleanup(), if it was changed in step1 (and VDOMs are used).

So one working option is for the end-user to set an environment variable named `NETMIKO_FORTINET_STANDARD=True` and then if this is set we bypass all four of the above items. 

In other words, the end-user configure the device so there is no output paging and then set the above environment variable. Once this, happens we can bypass all of the above.

I don't see how your current PR fixes the original issue you were running into?


Regards, Kirk

ktbyers:
@SaneiSaya I am going to close this as it doesn't look like it is actively being worked on any longer. If you want to continue working on it, let me know and we can re-open it.

PR #3395, opened 2024-03-05
Add pagination support

cknost:
Add pagination support.
Needed to use ruckus devices without enable mode.

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1951,6 +1951,13 @@ You can also look at the Netmiko session_log or debug log for more information.
         :param output: The returned output as a result of the command string sent to the device
         :type output: str
         """
+        backspace_char = "\x08 "
+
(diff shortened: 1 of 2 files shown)


ktbyers:
Netmiko generally requires the ability to disable output paging.

In order to really do this, we would need a lot more things to be added (test cases for all of the relevant platforms), support wired in for all the relevant platforms, reasonable regression testing against large set of platforms. This is a lot of work and I am probably not willing to do it.

You could probably bypass the `session_preparation` by setting `auto_connect=False` and then either call `netmiko_object._open()` and catch the exception and then write your own handler for output pagination. Either that or superclass the ruckus driver and insert your own session_preparation.

Noppes:
@cknost In the commit you add next_page_string and remove the enable/pagination disabling from the ruckus, but I dont see where you are setting the new next_page_string value for it?

PR #3397, opened 2024-03-07
Update ssh_autodetect.py to support h3c device auto detect.

Mr-Jayden:
When I used the autodetect method on H3C devices, the return was empty, so I added H3C related content to SSH_MAPPER_DICT

### DIFF
--- a/netmiko/ssh_autodetect.py
+++ b/netmiko/ssh_autodetect.py
@@ -166,6 +166,12 @@ SSH_MAPPER_DICT = {
         "priority": 99,
         "dispatch": "_autodetect_std",
     },
+    "h3c": {
+        "cmd": "display version",
+        "search_patterns": ["H3C Comware Software"],
+        "priority": 99,
+        "dispatch": "_autodetect_std",
+    },
     "hp_comware": {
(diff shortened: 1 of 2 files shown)


netmiko/ssh_autodetect.py
```diff
@@ -166,6 +166,12 @@
         "priority": 99,
         "dispatch": "_autodetect_std",
     },
+    "h3c": {
```
ktbyers, line 169: There is no `h3c` device_type in Netmiko. 

This probably should return `hp_compware` or we should just put in another key to the `ssh_dispatcher.py` mapping `h3c_compware` to the `HPComwareSSH` class.

We should probably do the latter.

Do you want to implement that fix? If not, I will update this PR and implement it (just let me know).

Mr-Jayden:
I have added h3c mapping according to your requirements. If I later discover that h3c no longer works through hp_comware, I will create and write the H3CComwareSSH class. I hope to get your consent.

ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/3632

PR #3402, opened 2024-03-13
Update ssh_autodetect.py added support to BIG-IQ

tramontano:
Added support to BIG-IQ variance from the F5's BIG-IP family


ktbyers:
Does the Big IQ actually work with the `f5_tmsh` driver?

Did you regression test the `show` and `config` tests again a Big IQ device? Please post the output here?

PR #3403, opened 2024-03-13
Update f5_tmsh_ssh.py fixed "quit"

tramontano:
Fixed "quit" closing session command, nothing is expected since the session is terminated immediately.

### DIFF
--- a/netmiko/f5/f5_tmsh_ssh.py
+++ b/netmiko/f5/f5_tmsh_ssh.py
@@ -22,7 +22,7 @@ class F5TmshSSH(NoConfig, BaseConnection):
         return output
 
     def exit_tmsh(self) -> str:
-        output = self._send_command_str("quit", expect_string=r"#")
+        output = self._send_command_str("quit", expect_string=r"")
         self.set_base_prompt()
         return output
 


netmiko/f5/f5_tmsh_ssh.py
```diff
@@ -22,7 +22,7 @@ def tmsh_mode(self, delay_factor: float = 1.0) -> str:
         return output
 
     def exit_tmsh(self) -> str:
-        output = self._send_command_str("quit", expect_string=r"#")
+        output = self._send_command_str("quit", expect_string=r"")
```
ktbyers, line 25: A null expect string is not valid/correct. 

Also it looks like whether `quit` causes the SSH to terminate `depends`:

```
You can exit tmsh by entering quit. Two different results may occur, depending on the user's terminal access type:

If you are granted bash access, you will be returned to bash.
If you are granted tmsh access, you will be logged out of the BIG-IP system.
```

From https://my.f5.com/manage/s/article/K12029#:~:text=You%20can%20exit%20tmsh%20by,will%20be%20returned%20to%20bash.

So would probably need to handle both cases (likely via a try/except block).

PR #3420, opened 2024-04-26
Update cisco_xr.py exit_config_mode function

rajjaisur:
Cisco_xr exit_config_mode function in case of uncommitted changes error.
The current function is not working in case of uncommitted changes.  because the output is until "uncommitted". function matching is for "Uncommitted changes found" .


ktbyers:
You are going to have to expand on the issue you are seeing i.e. the current code should work with this string from IOS-XR:

```
RP/0/0/CPU0:ios-xrv(config)#end
Uncommitted changes found, commit them before exiting(yes/no/cancel)? [cancel]:
```

ktbyers:
@rajjaisur Any updates on my question on why this change is needed?

rajjaisur:
No it will not because current code is only reading till Uncommitted not full sentence.

ktbyers (approved):

Issue #3424, opened 2024-04-28
MOTD prevents successful login attempts with CISCO


FriendlyGecko:
### Description of Issue/Question

When attempting to log into a device that uses a MOTD, the login fails. This is because the MOTD requires the user to first send a return character to wake it up. If I use putty to first console in and press enter, I can swap back to my netmiko script and log into the device.

### Setup
1 Laptop 
1 Console Cable
1 Network

### Netmiko version
```
netmiko==4.3.0
```

### Netmiko device_type (if relevant to the issue)
```
Cisco IOS, Cisco IOS Serial
```

### Steps to Reproduce the Issue
1. Configure a Cisco Network device with local login for line console 0
2. Configure a MOTD banner
3. [OPTIONAL] Configure SSH
4. Attempt to connect to the device

### Error Traceback
```
Traceback (most recent call last):
  File "C:\Users\User\PycharmProjects\intChange\test.py", line 15, in <module>
    net_connect = netmiko.ConnectHandler(**cisco1)
  File "C:\Users\User\PycharmProjects\intChange\venv\lib\site-packages\netmiko\ssh_dispatcher.py", line 399, in ConnectHandler
    return ConnectionClass(*args, **kwargs)
  File "C:\Users\User\PycharmProjects\intChange\venv\lib\site-packages\netmiko\base_connection.py", line 488, in __init__
    self._open()
  File "C:\Users\User\PycharmProjects\intChange\venv\lib\site-packages\netmiko\base_connection.py", line 493, in _open
    self.establish_connection()
  File "C:\Users\User\PycharmProjects\intChange\venv\lib\site-packages\netmiko\base_connection.py", line 1129, in establish_connection
    self.serial_login()
  File "C:\Users\User\PycharmProjects\intChange\venv\lib\site-packages\netmiko\cisco_base_connection.py", line 76, in serial_login
    return self.telnet_login(
  File "C:\Users\User\PycharmProjects\intChange\venv\lib\site-packages\netmiko\cisco_base_connection.py", line 194, in telnet_login
    raise NetmikoAuthenticationException(msg)
netmiko.exceptions.NetmikoAuthenticationException: Login failed: serial
```

### Relevant Python code
```python
import netmiko

cisco1 = {
    "username": "admin",
    "password": "password",
    "device_type": "cisco_ios_serial",
    "serial_settings": {"port": "COM5"}
}

net_connect = netmiko.ConnectHandler(**cisco1)
[...]

PR #3425, opened 2024-04-28
Added MOTD Wake Up

FriendlyGecko:
This may not be the best way to solve the issue, but it is a way.

## Issue we are trying to solve
When attempting to log into a device that uses a MOTD, the login fails. This is because the MOTD requires the user to first send a return character to wake it up. If I use putty to first console in and press enter, I can swap back to my netmiko script and log into the device without issue.

To solve the issue, I added a check (checks if the login process is already started) that will send a return character to the device to first wake it up before going to into the login process. If the user is already logged in, then all this will do is return a blank line (unless there is text in the prompt somehow).

This may be better, but taking critiques on it:

```
                try:
                    output = self.read_channel()
                    return_msg += output

                    # Search for username pattern / send username
                    if re.search(username_pattern, output, flags=re.I):
                        # Sometimes username/password must be terminated with "\r" and not "\r\n"
                        self.write_channel(self.username + "\r")
                        time.sleep(1 * delay_factor)
                        output = self.read_channel()
                        return_msg += output

                    # Search for password pattern / send password
                    if re.search(pwd_pattern, output, flags=re.I):
                        # Sometimes username/password must be terminated with "\r" and not "\r\n"
                        assert isinstance(self.password, str)
                        self.write_channel(self.password + "\r")
                        time.sleep(0.5 * delay_factor)
                        output = self.read_channel()
                        return_msg += output
                        if re.search(
                            pri_prompt_terminator, output, flags=re.M
                        ) or re.search(alt_prompt_terminator, output, flags=re.M):
                            return return_msg

                    # Support direct telnet through terminal server
                    if re.search(
                        r"initial configuration dialog\? \[yes/no\]: ", output
                    ):
                        self.write_channel("no" + self.TELNET_RETURN)
                        time.sleep(0.5 * delay_factor)
                        count = 0
                        while count < 15:
                            output = self.read_channel()
                            return_msg += output
                            if re.search(r"ress RETURN to get started", output):
                                output = ""
                                break
                            time.sleep(2 * delay_factor)
                            count += 1

                    # Check for device with no password configured
                    if re.search(r"assword required, but none set", output):
                        assert self.remote_conn is not None
                        self.remote_conn.close()
                        msg = (
                            "Login failed - Password required, but none set: {}".format(
                                self.host
                            )
                        )
                        raise NetmikoAuthenticationException(msg)

                    # Check if proper data received
                    if re.search(
                        pri_prompt_terminator, output, flags=re.M
                    ) or re.search(alt_prompt_terminator, output, flags=re.M):
                        return return_msg

                    # MOTD Wake Up
                    self.write_channel("\r")
                    time.sleep(1 * delay_factor)
                    output = self.read_channel()
                    return_msg += output

                    i += 1
```


FriendlyGecko:
This resolves #3424

FriendlyGecko:
I moved this to a draft, because it warrants a conversation on if this is the best way to resolve the issue or if it should be moved/tweaked

ktbyers:
@FriendlyGecko Which platform does this apply to?

Also is this when using a Console or directly via SSH to the device?

FriendlyGecko:
I will set up another lab and test it again, but as far as I remember I ran into the issue with both console and SSH, and this solved both. Please give me a little time to confirm though.

FriendlyGecko:
UPDATE: Latest update may make this pull request irrelevant. With my current home lab I was unable to replicate the issue that the pull request was intended to solve. I will test again with my previous lab which have a different IOS and see if that is related. 

If I am unable to replicate it there then I will close the pull request and the associated issue.

PR #3429, opened 2024-05-03
Garderos Driver

ktbyers:


ktbyers:
[enricostraehler](https://github.com/enricostraehler) Do you have a chance to regression test this and see if the tests still pass?

ktbyers:
I went ahead and merged this (after I did some additional review and cleanup).

We can fix / correct issues that we run into.

PR #3434, opened 2024-05-16
Add find_failover to cisco_asa, cisco_ftd, a10, f5 and paloalto

romio28:
Add find_failover to cisco_asa, cisco_ftd, a10, f5 and paloalto


ktbyers:
@romio28 Linting and tests are failing. Can you fix these?

PR #3440, opened 2024-06-12
fastiron - session preparation / _test_channel_read() not working after successful connection

johanek:
Hello,

On Brocade ICX 6450 I have be getting the following error when running `session_preparation()`:

```
---------------------------------------------------------------------------
ReadTimeout                               Traceback (most recent call last)
<ipython-input-6-2d99eca55c84> in <module>
----> 1 device._test_channel_read(pattern='')

~/.local/share/virtualenvs/napalm-examples-yv4OVgxK/lib/python3.9/site-packages/netmiko/base_connection.py in _test_channel_read(self, count, pattern)
   1239         new_data = ""
   1240         while i <= count:
-> 1241             new_data += self.read_channel_timing(read_timeout=20)
   1242             if new_data:
   1243                 return new_data

~/.local/share/virtualenvs/napalm-examples-yv4OVgxK/lib/python3.9/site-packages/netmiko/base_connection.py in read_channel_timing(self, last_read, read_timeout, delay_factor, max_loops)
    821
    822 """
--> 823             raise ReadTimeout(msg)
    824         return channel_data
    825

ReadTimeout:

read_channel_timing's absolute timer expired.

The network device was continually outputting data for longer than 20
seconds.

If this is expected i.e. the command you are executing is continually emitting
data for a long period of time, then you can set 'read_timeout=x' seconds. If
you want Netmiko to keep reading indefinitely (i.e. to only stop when there is
no new data), then you can set 'read_timeout=0'.

You can look at the Netmiko session_log or debug log for more information.
```

_test_channel_read() doesn't seem to find anything sent by the device, it seems to maybe have missed the prompt data from the previous command. If you skip running session_preparation you can successfully run commands on the device.

So in this change I'm sending an empty command to force the device to send the prompt again, and _test_channel_read will then succeed. 

If this isn't the right way to handle this, or there's more information you'd like then let me know 😊

### DIFF
--- a/netmiko/ruckus/ruckus_fastiron.py
+++ b/netmiko/ruckus/ruckus_fastiron.py
@@ -12,6 +12,7 @@ class RuckusFastironBase(CiscoSSHConnection):
 
     def session_preparation(self) -> None:
         """FastIron requires to be enable mode to disable paging."""
+        self.send_command("\n")
         self._test_channel_read()
         self.set_base_prompt()
         self.enable()


johanek:
This didn't really fix my issue. The problem was with setting up the connection in the napalm driver.

Issue #3177, opened 2023-04-24
telnetlib is deprecated


sleinen:
When adding pytest coverage to my own script that uses Netmiko, I noticed the following warning:

```
=================================================================================== warnings summary ====================================================================================
.venv/lib/python3.11/site-packages/netmiko/base_connection.py:30
  /Users/leinen/network/bgp-observer/.venv/lib/python3.11/site-packages/netmiko/base_connection.py:30: DeprecationWarning: 'telnetlib' is deprecated and slated for removal in Python 3.13
    import telnetlib

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
[...]
```

telnetlib is indeed one of several network-related modules that are scheduled for removal in Python 3.13, according to [PEP 594](https://peps.python.org/pep-0594/#deprecated-modules).  The table mentions [telnetlib3](https://telnetlib3.readthedocs.io/en/latest/) and [Exscript](https://pypi.org/project/Exscript/) as possible replacements.

I haven't looked at the two in detail, but according to [this StackOverflow response](https://stackoverflow.com/questions/75778543/python-telnetlib3-examples/75779938#75779938), Exscript could actually be closer to the original telnetlib than telnetlib3 is to telnetlib.  But I can't tell what option would make the most sense for Netmiko going forward.

GoreNetwork:
sadly gear that doesn't support current SSH is still widly deployed.

JohnVillalovos:
One suggestion in https://discuss.python.org/t/pep-594-take-2-removing-dead-batteries-from-the-standard-library/13508/59

Is to just copy `telnetlib.py` and use it. 

Would have to comply with their license: https://docs.python.org/3/license.html

ktbyers:
@JohnVillalovos Yeah, I guess we could just "vendor" it inside of Netmiko and copy their license inside that folder (and make a note on our main license).

AdamWill:
This is now rather more than an RFE, as telnetlib is removed in Python 3.13, which will be released in October (and is already in Fedora Rawhide and probably pending for other development distros).

dtantsur:
@ktbyers sorry for bothering you, any progress here? I'm stuck with Fedora packaging because of this issue. I can vendor telnetlib myself, but I'd rather at least coordinate with you (especially because of the licensing question).

dtantsur:
[...]

PR #3444, opened 2024-06-24
Vendor telnetlib library

ktbyers:
Fixes https://github.com/ktbyers/netmiko/issues/3177


ktbyers:
TODO: Regression test on PY3.8 through PY3.12 to some real devices using telnet.

ktbyers:
pyats/genie are causing PY3.12 to break. I will have to look at this separately.

ktbyers:
@dtantsur @adamwill

Let me know if you want to test and verify this will work for your use-case.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["poetry==1.3.2"]
+requires = ["poetry==1.8.3"]
```
dtantsur, line 2: Side comment: I'm forced to override this in the RPM packaging since the chances of having exactly this version of Poetry are not high.

dtantsur:
@ktbyers as long as it passes tests under Python 3.13, it's good for me, thanks much! Do you want to update `license-dependencies.txt`?

pyproject.toml
ktbyers: Okay, I will relax to make it >= 1.6.1 or something close to that. There was an issue with PY3.13 and Poetry ~1.3.1 (IIRC).

ktbyers:
@dtantsur Okay, specified a wider range for Poetry versions; cleaned up `license-dependencies.txt`, and broke out separate tests for the parsers (pyats/genie).

So most of the testing now works on PY38 through PY313

Still need to do some manual telnet testing using this branch before I want to merge, but I think everything else is good-to-go from my perspective.

dtantsur (approved):
This is great, thank you!

ktbyers:
Tested cisco_ios and telnet using PY3.11 (config and show commands)

dtantsur:
@ktbyers thank you so much for your cooperation! Could you make a release with this change? I can pick up the version from git, but it's somewhat easier with an official release.

ktbyers:
@dtantsur FYI, this version of Netmiko was released today to pypi (Netmiko 4.4.0)

dtantsur:
Yep, thank you, already uploaded to Fedora!

PR #3448, opened 2024-06-26
Fix aruba cx device_type naming

ktbyers:


ktbyers:
@Trailingslashes This is the same as what you had except I converted the <enters> back to standard newlines (unix form) as that is what the code base uses (and switching messed up the Git/GitHub history as every line in the file was marked as changed).

PR #3453, opened 2024-07-08
Add code 'ESC[c' into method 'strip_ansi_escape_codes'

max-dw-i:
Some 'MikroTik' devices add ANSI escape codes 'ESC[c' to the end of the output when in 'safe-mode'

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -2361,7 +2361,8 @@ You can also look at the Netmiko session_log or debug log for more information.
         ESC[6n       Get cursor position
         ESC[1D       Move cursor position leftward by x characters (1 in this case)
         ESC[9999B    Move cursor down N-lines (very large value is attempt to move to the
-                     very bottom of the screen).
+                     very bottom of the screen)
+        ESC[c        Unknown code (added, for example, by MikroTik in 'Safe-Mode').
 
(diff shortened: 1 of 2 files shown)


ktbyers:
@max-dw-i Can you show me an output capture with this character sequence? I am not seeing/finding any references to `ESC[c` looking at ANSI reference codes?

max-dw-i:
Yeah, I couldn't find either but as far as I understand there might be not typical codes.
Here is an output example:
1. We send `\u0018` (ctrl-x) to enter the `safe-mode` via `send_command_timing`:
```
[Safe Mode taken]
[user@HOSTNAME] <SAFE>
```
2. Then run any command:
```
/system identity print

name: HOSTNAME
\x1b[9999B[user@HOSTNAME] <SAFE> \x1b[c
```

ktbyers:
Looks like it is probably:

```
Device Status
The following codes are used for reporting terminal/display settings, and vary depending on the implementation:
Query Device Code	<ESC>[c
Requests a Report Device Code response from the device.
```

From here:

https://www2.ccs.neu.edu/research/gpc/VonaUtils/vona/terminal/vtansi.htm

ktbyers:
Superceded by: 

https://github.com/ktbyers/netmiko/pull/3457

PR #3458, opened 2024-07-08
Mikrotik free version ssh login fix (with minor updates)

ktbyers:


ktbyers:
@meganerd This should be the same as what you originally submitted. I made a few minor adjustments, but the behavior should be the same.

Let me know if you see any issues.

PR #3462, opened 2024-07-10
Alaxala 2600s and 3600s driver

ktbyers:


ktbyers:
cc: [inaba-vdom-0](https://github.com/inaba-vdom-0)

ktbyers:
FYI, I removed the autodetect code (as I thought it was too generic). We can revisit that in a separate PR if you are interested.

inaba-vdom-0:
@ktbyers Thank you. I don't have any other switches so I can't test it, but we'll see how it goes.

Issue #3405, opened 2024-03-19
Mikrotik version > 6.47 = 'export' command incomplete 


prtomasi:
### Description of Issue/Question

*Note*: Please check https://guides.github.com/features/mastering-markdown/
to see how to properly format your request.

### Netmiko version
(Paste verbatim output from `pip freeze | grep netmiko` between quotes below)

```
netmiko==4.2.0
```

### Netmiko device_type (if relevant to the issue)
(Paste `device_type` between quotes below)

```
    'device_type': 'mikrotik_routeros',
```

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```
no error, just incomplete output from 'export' command
```

### Relevant Python code
(Please try to essentialize your Python code to the minimum code needed to reproduce the issue)
(Paste the code between the quotes below)

```python
from netmiko import ConnectHandler

mikrotik = {
    'device_type': 'mikrotik_routeros',
    'host':   '10.1.1.68',
    'port':     2222,
    'username': 'dude',
    'password': 'not_real_password'
}

mikrotik_connection = ConnectHandler(**mikrotik)
print(mikrotik_connection.send_command(f'export', cmd_verify=False))

```
### Test

I noticed netmiko is breaking the output of 'export' command when the Mikrotik RouterOS version is > 6.47

The following was tested with  6.49.8 version
```# mar/19/2024 11:25:34 by RouterOS 6.49.8```

With the python code above, only 27 lines (from a total of 275 lines) were captured by netmiko, no errors were generated
```
line 22 --> /interface vlan
line 23 --> add disabled=yes interface=sfp-sfpplus1 name=v450 vlan-id=450
line 24 --> add interface=sfp-sfpplus1 name=v495 vlan-id=495
line 25 --> add interface=sfp-sfpplus1 name=v1904 vlan-id=1904
line 26 --> add interface=sfp-sfpplus1 name=v1906 vlan-id=1906
line 27 --> add interface=sfp-sfpplus1 name=v2056 vlan-id=2056
```
#
When I use PuTTY to run 'export' command in the same device, the output is complete and has 275 lines
```
...
line 28 --> /interface ethernet switch port
...
line 275 --> set default-group=deny use-radius=yes
```

line 278 --> prompt:
```
[user@RTR-FW-NNN-NNN-01] >
```
#

I asked for help at Mikrotik Forum, but I was ignored:

[https://forum.mikrotik.com/viewtopic.php?t=205175](Forum)

[...]

PR #3467, opened 2024-07-27
Bugfix Mikrotik version > 6.48 = 'export' command incomplete #3405 

Tonygratta:
This fix generates a regex string for search pattern using in send_command function for MikroTik devices. This regex expression searches not just the prompt string, but the prompt string exactly at the end of the output string, without any symbols after it. It prevents premature termination of the collecting data process.

Checked on hardware:
- RouterBOARD 941-2nD FW v6.42.1, 6.49.7, 6.49.15 
- RouterBOARD 951-2n FW v6.49.7


ktbyers:
Does just switching to dumb terminal setting make this change unnecessary?

See:
https://github.com/ktbyers/netmiko/issues/3405

Tonygratta:
Tried that, but unfortunately it doesn't work.

Issue #3469, opened 2024-08-02
netmiko import pysnmp error


ghost:
### Description of Issue/Question

I have netmiko4.4.0 and pysnmp6.2.4 installed, and when I run the project, I get an error: pysnmp not installed; please install it: 'pip install pysnmp'
The investigation found that:
The source code is like this package import:
from pysnmp.entity.rfc3413.oneliner import cmdgen
In pysnmp6.2.4, cmdgen has a cmdgen.py file in the path: pysnmp/entity/rfc3413
netmiko4.4.0 was released on July 9, 2024.
pysnmp&gt; 5.* versions are all released after July 17, 2024.
If I use pysnmp earlier than 6.2.2, there is no problem, it appears that the pysnmp author changed the file structure in 6.2.2

### Netmiko version

```
netmiko==4.4.0
```

### Netmiko device_type (if relevant to the issue)
(Paste `device_type` between quotes below)

```
pysnmp==6.2.4
```

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```
File "/home/new_autonet/autonet/urls.py", line 19, in <module>
    from application.config_task_manage.views import EmailSignView
  File "/home/new_autonet/application/config_task_manage/views.py", line 17, in <module>
    from core.system_manage.system_handle import get_session_aaa_data
  File "/home/new_autonet/core/system_manage/system_handle.py", line 18, in <module>
    from utils.common.auths import NetAuths
  File "/home/new_autonet/utils/common/auths.py", line 25, in <module>
    from netmiko.snmp_autodetect import SNMPDetect
  File "/usr/local/lib/python3.9/site-packages/netmiko/snmp_autodetect.py", line 32, in <module>
    raise ImportError("pysnmp not installed; please install it: 'pip install pysnmp'")
ImportError: pysnmp not installed; please install it: 'pip install pysnmp'
```

### Relevant Python code
(Please try to essentialize your Python code to the minimum code needed to reproduce the issue)
(Paste the code between the quotes below)

```python

```

ktbyers:
Okay...makes sense for 5+ years there were no updates to pysnmp and then in July 2024 they released a version 5 and a version 6.

So what is needed.
* A PR to add support for pysnmp version 6 (and maybe 5) in Netmiko

In order to use the current released version of Netmiko, a user would need to install and use the older version of pysnmp (since it looks like they made a backwords incompatible change).

Tonygratta:
[...]

PR #3472, opened 2024-08-06
Using async pysnp getCmd

Tonygratta:
Bugfix #3469 The sync pysnmp API has been deprecated and removed since version 6.2.0. This commit replaces sync getCmd with async getCmd pysnmp method.

### DIFF
--- a/netmiko/snmp_autodetect.py
+++ b/netmiko/snmp_autodetect.py
@@ -25,9 +25,10 @@ from typing import Optional, Dict, List
 from typing.re import Pattern
 import re
 import socket
+import asyncio
 
 try:
-    from pysnmp.entity.rfc3413.oneliner import cmdgen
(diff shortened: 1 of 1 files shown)


ktbyers:
@Tonygratta I didn't realize this had been moved to an async solution. This won't work for Netmiko which is entirely synchronous.

ktbyers:
Let me look at this some more...

ktbyers:
I think we need to make this backwards compatible i.e. try to use the old solution first and then only try the new solution if the old doesn't work.

I say this as the new solution is very new and they changed the code meaningfully (i.e. they converted over to an async solution).

Let me know if you want to take a first pass on this, if not, I will do it.

ktbyers:
FYI, I started working on this.

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3473

Issue #3482, opened 2024-08-19
Arista driver with SCP driver and bash >= 5.1 encounters new ANSI escape code


jniec-js:
### Description of Issue/Question

Beginning with bash 5.1, bracketed-paste-mode is enabled by default: https://forum.endeavouros.com/t/bash-5-1-bracketed-paste-now-on-by-default/10258

This causes issues when trying to scp a file to a remote unix system using bash 5.1, as the first line is now escape characters instead of the expected header line as the source code indicates here: https://github.com/ktbyers/netmiko/blob/f0041bb1a8526f054292af2174c3412931ee565e/netmiko/scp_handler.py#L177

The actual output Netmiko sees is:
```Parsing error, unexpected output from /bin/df -k /mnt/flash:
[0m[?2004l
Filesystem     1K-blocks    Used Available Use% Mounted on
/dev/mmcblk0p1   7409788 3047916   4345488  42% /mnt/flash
```

### Netmiko version

```
In [2]: netmiko.__version__
Out[2]: '4.3.0'
```

### Netmiko device_type (if relevant to the issue)
(Paste `device_type` between quotes below)

```
arista_eos
```
Note: Starting in 4.32 train, Arista has upgraded the version of bash to 5.1 on their devices, so this requires a device running EOS 4.32 or higher. 

```
 bash --version
GNU bash, version 5.1.8(1)-release (x86_64-redhat-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

### Error Traceback
(Paste the complete traceback of the exception between quotes below)

```
Traceback (most recent call last):
  File "/...snip.../tmp/netmiko-debugging/./repro.py", line 25, in <module>
    netmiko.file_transfer(cur, source_file="test.txt", dest_file="test.txt", file_system="/mnt/flash")
  File "/...snip.../lib/python3.10/site-packages/netmiko/scp_functions.py", line 149, in file_transfer
    verifyspace_and_transferfile(scp_transfer)
  File "/...snip.../lib/python3.10/site-packages/netmiko/scp_functions.py", line 48, in verifyspace_and_transferfile
    if not scp_transfer.verify_space_available():
  File "/...snip.../lib/python3.10/site-packages/netmiko/scp_handler.py", line 213, in verify_space_available
[...]

PR #3483, opened 2024-08-20
Add code_disable_bracketed_paste_mode to set of ansi escape codes to strip and strip ansi escape codes on Arista EOS devices

joshniec:
Closes https://github.com/ktbyers/netmiko/issues/3482

Beginning in Bash 5.1, the default is to enable bracketed-paste-mode, which results in escape code characters being sent back to the Netmiko client. 

This PR adds `disable-bracketed-paste-mode` to the set of ansi escape codes that are stripped from the response from devices and updates the session preparation method for Arista devices to strip ansi escape codes from the response from Arista EOS devices. Beginning in the Arista EOS 4.32 train, the default bash version used is 5.1.

### DIFF
--- a/netmiko/arista/arista.py
+++ b/netmiko/arista/arista.py
@@ -13,6 +13,7 @@ class AristaBase(CiscoSSHConnection):
 
     def session_preparation(self) -> None:
         """Prepare the session after the connection has been established."""
+        self.ansi_escape_codes = True
         self._test_channel_read(pattern=self.prompt_pattern)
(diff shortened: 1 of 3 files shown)


jniec-js:
@ktbyers wondering if you could approve the workflows to run/review and include this in the next release? This is blocking/breaks our tooling that relies heavily on Netmiko. 

Thank you,

ktbyers (approved):

Issue #3209, opened 2023-05-15
scp_handler.py   process_md5() failed with linux ubuntu 22.04 lts


zcy382725:
log:

{'device_type': 'linux', 'host': 'some_ip', 'username': 'root', 'use_keys': True, 'key_file': 'KEY', 'ssh_config_file': './ssh_config'}
Traceback (most recent call last):
  File "/builds/infra/sw_dump/dump.py", line 35, in <module>
    output = file_transfer(net_connect,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/netmiko/scp_functions.py", line 122, in file_transfer
    with TransferClass(**scp_args) as scp_transfer:
         ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/netmiko/ssh_dispatcher.py", line 503, in FileTransfer
    return FileTransferClass(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/netmiko/linux/linux_ssh.py", line 179, in __init__
    super().__init__(
  File "/usr/local/lib/python3.11/site-packages/netmiko/scp_handler.py", line 112, in __init__
    self.remote_md5(remote_file=source_file) if hash_supported else None
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/netmiko/linux/linux_ssh.py", line 214, in remote_md5
    dest_md5 = self.process_md5(dest_md5).strip()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/netmiko/linux/linux_ssh.py", line 219, in process_md5
    return super(LinuxFileTransfer, LinuxFileTransfer).process_md5(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/netmiko/scp_handler.py", line 359, in process_md5
    raise ValueError(f"Invalid output from MD5 command: {md5_output}")
ValueError: Invalid output from MD5 command: 
[...]

PR #3486, opened 2024-08-22
linux_ssh: fix for 'Invalid output from MD5 command'

pestophagous:
I believe that this:

fixes issue #3209

Original credit goes to commenter on the issue: https://github.com/ktbyers/netmiko/issues/3209#issuecomment-2245991018

Locally I am using this change and it fixes the problem.

### DIFF
--- a/netmiko/linux/linux_ssh.py
+++ b/netmiko/linux/linux_ssh.py
@@ -211,7 +211,7 @@ class LinuxFileTransfer(CiscoFileTransfer):
                 remote_file = self.source_file
         remote_md5_cmd = f"{base_cmd} {self.file_system}/{remote_file}"
         dest_md5 = self.ssh_ctl_chan._send_command_str(remote_md5_cmd, read_timeout=300)
-        dest_md5 = self.process_md5(dest_md5).strip()
+        dest_md5 = self.process_md5(dest_md5.strip()).strip()
         return dest_md5
 
     @staticmethod


pestophagous:
I really hope this fix (or any fix for the minor bug) can be merged, because I very much want to use the only-copy-when-necessary feature. It's a great feature. Thank you for thinking of it in the design of your library.

ktbyers (approved):

pestophagous:
@ktbyers thank you for your time and your FOSS service!  What's the best way for me to watch the release rollout and know which release-version (for `requirements.txt`) will contain this fix?

ktbyers:
Yeah, I don't know when I will do the next release.

I assume you can just `git clone` the develop branch or to the specific commit containing that fix?

pestophagous:
> Yeah, I don't know when I will do the next release.
> 
> I assume you can just `git clone` the develop branch or to the specific commit containing that fix?

Ah yes. Branch name or commit will work. Thanks a bunch!

pestophagous:
> Branch name or commit will work. Thanks a bunch!

In case anyone else lands on this webpage and wants to `pip install -r` a version that has this linux md5 fix, this is exactly the line that I ended up with in `requirements.txt`:

```
netmiko @ git+https://github.com/ktbyers/netmiko@eefa4c4cc3 # can use release semver after next release
```

PR #3489, opened 2024-08-26
Change Sophos module allowing the user to navigate the interactive menu prompt.

dLoProdz:
The default functionality of Netmiko is passing an initial carriage return once the connection is established, this was breaking the compatibility with the Sophos interface which displays an interactive prompt where you need to make a numeric selection and press enter, this would set up the session in a bad place to pass the command to reach the Device Console as you need another carriage return before passing the numeric selection.

With the changes made now the user can decide whether to use any of these predefined functions of the interactive main menu, including stepping into the Device Console, or even Advanced Shell, by passing the appropriate menu selections to the connection.

### DIFF
--- a/netmiko/sophos/sophos_sfos_ssh.py
+++ b/netmiko/sophos/sophos_sfos_ssh.py
@@ -15,7 +15,7 @@ SOPHOS_MENU_DEFAULT = os.getenv("NETMIKO_SOPHOS_MENU", "4")
 class SophosSfosSSH(NoEnable, NoConfig, CiscoSSHConnection):
     def session_preparation(self) -> None:
         """Prepare the session after the connection has been established."""
-        self._test_channel_read(pattern=r"Main Menu")
+        self._test_channel_read(pattern=r"Select Menu Number")
         """
         Sophos Firmware Version SFOS 18.0.0 GA-Build339
 
@@ -32,9 +32,7 @@ class SophosSfosSSH(NoEnable, NoConfig, CiscoSSHConnection):
 
             Select Menu Number [0-7]:
         """
-        self.write_channel(SOPHOS_MENU_DEFAULT + self.RETURN)
-        self._test_channel_read(pattern=r"[#>]")
-        self.set_base_prompt()
+        self.send_command_expect("\r", expect_string=r"Select Menu Number")
         # Clear the read buffer
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()


ktbyers:
[capmerah](https://github.com/capmerah) does this change work for you?

ktbyers:
Breaking change...as it leaves your Sophos device in a different state than the previous code (but this probably okay).

dLoProdz:
Thanks @ktbyers! This state change is probably for the best as Sophos has different consoles/shells for different use cases, Sophos admins will probably be happy with this.

capmerah:
Hi @ktbyers , @dLoProdz,

Will test this change tomorrow as I have other projects to finish. 

Thanks for the contribution.

capmerah:
Hi @ktbyers / @dLoProdz ,

Unfortunately, the results are still the same.

dLoProdz:
@capmerah you sure you are using the new code?

Try this:

...
connection.send_command_expect("[numeric_selection]", expect_string=r'Select Menu Number')
...

capmerah:
Hi @dLoProdz , It finally works!

I code the following:
`from netmiko import ConnectHandler
import os

log_directory = rf"C:\Users\name\OneDrive\username\Backup config\Sophos Firewall\Netmiko"
os.makedirs(log_directory, exist_ok=True)


ipaddrs = ["10.1.1.1"]

devices = [
    {
        "device_type": "sophos_sfos_ssh",
        "host": ip,
        "username": "admin",
        "password": "password",
        "session_log": os.path.join(log_directory, f"netmiko_session_{ip}.log"),
    }
    for ip in ipaddrs
]

for device in devices:
    print(f'Connecting to the device: {device["host"]}')

    with ConnectHandler(**device) as net_connect:  
        output = net_connect.send_command_expect("5\n3", expect_string=r'Select Menu Number')
        output = net_connect.send_command_expect("ls")
        print(output)
` 


And it finally show the exact output.

But since the file system for Sophos SG 430 is Linux based. With normal SSH, it will show list of files as the picture below.

Screenshot 2024-08-30 145617

With Netmiko output, it will become jumbled up like the picture below:

Screenshot 2024-08-30 145755

At least it can able to connect to Sophos with Netmiko.
Thanks for the contribution.

ktbyers:
@capmerah There is a good chance that is expected. Netmiko sets the terminal width to be wider (generally) as it is for automation and not intending to be a person.

So with the `ls` command, you would typically programmatically process the output to get all of the file names or do an `ls -1` if you want everything in a single column.

PR #3494, opened 2024-09-03
Add argument to send_command and send_command_timing to support raising exception for parsers

ktbyers:
TextFSM/TTP/Genie failures should result in a NetmikoParsingException when this flag is set and an error occurs.


ktbyers:
@Kircheneer I made some updates to make this more ubiquitous across TTP/Genie.

I also changed the behavior for TextFSM such that NetmikoParsingException is raised both for a failure to find a template and also for a valid template with some set of data that isn't parsed properly (or worded differently CLITable exception won't be raised generally NetmikoParsingException would be raised with a different message).

Let me know if this causes issues (part of the reason for me wanting to do this was to make the behavior more consistent across TTP and Genie).

ktbyers:
@tom0010 See comments to @Kircheneer (let me know if you have any feedback, otherwise, will probably merge tomorrow).

We can obviously adjust after it is merged (if necessary).

Kircheneer (approved):
LGTM

PR #3514, opened 2024-10-23
Update utilities.py

MrVicknair:
added a minor correction to the way arguments get passed to the pkg_resources.path() function

### DIFF
--- a/netmiko/utilities.py
+++ b/netmiko/utilities.py
@@ -319,9 +319,7 @@ Alternatively, `pip install ntc-templates` (if using ntc-templates).
                     if _skip_ntc_package:
                         raise ModuleNotFoundError()
             else:
-                with pkg_resources.path(
-                    package="ntc_templates", resource="parse.py"
-                ) as posix_path:
+                with pkg_resources.path("ntc_templates", "parse.py") as posix_path:
                     # Example: /opt/venv/netmiko/lib/python3.9/site-packages/ntc_templates/templates
                     template_dir = str(posix_path.parent.joinpath("templates"))
                     # This is for Netmiko automated testing


ktbyers:
Why make this change? 

named arguments are generally safer than positional arguments--was there something specific that was breaking that necessitates this?

MrVicknair:
I received an error while attempting to use the named argument 'package' with importlib.resources.

This behavior may be reproduced with python 3.13 , netmiko 4.4.0, setuptools 75.2.0

 TypeError: path() got an unexpected keyword argument 'package'

This behavior can be seen while setting textFSM=True, i.e:
output = net_connect.send_command('show version', use_textfsm=True)

MrVicknair:
@ktbyers  - While the proposed change does fix the issue I encountered, this code will never be ran with properly set NET_TEXTFSM environment value pointing to the ntc_templates\templates directory.

ktbyers:
Okay, I am going to re-open this as I want to reproduce the issue/behavior. I am suspecting it is a Python 3.13 change (that I haven't tested/fixed yet).

ktbyers:
If you are running Python 3.13, this code (which is already in Netmiko) will be run anyways:

```python
        try:
            # New API for Python 3.13+
            if sys.version_info >= (3, 13):
                with pkg_resources.path("ntc_templates", "parse.py") as posix_path:
                    # Example: /venv/netmiko/lib/python3.13/site-packages/ntc_templates/templates
                    template_dir = str(posix_path.parent.joinpath("templates"))
                    # This is for Netmiko automated testing
                    if _skip_ntc_package:
                        raise ModuleNotFoundError()
```

In other words, this fix has already been implemented.

ktbyers:
Fix is here:

https://github.com/MrVicknair/netmiko/commit/92ca035c07fa828b03fb966e3d291a7eae2e3d19

it is only in the develop branch right now.

PR #3515, opened 2024-10-24
Remove unused dependencies

cjwatson:
`setuptools` and `cffi` still appear to be indirect dependencies, but there's no need to depend on them directly when they aren't used directly.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2963,13 +2963,13 @@ jeepney = ">=0.6"
 
 [[package]]
 name = "setuptools"
-version = "75.1.0"
+version = "75.2.0"
 description = "Easily download, build, install, upgrade, and uninstall Python packages"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "setuptools-75.1.0-py3-none-any.whl", hash = "sha256:35ab7fd3bcd95e6b7fd704e4a1539513edad446c097797f2985e0e4b960772f2"},
-    {file = "setuptools-75.1.0.tar.gz", hash = "sha256:d59a21b17a275fb872a9c3dae73963160ae079f1049ed956880cd7c09b120538"},
+    {file = "setuptools-75.2.0-py3-none-any.whl", hash = "sha256:a7fcb66f68b4d9e8e66b42f9876150a3371558f98fa32222ffaa5bced76406f8"},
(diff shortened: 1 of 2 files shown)


ktbyers:
PY3.12 and later requires setuptools to be installed:

https://github.com/TDAmeritrade/stumpy/issues/950

`cffi` I would need to research.

ktbyers:
Things break without `setuptools`...for example:

```
  File "/home/runner/.cache/pypoetry/virtualenvs/netmiko-Jn__E6G5-py3.12/lib/python3.12/site-packages/pylama/lint/__init__.py", line 10, in <module>
    from pkg_resources import iter_entry_points
ModuleNotFoundError: No module named 'pkg_resources'
```

Yes, you could say this isn't a direct Netmiko dependency, but there isn't really a case where I don't want `setuptools` (it is just a pain without it).

cjwatson:
That seems to be https://github.com/klen/pylama/issues/249 (pylama has had no commits at all for two years, so seems to have become unmaintained).  A possibility would be to move setuptools to a development dependency; I've pushed an update to try that.

I can see that if all you do is run netmiko from a poetry environment then this doesn't really make much difference.  But if you're installing it from a wheel in some other way, then setuptools isn't intrinsically needed (aside from uses of `pkg_resources` that are deprecated nowadays, it should only be needed for the sdist → wheel step), and it's cleaner to avoid the install-time requirement if possible.  We've been trying to start cleaning this sort of thing up in Debian packaging recently, although there's certainly a lot to do since historically a lot of people just slapped in `install_requires=["setuptools"]` without really thinking about it.

ktbyers:
Yes, I can get rid of pylama, but I am pretty adverse to needing to fix my CI-CD because Python decided to remove setuptools (and PY3.9 - PY3.11 would still have it installed internally i.e. it is not a rush to fix this in Netmiko).

ktbyers:
Also, I invariably run into issues whenever `setuptools` isn't there (so I basically now `pip install setuptools` as a first step after making a virtual environment).

But your needs and my needs are a bit at odds here...i.e. I get a lot of pain at all the breakage this causes and issues it creates (even when it is not directly Netmiko, but other dependencies).

I will think about the dev dependency option as that might be a good middle ground.

PR #3521, opened 2024-11-01
Fix to support cisco apic paging

yone2ks:
I updated the CiscoApicSSH class to disable paging after the connection is established. 
Paging is enabled by default on Cisco APIC, but because CiscoApicSSH inherits from LinuxSSH, paging was not disabled after connection. 

This update addresses the issue by calling disable_paging from CiscoSSHConnection.


ktbyers:
Thanks @yone2ks

PR #3531, opened 2024-11-08
Nokia srl prompt stripping

ktbyers:


ktbyers:
@jeffkala Added some tests and converted the one method to a private method (as more consistent with standard Netmiko behavior).

netmiko/nokia/nokia_srl.py
```diff
@@ -53,6 +53,32 @@ def session_preparation(self) -> None:
...
 
+    def strip_prompt(self, *args: Any, **kwargs: Any) -> str:
+        """Strip the line1 of multiline prompt from the output."""
+        a_string = super().strip_prompt(*args, **kwargs)
+        return self._strip_context_items(a_string)
+
+    def _strip_context_items(self, a_string: str) -> str:
+        """Strip NokiaSRL-specific output.
+
+        Nokia will put extra context is line 1 of the prompt, such as:
```
hellt, line 64:
```suggestion
        Nokia will put extra context in the 1st line of the prompt, such as:
```

netmiko/nokia/nokia_srl.py
```diff
@@ -53,6 +53,32 @@ def session_preparation(self) -> None:
             self.disable_paging(command=command, cmd_verify=True, pattern=r"#")
         self.set_base_prompt()
 
+    def strip_prompt(self, *args: Any, **kwargs: Any) -> str:
+        """Strip the line1 of multiline prompt from the output."""
```
hellt, line 57: is this statement correct about the 1st line? Shouldn't the strip tackle the two lines?

jeffkala: Still calling super which I'd assume would strip the original prompt.

ktbyers: I updated the docstrings, the `super()` call will get rid of `A:srl1#`.

The `_strip_context_items` call will get rid of things like `--{ running }...`

ktbyers:
Look good @jeffkala @hellt ?

jeffkala (approved):

netmiko/nokia/nokia_srl.py
hellt: ah, I see, by "prompt" you mean only one line of the SR Linux prompt.
Ok, lgtm

PR #3535, opened 2024-11-13
Add Palo Alto Log Collector Compatibility

opachano:
- Updates self.write_channel and _test_channel_read in paloalto_panos.py for Palo Alto Log Collector compatibility.

Context: Palo Alto Devices set up as log collectors have no valid "show admins' command. So this module will always fail for these devices. "show system info" is available for firewall and log collector setups.


opachano:
@ktbyers Please let me know if you need me to provide more information/context around this fix or if you have any comments for this PR, I would be more that happy to implement any suggestions. I have been using a patched version of this module with these changes included and it has been working without issues.

Issue #3401, opened 2024-03-12
Improvement for Huawei SmartAX


evilmonkey19:
### Description of Issue/Question
At least, in most of the current huawei smartax versions there is the possibility of using the command `mmi-mode enable` for doing machine-to-machine communications (what netmiko does). I suggest to enable the possibility to use this command to improve the performance and the robustness of the Huawei smartax driver. I will dig and come up with factual data of how does it improves the current implementation.

As well, the robustness is something which the current implementation lacks of in certain cases. For instance, if there is a command for which you don't have enough permissions, it gives an error. To further clarify, if you try to use for example `display ont info 0/0/0` and you don't have the enough permissions to use this command, the current implementation doesn't work.

However, there is a downside (as always) to use this command. It can only be used when `config` can be used by the user in the OLT. This means that it will not be a one-ring fits everything mode. I will explore how much it improves when enabled.

### Setup
Python 3.11

### Netmiko version
Netmiko 4.3.0

### Netmiko device_type (if relevant to the issue)
`huawei_smartax`

evilmonkey19:
May I ask for suggestion on how to do the testing as well? I think it is a good time as well to add some good testing in this case 😄

evilmonkey19:
I've just started doing some improvements (really small ones) and those are some starting improvements

![imagen](https://github.com/ktbyers/netmiko/assets/56732262/e735e55e-aa6e-419b-b5f9-7bf69fc77871)

ktbyers:
I am going to close this issue as we are working on this (using other issues for tracking).

PR #3551, opened 2024-12-24
Huawei smartax improve closing

evilmonkey19:
Improve closing of Huawei SmartAX so the terminal disconnect correctly. Related to this issue #3401 although there is no improvement in speed. I don't think we will that much, but we can do it if needed.


ktbyers:
@evilmonkey19 FYI, linting failure.

netmiko/huawei/huawei_smartax.py
```diff
@@ -131,3 +131,19 @@ def save_config(
...
+        # If after 30 seconds the device hasn't logged out, force it
+        while time.time() - start_time < 30:
+            output += self.read_channel()
+            if output.endswith("Are you sure to log out? (y/n)[n]:"):
+                self.write_channel("y" + self.RETURN)
+                output = ''
+                continue
+            if output.endswith("Configuration console exit, please retry to log on\n"):
+                break
+            time.sleep(0.01)
```
ktbyers, line 149: I would probably break this into `new_output` and `output` and check for patterns in `new_output`. Using `endswith` sounds a bit dangerous to me as any extra text/newline will cause this to not work.

Technically since we don't return `output` or do anything with it, we could just capture `new_output` and not even keep a running `output`.

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3555

netmiko/huawei/huawei_smartax.py
ktbyers: Nevermind...let's just leave it.

PR #3555, opened 2025-01-06
Huawei smartax improve closing

ktbyers:


ktbyers:
@evilmonkey19 Can you provide details on when you would see each of these messages?

```
            if output.endswith("Are you sure to log out? (y/n)[n]:"):

            if output.endswith("Configuration console exit, please retry to log on\n"):
```

Also how do the above messages relate to?

```
 mmi-mode enable
```

Is this message ubiquitous on logout:

```
Configuration console exit, please retry to log on
```

Obviously, if this message doesn't show up, then we are going to wait 30s before we logout (which is a very large delay)--so I am worried/concerned about these changes.

evilmonkey19:
Hi @ktbyers!

I have check it out right now. And this is what I see:

- In the `mmi-mode enable` the exit output is as follows:
```
OLT#quit 

  Configuration console exit, please retry to log on
```
Here I don't see the message `Are you sure to log out? (y/n)[n]:`. Currently, when connecting using Netmiko it is not using this mode. This mode for instance doesn't print the output while typing, only after the command has been issued.

- In the `scroll` + `undo smart` (the actual way to enter using Netmiko) is as follows the exit command:
```
OLT#quit 
  Check whether system data has been changed. Please save data before logout. Are you sure to log out? (y/n)[n]:y

  Configuration console exit, please retry to log on
Connection to 1.1.1.1 closed.
```

In this mode the message `Are you sure to log out? (y/n)[n]:` appears.

Personally, I feel like the `mmi-mode enable` is better, but you need to enter to the config level. I would like to use it, but I discourage using it as it is in the config level and not enable. What do you think @ktbyers ?

ktbyers:
@evilmonkey19 How different is the behavior of the Huawei smartax` when using `mmi-enable`?

At a high-level, what changes would need to be made to support this `mmi-enable`? Also do all Huawei SmartAX have this mode?

evilmonkey19:
Basically, you need to enter up to the config level (which takes roughly right now 10 seconds). All Huawei SmartAX as long as I know they have this mode. Basically the `mmi-mode` is a combination of 4 commands in one: `interactive`, `scroll`, `smart` and `infoswitch cli`.

ktbyers:
@evilmonkey19 

We could make a private method `_enable_mmi_mode()` and `_disable_mmi_mode()` which an automation user can call to enable/disable mmi_mode in their Netmiko code.

If we do this, then we would need to know what other things in the Netmiko code need to change:
1. _disable_smart_interaction() 

https://github.com/ktbyers/netmiko/blob/develop/netmiko/huawei/huawei_smartax.py#L44

2. Slow cleanup() call -- likely due to not gracefully closing the SSH connection


Anything else we would need to do?

evilmonkey19:
I will give it a try developing the mmi_mode and the disable mmi_mode.

evilmonkey19:
I also add the fix from the issue #3559 to this PR

evilmonkey19:
I've opened the pull-request #3561  because i couldn't write on the branch

evilmonkey19:
Done in #3685

ktbyers:
I think this PR is now obsolete so closing (cc @evilmonkey19 ).

evilmonkey19:
I can confirm that this PR is totally obsolete by #3695, #3688, #3686 and  #3687

PR #3556, opened 2025-01-06
genexis_solt33_telnet support

ktbyers:


ktbyers:
@bsdemon Broke out `genexis solt33 telnet` to be a separate PR--to make reviewing the PR easier.

PR #3561, opened 2025-01-25
adding the mmi-mode, improving the config modes and also adding graceful disconnection

evilmonkey19:


evilmonkey19:
Hi @ktbyers. What would we need to add in order to merge this branch into develop?

ktbyers:
@evilmonkey19 This is now obsolete--correct?

I am going to close this as I think we implemented this in a different PR, but just let me know if we still need this for some reason.

PR #3562, opened 2025-01-25
Huawei smartax ssh autodetect

evilmonkey19:
Adding support for Huawei SmartAX. The `\n` is important because by default this device has a smart mode completion. This way the final output would be like this:
```
OLT>display system sys-info 
{ <cr>||<K> }: 

  Command:
          display system sys-info
  --------------------------------------------------
  The main service identification of this node:
    1

  The IP address of this node:
    1.1.1.1

  The description of this node:
    Huawei Integrated Access Software
  --------------------------------------------------
```


ktbyers:
I separated only the SSH autodetect changes in this PR:

https://github.com/ktbyers/netmiko/pull/3574

PR #3568, opened 2025-02-05
H3C switches are too slow in executing commands

LeisureZhao:
Compared to Cisco switches, H3C switches are too slow in executing
commands because the value of the pattern is not set in the
check_config_mode function. This function is called twice in both
the config_mode and exit_config_mode, each time using read_channel_timing
instead of read_until_pattern, which greatly extends the duration.
This commit set default pattern of hp_comware to save time when
checking the config mode.

### DIFF
--- a/netmiko/hp/hp_comware.py
+++ b/netmiko/hp/hp_comware.py
@@ -37,10 +37,10 @@ class HPComwareBase(CiscoSSHConnection):
         return super().exit_config_mode(exit_config=exit_config, pattern=pattern)
 
     def check_config_mode(
-        self, check_string: str = "]", pattern: str = "", force_regex: bool = False
+        self, check_string: str = "]", pattern: str = r"[>]|]", force_regex: bool = False
     ) -> bool:
         """Check whether device is in configuration mode. Return a boolean."""
-        return super().check_config_mode(check_string=check_string)
+        return super().check_config_mode(check_string=check_string, pattern=pattern)
 
     def send_config_set(
         self,


ktbyers:
@LeisureZhao What would be the value of `explicit_pattern` here?

LeisureZhao:
> @LeisureZhao What would be the value of `explicit_pattern` here?

I tried to set explicit_pattern to r"[>]|]", which means that when a pattern with > or ] appears, the read will stop. The configuration example is as follows:
```
switch_config = {
'device_type': 'hp_comware',
'host': xxx,
'username': xxx,
'password': xxx,
'explicit_pattern': r"[>]|]"
}
```
Considering the need to be compatible with past usage habits, the code logic remains unchanged when the explicit_pattern is not set. If it is deemed that the execution time of the command is too long, you can set the explicit_pattern configuration item for hp_comware. The pattern can be customized according to the actual switch.

ktbyers:
@LeisureZhao I think we should implement this as a defined pattern on `check_config_mode` and not as an argument that we pass in via `__init__.py`.

Does that sound reasonable.

Can you restructure the PR to implement that?

LeisureZhao:
> @LeisureZhao I think we should implement this as a defined pattern on `check_config_mode` and not as an argument that we pass in via `__init__.py`.
> 
> Does that sound reasonable.
> 
> Can you restructure the PR to implement that?

Completed modifications, please review.

netmiko/hp/hp_comware.py
```diff
@@ -37,10 +37,10 @@ def exit_config_mode(self, exit_config: str = "return", pattern: str = r">") ->
         return super().exit_config_mode(exit_config=exit_config, pattern=pattern)
 
     def check_config_mode(
-        self, check_string: str = "]", pattern: str = "", force_regex: bool = False
+        self, check_string: str = "]", pattern: str = r"[>]|]", force_regex: bool = False
```
ktbyers, line 40: @LeisureZhao Why the `|` (pipe character) here?

ktbyers:
Superseded by https://github.com/ktbyers/netmiko/pull/3573

PR #3575, opened 2025-02-07
Aruba CX SSH autodetect

ktbyers:


ktbyers:
FYI, @robinreinhardt I extracted your SSH autodetect code and made some slight changes). The Aruba AOS-CX driver was created in a subsequent PR (so the other parts of your original PR had been superseded).

PR #3579, opened 2025-02-08
Dell SCP Sonic

ktbyers:


ktbyers:
@aravindmani-1 FYI

ktbyers:
@aravindmani-1 This should be the same as what you created. I clarified the naming of `_enter_cli()` and `_return_shell()` (as the previous names were confusing).

Additionally, I only defined these in the DellSonicSSH class and eliminated the duplicate definition.

Let me know if you see any issues.

I am going to go ahead and merge this and we can fix any issues.

PR #3583, opened 2025-02-19
Optilink GOLT Driver

ktbyers:


netmiko/optilink/op_golt944.py
```diff
@@ -0,0 +1,13 @@
...
+
+class OptilinkGOLT944Telnet(HuaweiTelnet):
+    """Optilink GOLT 944 telnet driver"""
+
+    def session_preparation(self) -> None:
+        self.ansi_escape_codes = True
+        self._test_channel_read(pattern=r"[>\]]")
+        self.set_base_prompt()
+        self.config_mode()
+        self.disable_paging(command="screen-rows per-page 0")
```
ktbyers, line 12: @bsdemon Are you sure that is the right command to disable output paging? This is what you would typically use on a Ciena SAOS and not on a Huawei based device?

ktbyers:
@bsdemon Was the entering/exiting of `enable` mode necessary for the GOLT944 device? It didn't really look like it to me, but I was basing this on Huawei behaviors.

netmiko/optilink/op_golt944.py
bsdemon: Yes, it is.

bsdemon:
@ktbyers I don't have access to these devices anymore, but I'm certain there were issues exiting from enable mode, and that's why I had overridden `exit_enable_mode`.

ktbyers:
@bsdemon Okay, I ripped out the GOLT 944 driver since it didn't sound like we had any way to validate it. I left the GOLT 924 driver (from reading the history of this, it sounded like it was done).

PR #3589, opened 2025-02-23
Adding new platform: Huawei ONT

evilmonkey19:
This is driver has been tested against Huawei EchoLife HG8310M
https://support.huawei.com/enterprise/en/optical-access/echolife-hg8310m-pid-9028636

This platform is peculiar as it doesn't have enable or config and neither save mode (all is safe immediately).

sample output using the SU mode:
```

Welcome Visiting Huawei Home Gateway
Copyright by Huawei Technologies Co., Ltd.

Login:root
Password:
WAP>
WAP>
WAP>
WAP>su
success!
SU_WAP>
SU_WAP>
SU_WAP>quit
success!
WAP>
WAP>logout

```

### DIFF
--- a/netmiko/huawei/huawei_ont.py
+++ b/netmiko/huawei/huawei_ont.py
@@ -1,0 +1,96 @@
+import re
+
+from netmiko.no_enable import NoEnable
+from netmiko.no_config import NoConfig
+from netmiko.cisco_base_connection import CiscoBaseConnection
+from netmiko.exceptions import NetmikoAuthenticationException
+
+
+class HuaweiONTBase(NoEnable, NoConfig, CiscoBaseConnection):
+    prompt_pattern = r"WAP>"
+    su_prompt_pattern = r"SU_WAP>"
+
+    def session_preparation(self) -> None:
+        """Prepare the session after the connection has been established."""
+        self.ansi_escape_codes = True
+        # The _test_channel_read happens in special_login_handler()
+        self.set_base_prompt()
+
+    def save_config(
+        self, cmd: str = "save", confirm: bool = True, confirm_response: str = "y"
+    ) -> str:
+        """In Huawei ONTs, there is no save command. All changes are immediate."""
+        raise NotImplementedError("Save config is not supported on Huawei ONTs.")
+
+    def check_su_mode(self) -> bool:
+        """Check if the device is in su mode."""
+        current_prompt = self.find_prompt()
+        return bool(re.search(self.su_prompt_pattern, current_prompt))
+
+    def su_mode(
+        self, cmd: str = "su", pattern: str = "", re_flags: int = re.IGNORECASE
+    ) -> str:
+        """Attempt to become super user."""
+        output = ""
+        if not self.check_su_mode():
+            self.write_channel(self.normalize_cmd(cmd))
+            self.read_until_pattern(pattern="success!", re_flags=re.IGNORECASE)
+            output += self.read_until_pattern(pattern=pattern, re_flags=re_flags)
+        return output
+
+    def exit_su_mode(
+        self, cmd: str = "quit", pattern: str = "", re_flags: int = re.IGNORECASE
+    ) -> str:
+        """Exit super user mode."""
+        output = ""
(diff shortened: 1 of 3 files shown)


netmiko/huawei/huawei_ont.py
```diff
@@ -0,0 +1,96 @@
...
+
+from netmiko.no_enable import NoEnable
+from netmiko.no_config import NoConfig
+from netmiko.cisco_base_connection import CiscoBaseConnection
+from netmiko.exceptions import NetmikoAuthenticationException
+
+
+class HuaweiONTBase(NoEnable, NoConfig, CiscoBaseConnection):
+    prompt_pattern = r"WAP>"
+    su_prompt_pattern = r"SU_WAP>"
```
ktbyers, line 11: Are these WAP and SU_WAP patterns hard-coded (i.e. I assume you can't change this `WAP` name)?

netmiko/huawei/huawei_ont.py
```diff
@@ -0,0 +1,96 @@
...
+    def save_config(
+        self, cmd: str = "save", confirm: bool = True, confirm_response: str = "y"
+    ) -> str:
+        """In Huawei ONTs, there is no save command. All changes are immediate."""
+        raise NotImplementedError("Save config is not supported on Huawei ONTs.")
+
+    def check_su_mode(self) -> bool:
+        """Check if the device is in su mode."""
+        current_prompt = self.find_prompt()
+        return bool(re.search(self.su_prompt_pattern, current_prompt))
```
ktbyers, line 28: The `su_mode` methods should probably all be converted over to `enable` mode methods (i.e. we should just use `check_enable_mode`, `enable_mode`, and `exit_enable_mode` methods here).

Do you see any issues with this?

netmiko/huawei/huawei_ont.py
```diff
@@ -0,0 +1,96 @@
...
+        pwd_pattern: str = r"Password:|password:",
+        delay_factor: float = 1.0,
+        max_loops: int = 20,
+    ) -> str:
+        """Telnet login for Huawei Devices"""
+        output = ""
+        return_msg = ""
+        try:
+            # Search for username pattern / send username
+            output = self.read_until_pattern(pattern=username_pattern, re_flags=re.I)
```
ktbyers, line 74: Does the existing telnet_login method work (in the parent class)? It would be good if we could re-use the parent method. If not, we probably should add a looping construct as part of this to increase the reliability of the login process.

ktbyers (changes requested):

netmiko/huawei/huawei_ont.py
evilmonkey19: Yeees, I have seen them hardcoded. I can tell for all ONTs from Huawei because each one of them has different commands. Nonetheless, so far I haven't seen any that does not have this pattern hardcorded

netmiko/huawei/huawei_ont.py
evilmonkey19: Looks perfect to me! Perhaps it might be cleaner :)

evilmonkey19:
What is it left to push intro develop?

Noppes:
Seems fine, if I look at https://forum.huawei.com/enterprise/intl/en/thread/blog/667246086517047296?blogId=667246086517047296 the WAP> part indeed seems hardcoded.

I do have a question though, if you use the shell command do you go into some sort of config mode? Because if I look at https://forum.huawei.com/enterprise/intl/en/thread/blog/667245482952507393?blogId=667245482952507393 there does seem to be some config commands? So far Im unable to find how to enter this config mode though.

evilmonkey19:
Hi! Thanks for your comments. To answer you question, the last link is not the ONT but rather the OLT CLI. That is why if you see the commands closely you can poll other ONTs. Actually the ONTs inside, at least the Huawei ones, are a Linux Dopa internally. That's the "config" mode.

ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/3800

PR #3595, opened 2025-02-26
Adva driver fixes

ktbyers:
Linting plus setting `terminator` argument.


ktbyers:
@Gatorjosh14 Fixed the linting.

Note, I set the `terminator` parameter in send_config_set() to `terminator: str = r"-->",` (as None is not allowed here--due to it being a string). I thought this was the correct setting for it (on quick glance at the code).

Issue #3590, opened 2025-02-23
ipinfusion_ocnos module does not implement commit method


pklimai:
The ipinfusion_ocnos module does not implement commit() method, however, OcNOS, at least its modern versions, require commits to apply configuration. Shall I create a pull request? Or anyone?

ktbyers:
@pklimai Yes, can you create a pull-request add add the commit method. You can reference the Juniper or IOS-XR driver if you need examples.

pklimai:
@ktbyers thanks, I'm working on this. It looks like I need to implement several extra methods, for operations like "abort transaction" and "confirm-commit". Is it OK to have them implemented for vendor-specific connection class, even if they are not present in base classes?

pklimai:
@ktbyers Thanks, this is done now (and also added _cancel_commit method).

pklimai:
Please LMK if you need anything else to proceed merging it.

PR #3596, opened 2025-02-27
Add support for IPInfusion OcNOS commit model

pklimai:
As discussed in issue #3590, the ipinfusion_ocnos module does not currently implement commit() method, however, modern versions of OcNOS require commits to apply configuration.

This PR implements the required commit() method. In addition, confirm_commit() method for confirm-commit operation and abort_transaction() method for abort-transaction operation are implemented. The send_config_set() method is slightly modified to raise exceptions in case config load errors are detected.

I tested this code with OcNOS 6.3, 6.4 and 6.5 versions.

### DIFF
--- a/netmiko/ipinfusion/ipinfusion_ocnos.py
+++ b/netmiko/ipinfusion/ipinfusion_ocnos.py
@@ -1,5 +1,5 @@
 import time
-from typing import Any, Optional, Union, Sequence, Iterator, TextIO
+from typing import Any, Optional
 from socket import socket
 
 from netmiko._telnetlib.telnetlib import (
@@ -34,7 +34,7 @@ class IpInfusionOcNOSBase(CiscoBaseConnection):
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()
 
-    def send_config_set(self, *args, **kwargs) -> str:
+    def send_config_set(self, *args: Any, **kwargs: Any) -> str:
         """Send config command(s). Requires separate calling of commit to apply."""
 
         # Do not exit config mode - it won't allow until you commit or abort transaction
@@ -42,11 +42,13 @@ class IpInfusionOcNOSBase(CiscoBaseConnection):
         error_markers = [
             "% Invalid input detected at '^' marker.",
             "% Parameter not configured",
-            "% Ambiguous command:"
+            "% Ambiguous command:",
         ]
         for error_marker in error_markers:
             if error_marker in output:
-                raise ValueError(f"Send config set failed with the following errors:\n\n{output}")
+                raise ValueError(
+                    f"Send config set failed with the following errors:\n\n{output}"
+                )
         return output
 
     def commit(
@@ -54,7 +56,7 @@ class IpInfusionOcNOSBase(CiscoBaseConnection):
         confirm: bool = False,
         confirm_delay: Optional[int] = None,
         description: str = "",
-        read_timeout: float = 120.0
+        read_timeout: float = 120.0,
     ) -> str:
         """
         Commit the candidate configuration.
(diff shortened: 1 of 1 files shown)


pklimai:
Please LMK if you need anything else in regards to this PR

pklimai:
Hi @ktbyers, can you please approve the workflow?

pklimai:
Hi @ktbyers can you please approve? Do you need anything else to merge this PR?

ktbyers:
I will try to review this in the next week.

netmiko/ipinfusion/ipinfusion_ocnos.py
```diff
@@ -28,15 +28,156 @@ def session_preparation(self) -> None:
...
+        self.send_command("terminal no monitor")
         # Clear the read buffer
         time.sleep(0.3 * self.global_delay_factor)
         self.clear_buffer()
 
+    def send_config_set(self, *args: Any, **kwargs: Any) -> str:
+        """Send config command(s). Requires separate calling of commit to apply."""
+
+        # Do not exit config mode - it won't allow until you commit or abort transaction
+        output = super().send_config_set(*args, **kwargs, exit_config_mode=False)
```
ktbyers: We probably should do the following here:

```
# Default 'exit_config_mode' to False unless it is explicitly overwritten
exit_config_mode = kwargs.get("exit_config_mode", False)
```

ktbyers: On error_markers, this probably should be outside of Netmiko directly and should just be called with existing `error_pattern` argument. Additionally, I am reluctant to change the default behavior for a specific driver (i.e. to default to having the error patterns specified and active in the driver).

So I would recommend we remove the error_markers and you can implement this in your Netmiko calling code as something like:

```python
        error_markers = [
            "% Invalid input detected at '^' marker.",
            "% Parameter not configured",
            "% Ambiguous command:",
        ]
       err_pattern1, err_pattern2, err_pattern3 = [re.escape(marker) for marker in error_markers]
       error_pattern(rf"(?:{err_pattern1}|{error_pattern2}|{err_pattern3})")
```

Then make your send_config_set() and pass in the `error_pattern` argument.

Assuming we do this, then `ipinfusion_ocnos` send_config_set would become just the `exit_config_mode` change and the `super()` call.

netmiko/ipinfusion/ipinfusion_ocnos.py
```diff
@@ -28,15 +28,156 @@ def session_preparation(self) -> None:
...
+        description: str = "",
+        read_timeout: float = 120.0,
+    ) -> str:
+        """
+        Commit the candidate configuration.
+
+        default (no options):
+            command_string = commit
+        confirm and confirm_delay:
+            command_string = commit confirmed <confirm_delay>
```
ktbyers: We should remove `commit confirm` from the `commit` method. These should be implemented as a private method as per the communication in https://github.com/ktbyers/netmiko/issues/3590

ktbyers: Same with `confirm_delay` this argument should be moved to a private `_commit_confirm()` method.

ktbyers: Or worded slightly differently, Netmiko's `commit()` method only commits previous changes--it does not support `commit confirm`, this would need to be done entirely in private methods.

pklimai:
Thanks @ktbyers ! I made modifications as per your suggestions, please check.

ktbyers:
@pklimai My bad on the `commit`...I forgot that we added this `confirm` and `confirm_delay` arguments into the Juniper driver. I will revert your changes back to what you had earlier on this.

I actually remembered this while semi-sleeping last night :-)

pklimai:
@ktbyers Thanks for clarification, indeed I implemented commit() similar to how it was done for Junos. Now I reverted and just fixed send_config_set(). Please check.

ktbyers:
@pklimai Thanks for contributing this.

Note, I am going to change `description` argument to be `comment` as that is the argument I used with most of the other drivers that use commit. Let me know if you see any issues with this.

pklimai:
Thank you very much @ktbyers. I am fine with using `comment`.

PR #3602, opened 2025-03-05
Aviat WTM Driver (updated)

ktbyers:


ktbyers:
This PR should be good except we need to add a `commit()` method.

ktbyers:
@gt732 Do you know what `commit` looks like from a CLI perspective on these devices--can you post an example here?

PR #3614, opened 2025-03-13
Telnet support for cisco_nxos devices

Noppes:
Commit adds telnet support for cisco_nxos devices


ktbyers:
@Noppes Did you run the `show` and `config` tests against a device using `telnet`? I don't have an easy way to regression test this?

PR #3618, opened 2025-03-17
edgecore_sonic added

Noppes:
Added edgecore_sonic, a linux based device


ktbyers:
I think we should make the key be `edgecore_sonic` and not `edgecore_linux` (as I am somewhat anticipating we might ultimately need something `sonic` specific here.

ktbyers (approved):

PR #3623, opened 2025-03-21
NEC IX: Make exit_config_mode behavior more standard

ktbyers:


ktbyers:
FYI @inaba-vdom-0

ktbyers:
@inaba-vdom-0 Let me know if you see other issues with this (besides the more limited show permissions that you mentioned in the original PR).

inaba-vdom-0:
@ktbyers 
I believe there are two possible behaviors when exiting ```config_mode()``` using the "exit" command.  

## Behavior when inside an interface or similar hierarchy  
```
router# svintr-config
Enter configuration commands, one per line. End with CNTL/Z.
router(config)#
router(config)# interface GigaEthernet2.0
router(config-GigaEthernet2.0)#
router(config-GigaEthernet2.0)# exit
router(config)#
```

## Behavior when at the top menu  
```
router# svintr-config
Enter configuration commands, one per line. End with CNTL/Z.
router(config)#
router(config)# ntp ip enable
router(config)# exit
router#
```

To ensure consistency in the implementation of ```exit_config_mode()```, I believe the configure command is necessary.  
```
router(config)# interface GigaEthernet2.0
router(config-GigaEthernet2.0)# configure?
  configure  -- Go back to top menu
router(config-GigaEthernet2.0)# exit?
  exit  -- Exit from interface configuration mode
router(config-GigaEthernet2.0)# exit
router(config)# configure?
  configure  -- Go back to top menu
router(config)# exit?
  exit  -- Exit from configure mode
router#
```

Additionally, in the ix series, session disconnections do not reset the configuration hierarchy, meaning that the hierarchy may vary when entering config mode.  
The configure command in config mode serves as a "Go back to top menu" function, and I believe it should be implemented in the ```config_mode()``` command to restore the hierarchy to its original state.  
```
router# svintr-config
Enter configuration commands, one per line. End with CNTL/Z.
router(config-GigaEthernet2.0)#
router(config-GigaEthernet2.0)# configure
router(config)#
```

ktbyers:
@inaba-vdom-0 So shouldn't `exit_enable_mode` also behave the way you described above? In other words, I saw you implemented a double `exit` call there--shouldn't that use this `continue` + `exit` pattern?

inaba-vdom-0:
@ktbyers 
currently reconsidering this implementation, please give us some time.
If a user wants to use the show command in ```send_command```, they should use enable/config mode, but is this left up to the user's script implementation?

ktbyers:
Yes, they should call `enable()` before executing the `show_command` (since this platform has decided a lot of `show` commands need to be executed from this `svintr-config` mode.

Basically, this vendor has made some not great implementation choices on how their device behaves and I think the important thing from a Netmiko perspective is to maintain Netmiko behavior consistency (and that will require the code writer using NEC IX will have to do some extra steps).

inaba-vdom-0:
@ktbyers The implementation of `exit_config_mode` looks good. However, `session_preparation` and `set_base_prompt` still need some modifications, so I’m planning to create a separate pull request for those.

ktbyers:
@inaba-vdom-0 Okay, I made updates here. Basically, I made config_mode(), check_config_mode(), and exit_config_mode() identical to the corresponding `enable` calls. In other words, both `enable` and `config` calls do the exact same thing.

For both `enable()` and `config_mode()`, I have the device execute `svintr-config` and then execute `configure`. For exit_enable_mode() and exit_config_mode(), I have the device execute `configure` (go to the top-level of config) and then execute `exit` (completely exit `svintr-config`).

Let me know if you see any issues with the above behaviors.

I changed session_preparation to just make the `config_mode` and `exit_config_mode` calls (as there was no longer a purpose in making both `enable()` and `config_mode()` calls.

Is there some other issue with `set_base_prompt()`?

ktbyers:
@inaba-vdom-0 I am going to go ahead and merge this as I think it is the right starting point (from what you have stated above). If there are other issues, we can fix them in a separate PR.

inaba-vdom-0:
@ktbyers Thank you. There were no issues with the changes. I had a similar idea regarding the modifications to `session_preparation()`.  
As for `set_base_prompt()`, I initially implemented it similarly to `cisco_ios`, but since the prompt is not shortened, that change turned out to be unnecessary.  
The `base_prompt[:16]` part should be removed, and it should simply return `super().set_base_prompt()`.

Issue #2106, opened 2021-01-10
file_transfer for Linux doesn't honor environment variables for Linux prompts


dcsaki:
When attempting to transfer a file to a linux host,
where host has prompt has pattern:  PS1=\u@\h:\w>
I am getting error from base_connection.py send_command:
E           OSError: Search pattern never detected in send_command: [\$#]

I have already set environment variable: NETMIKO_LINUX_PROMPT_ALT = ">"  to prevent similar problem detecting login prompt.
I suspect this problem is due to hard-coding of strings  r"[\$#]"  in scp_handler.py since when I change it to "[\$#>]" (in 3 places) my file transfer succeeded.

ktbyers:
@dcsaki Yes, do you want to submit a PR on this?

dcsaki:
@ktbyers I could attempt to submit a PR, but am a rookie with github & not sure of the process.

ktbyers:
@dcsaki No worries...

ktbyers:
Fixed in referenced commit.

PR #3630, opened 2025-03-28
Linux SCP updates to support environment variables for prompts

ktbyers:


ktbyers:
Fixes https://github.com/ktbyers/netmiko/issues/2106

PR #3643, opened 2025-04-07
cumulus_vx added

Noppes:
Added cumulus_vx driver for cumulus vx devices


netmiko/ssh_autodetect.py
```diff
@@ -146,6 +146,12 @@
         "priority": 99,
         "dispatch": "_autodetect_std",
     },
+    "cumulus_vx": {
+        "cmd": "uname -a",
+        "search_patterns": [r"cumulus"],
+        "priority": 99,
+        "dispatch": "_autodetect_std",
+    },
```
ktbyers, line 154: Does Cumulus VX have any differentiation on `uname -a` from a real Physical Cumulus switch?

Noppes: Im not a 100% sure, this is what `uname -a` gives:
`Linux cumulus 5.10.0-cl-1-amd64 #1 SMP Debian 5.10.162-1+cl5.4.0u1 (2023-01-20) x86_64 GNU/Linux`
I can change it to check `Linux cumulus` instead and lower its priority

ktbyers: @Noppes Okay, we need something that differentiates `Cumulus VX` from `Cumulus` that is not VX. Or if these two things are identical then we should just make a `Cumulus` driver and name it `cumulus_linux`.

Basically I don't want to misidentify non-VX devices via autodetect.

ktbyers: Thinking about this a bit more...probably should just name the device_type as `cumulus_linux` (unless we have good reasons to think `VX` needs to be different).

Noppes: Ive renamed it to cumulus_linux

ktbyers (approved):

Issue #3261, opened 2023-07-31
Intermittent ReadTimeout exceptions with cmd_verify=True


josephwhite13:
Hello, I'm in an environment where TACACS can take ~7-15 seconds to verify the command being sent to a device. Because of this, send_command and send_command_timing intermittently raise ReadTimeout exceptions when called with `cmd_verify=True`

I was able to resolve my issue with doing `new_data = self.command_echo_read(cmd=cmd, read_timeout=read_timeout)`. I don't mind putting in a PR if this seems like a reasonable solution.

Here's an excerpt of my stack-trace:
```
  File "/virtualenv/lib64/python3.11/site-packages/netmiko/base_connection.py", line 1740, in send_command
    new_data = self.command_echo_read(cmd=cmd, read_timeout=10)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/virtualenv/lib64/python3.11/site-packages/netmiko/base_connection.py", line 1452, in command_echo_read
    new_data = self.read_until_pattern(
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/virtualenv/lib64/python3.11/site-packages/netmiko/base_connection.py", line 721, in read_until_pattern
    raise ReadTimeout(msg)
netmiko.exceptions.ReadTimeout: 
Pattern not detected: '<< Command I sent to the device >>' in output.

Things you might try to fix this:
1. Adjust the regex pattern to better identify the terminating string. Note, in
many situations the pattern is automatically based on the network device's prompt.
2. Increase the read_timeout to a larger value.

You can also look at the Netmiko session_log or debug log for more information.
```

ktbyers:
@josephwhite13 Yeah, we could probably do that for the `send_command` method (if you want to submit a PR on it).

Reference this issue in the PR so I remember what it is associated with.

PR #3653, opened 2025-04-13
Add the cmd_verify_timeout parameter to adjust the wait time for echo

MutoShota:
The `cmd_verify` parameter in both `send_command` and `send_command_timing` was previously fixed at 10 seconds.

This update introduces an argument to make this timeout adjustable.This change accommodates devices that take longer than 10 seconds to return an echo, addressing the issue described in #3261

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -1522,6 +1522,7 @@ A paramiko SSHException occurred during connection creation:
         ttp_template: Optional[str] = None,
         use_genie: bool = False,
         cmd_verify: bool = False,
+        cmd_verify_timeout: float = 10.0,
         raise_parsing_error: bool = False,
     ) -> Union[str, List[Any], Dict[str, Any]]:
         """Execute command_string on the SSH channel using a delay-based mechanism. Generally
@@ -1561,6 +1562,8 @@ A paramiko SSHException occurred during connection creation:
 
         :param cmd_verify: Verify command echo before proceeding (default: False).
 
+        :param cmd_verify_timeout: Maximum time to wait looking for echo. (Default: 10.0).
+
         :param raise_parsing_error: Raise exception when parsing output to structured data fails.
         """
         if delay_factor is not None or max_loops is not None:
@@ -1574,7 +1577,7 @@ A paramiko SSHException occurred during connection creation:
 
         cmd = command_string.strip()
         if cmd and cmd_verify:
-            new_data = self.command_echo_read(cmd=cmd, read_timeout=10)
+            new_data = self.command_echo_read(cmd=cmd, read_timeout=cmd_verify_timeout)
         output += new_data
         output += self.read_channel_timing(
             last_read=last_read, read_timeout=read_timeout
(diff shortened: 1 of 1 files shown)


ktbyers:
Why does it take so long for your device to echo the command?

MutoShota:
> Why does it take so long for your device to echo the command?

When connecting directly to the equipment via Telnet or SSH, there are almost no issues. However, there are situations where operations are performed via the serial console server, and it takes about 20 seconds for the echo to return. This adjustment is necessary to address that issue. If there are other solutions, please let me know.

ktbyers:
@MutoShota I am not really understanding why command echo takes 20s...that is obviously an unusable CLI.

Does it work if you just disable cmd_verify globally?

Can you send me an example command/situation where the command echo is taking that long?

MutoShota:
@ktbyers 

I apologize for not providing a concrete example earlier. This issue does not occur with all commands but under specific conditions and commands. Here is a simplified method to reproduce the issue.

Please create 60,000 interfaces using sub-interfaces and logical tunnels on a Juniper vMX. In that environment, execute `show interfaces terse | foo.log`. You should observe that the echo after the `|` is delayed. Although it takes less than 10 seconds on vMX, in the physical device we are working with, it takes more than 20 seconds for the string after the `|` to return.

We considered disabling cmd_verify, but handling the prompt becomes cumbersome. Therefore, we are motivated to extend the timeout value for cmd_verify.

ktbyers:
@MutoShota Okay, I think the proper solution for this situation is just to disable cmd_verify for the one command.

In other words,

```python
net_connect.send_command("show interfaces terse | save foo.log", cmd_verify=False)
```

I am not sure I understood this part of your response, "but handling the prompt becomes cumbersome"?


Regards,

Kirk

MutoShota:
@ktbyers 
We thought we could solve the issue by disabling cmd_verify. 

As a result, after executing the command, several line breaks occurred, and then a ReadTimeout error was triggered. The search pattern during the ReadTimeout was looking for the string following `save` or the pattern `Wrote ~ lines ~` output by the `save` command, rather than the prompt. The cumbersome prompt handling I mentioned refers to this situation.

Looking at the implementation of `send_command`, we couldn't identify the reason why line breaks are inserted after [`write_channel`](https://github.com/ktbyers/netmiko/blob/996a6d3600c824565c9e10210be2a189021159df/netmiko/base_connection.py#L1786) and why the prompt couldn't be found.

To address this, we are using netmiko with an increased read_timeout passed to command_echo_read. It would be helpful if this read_timeout could be adjusted via an argument.

ktbyers:
I don't think I want to add another argument for Netmiko's init method (as this sounds like a very rare, edge case).

I am happy to look into your other issue, though, if you provide more details regarding what is happening when you try to use `send_command` with cmd_verify=False.

The full exception stack trace and debugging output would be helpful.

MutoShota:
Hi @ktbyers

I apologize for the delayed response. 

After giving it some thought, I believe that instead of adding a new argument, setting cmd_verify=False and specifying expect_string is the more appropriate approach.
I would like to provide logs, but unfortunately, sharing real logs is difficult as they would require masking. With that in mind, I’ll close this PR.

Thank you for suggesting solutions.

PR #3660, opened 2025-04-15
Asterfusion driver updates

ktbyers:


ktbyers:
FYI, @Yukissss and @Noppes

ktbyers:
@Yukissss I renamed the class name to be more consistent with Netmiko naming conventions.

I also didn't see any purpose to having `send_config_set`, `send_config_from_file`, and `send_command` in the Asterfusion module (i.e. it looked like we should just be able to use the parent method in each of these cases).

I also made `cli_mode` a private attribute i.e. `_cli_mode` as it is not a part of the standard Netmiko __init__ arguments.

Finally, I renamed your `write_config` method as `save_config` to be more consistent with Netmiko standard defitions.

Let me know if you see any issues here.

ktbyers:
@Yukissss Your test change states:

```
    if net_connect.device_type == "asterfusion_asternos":
        assert pytest.skip("Disable Paging not applied by default on this platform")
```

But then you have a `disable_paging` method defined, can you expand on this? Does the `terminal raw ouptut` not work for disabling paging?

Yukissss:
@ktbyers Thank you for your prompt response and suggested modifications. Regarding the paging issue, I've investigated further and found that it was likely caused by our internal development issue where the terminal raw-output setting couldn't properly disable paging for the show version command.

This particular issue was actually resolved through a ticket several months ago. Therefore, we can now remove the additional handling in this test case as it's no longer necessary.

Let me know if you'd like me to submit these changes or if you have any other suggestions.

ktbyers:
@Yukissss Thanks for your contribution.

PR #3670, opened 2025-05-02
ADD: Support for LANCOM LCOS SX 5

Nivispluma:
### Title
ADD: Support for LANCOM LCOS SX 5

### Description
This PR introduces a new driver for LANCOM devices running LCOS SX 5, which is the OS for the Enterprise models.

### Peculiarity
LANCOM does not allow `Exec` commands to be run in `Config` Mode, therefore a `config-mode-check` is added to `save_config` and `cleanup`.
SSH Connection to LCOS SX 5 devices normally start in the `EXEC` Mode. This Mode however offers only inconsistent command usage. Therefore the driver will elevate the Session to `privileged EXEC` Mode by default during `session_preparation`
The `terminal width` commands dont work on LCOS SX 5

### Testing
Tested on:
- XS-6128QF

A File with full `show` and `config` test result is attached.

[2025-05-02_netmiko-lancom-driver-LCOSSX5-test.txt](https://github.com/user-attachments/files/20014014/2025-05-02_netmiko-lancom-driver-LCOSSX5-test.txt)

### DIFF
--- a/netmiko/lancom/lancom_lcossx5.py
+++ b/netmiko/lancom/lancom_lcossx5.py
@@ -1,0 +1,97 @@
+from netmiko.cisco_base_connection import CiscoSSHConnection
+
+
+class LancomLCOSSX5SSH(CiscoSSHConnection):
+    def session_preparation(self) -> None:
+        """
+        Prepare the session after the connection has been established.
+
+        The connection will enter `Privileged EXEC` by default, as the `EXEC` mode
+        offers inconsistent command options
+        """
+        self._test_channel_read()
+        super().send_command_timing(
+            "enable",
+            strip_prompt=False,
+            strip_command=False,
+        )
+        super().set_base_prompt()
+        super().disable_paging()
+        self.clear_buffer()
+
+    def set_terminal_width(self, *args, **kwargs) -> str:
+        """
+        LCOS SX 5 does not support 'terminal width', therefore skip it.
+        """
+        return ""
+
+    def check_config_mode(
+        self,
+        check_string: str = "(Config)#",
+        pattern: str = "#",
+        force_regex: bool = False,
+    ) -> bool:
+        """
+        Checks if the device is in configuration mode or not.
+
+        :param check_string: Identification of configuration mode from the device
+        :type check_string: str
+
+        :param pattern: Pattern to terminate reading of channel
+        :type pattern: str
+
+        :param force_regex: Use regular expression pattern to find check_string in output
+        :type force_regex: bool
+
+        :return: True if in configuration mode, False if not
+        """
+        return super().check_config_mode(
+            check_string=check_string, pattern=pattern, force_regex=force_regex
+        )
+
+    def exit_enable_mode(self, exit_command: str = "end") -> str:
+        """Exits enable (privileged exec) mode."""
+        return super().exit_enable_mode(exit_command=exit_command)
+
+    def cleanup(self, command: str = "logout") -> None:
+        """
+        Cleanup / Gracefully exit the SSH session
+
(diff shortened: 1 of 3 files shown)


netmiko/lancom/lancom_lcossx5.py
```diff
@@ -0,0 +1,98 @@
...
+        super().send_command_timing(
+            "enable",
+            strip_prompt=False,
+            strip_command=False,
+        )
+        super().set_base_prompt()
+        super().disable_paging()
+        self.clear_buffer()
+
+    def set_terminal_width(self, *args: Any, **kwargs: Any) -> str:
```
Noppes, line 23: There is no need to override the set_terminal_width. It's only ever called from session_preparation which you already override. If you do really want to keep it, you should have it `raise NotImplementedError`

Noppes:
Looking at https://www.lancom-systems.de/fileadmin/download/documentation/CLI-Reference/MA_LCOS-SX-5.20-CLI-Reference_EN.pdf, it says to enter config mode its `configure` and not `configure terminal` like in cisco, is that correct?

netmiko/lancom/lancom_lcossx5.py
```diff
@@ -0,0 +1,98 @@
...
+class LancomLCOSSX5SSH(CiscoSSHConnection):
+    def session_preparation(self) -> None:
+        """
+        Prepare the session after the connection has been established.
+
+        The connection will enter `Privileged EXEC` by default, as the `EXEC` mode
+        offers inconsistent command options
+        """
+        self._test_channel_read()
+        super().send_command_timing(
```
Noppes, line 14: any reason you are calling send_command for this instead of just using self.enable()?

Nivispluma:
Sorry for not answering... had a lot going on :-/

> Looking at https://www.lancom-systems.de/fileadmin/download/documentation/CLI-Reference/MA_LCOS-SX-5.20-CLI-Reference_EN.pdf, it says to enter config mode its `configure` and not `configure terminal` like in cisco, is that correct?
> 
> I also see that you check for `(Config)#` with a capital C is that correct?

The (official) Documentation is not explicit enough here. While `configure` would be enough to enter config mode it just auto completes to `configure terminal`. Therefore i chose to just leave the `configure terminal`. 

The capital `C` is correct (as you can see below). I think that's because LANCOM is a German vendor
```
(XS-6128QF)>
(XS-6128QF)>enable

(XS-6128QF)#con?

configure

(XS-6128QF)#configure ter?

terminal

(XS-6128QF)#configure ?

<cr>                     Press enter to execute the command.
terminal                 Enter into Global Config Mode.

(XS-6128QF)#configure

(XS-6128QF)(Config)#
```

netmiko/lancom/lancom_lcossx5.py
Nivispluma: i have removed the redundant function

netmiko/lancom/lancom_lcossx5.py
Nivispluma: i tested multiple versions with self.enable(), and i wasn't able to get it working. I think it has to do with LANCOM not prompting for enable passwords at all.
Therefore i chose to use this workaround.

Nivispluma: *EDIT: i had the following Error:
```
ValueError: Failed to enter enable mode. Please ensure you pass the 'secret' argument to ConnectHandler.
```
However i was able to resolve this by passing the `enable_pattern` which results in `self.enable(enable_pattern=r"#")`.
Now its working as you suggested :)

ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/3806

PR #3683, opened 2025-05-27
add matched error message to ConfigInvalidException message

### LINKED ISSUE #3682: ConfigInvalidException should include the error produced by the invalid config
Currently, when you try to send config to a switch with, for example:

```python
<netmiko BaseConnection instance>.send_config_set(
    config.splitlines(),
    error_pattern=r"%.*",
)
```
if one of those lines of config is invalid and produces output on the switch like `% Invalid input` or `Vlan ### is not configured`, it would be really nice if netmiko were to raise an error like `ConfigInvalidException("Invalid input detected at command: {cmd}, error message: {matched_error}")`

alextremblay:
Fixes #3682

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -2474,9 +2474,12 @@ You can also look at the Netmiko session_log or debug log for more information.
 
     def paramiko_cleanup(self) -> None:
         """Cleanup Paramiko to try to gracefully handle SSH session ending."""
+        if self.remote_conn is not None:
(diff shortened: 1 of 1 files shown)


ktbyers:
I like this idea--the main other thing I see that the PR needs is some way that it is tested.

ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/3811

To add tests.

Issue #3681, opened 2025-05-27
Huawei Issue Container


ktbyers:
High-level container for various issues on Huawei

ktbyers:
cc: @evilmonkey19

ktbyers:
I assume we want to start with the Smart AX and mmi-mode?

I think this probably should be its own separate driver. We have a few options here:
1. Use device_type "huawei_smartax" (i.e. replace the current "huawei_smartax" driver with this and create some sort of `huawei_smartaxlegacy` driver
2. Completely eliminate the old "huawei_smartax" driver and replace it with the new `mmi-mode` code. Note, we probably shouldn't do this unless we are confident nothing will need the old driver.
3. Create a new driver `huawei_smartaxmmi` and leave the old one as-is (i.e. `huawei_smartax`).

How common is `mmi-mode` in the `smartax` platform?

evilmonkey19:
I would start improving the current `huawei_smartax` and later introduce the `huawei_smartaxmmi` driver. Before, I thought about adding the mmi-mode as an option when introducing the credentials as follows:

```python
from netmiko import ConnectHandler

credentials = {
  "hostname": "localhost",
  "username": "admin",
  "password": "admin",
  "port": 22,
  "mmi-mode": True,  # by default would be False
}

with ConnectHandler(**credentials) as conn:
  pass
```

Anyway, this is the planification i have in mind for improving the Huawei SmartAX template:
- [ ] Turn off the alarms on the cli as it breaks the driver. There is a command: `infoswitch cli OFF`. This command applies only to the current session.
- [ ] Improving the closing time. Do a graceful closing as right now it does it directly.
- [ ] Improve accessing the `enable` and `config` mode which takes some time now.
- [ ] Enable the `mmi-mode`. I want to open the discussion if it is better the way i propose or implement a separate driver. Next comment.

ktbyers:
So in these first three items--is there anything in those that depend on `mmi-mode`?

```
Turn off the alarms on the cli as it breaks the driver. There is a command: infoswitch cli OFF. This command 
applies only to the current session.

Improving the closing time. Do a graceful closing as right now it does it directly.

Improve accessing the enable and config mode which takes some time now.
```

[...]

PR #3684, opened 2025-05-29
Huawei SmartAX driver: disable alarms

evilmonkey19:
Related to #3681

*This PR is the base for upcoming changes.*

It fixes the case when there are alarms shown in the CLI which might happen. When an alarm is shown it breaks the driver.

The following has been done:
- A new command is being sent when establishing the command: `infoswitch cli OFF`. This command must be sent using the enable mode, but after being sent it returns to the base mode.
- A simple ENUM with the states as certain command needs certain states. See `infoswitch cli OFF`.
- A __send_command function only for the huawei_smartax module has been implemented. It is similar to the _disable_smart_interaction (which has been removed) and it add the modes.
- Small changes on a 2 logs on line 80 and 81.

### DIFF
--- a/netmiko/huawei/huawei_smartax.py
+++ b/netmiko/huawei/huawei_smartax.py
@@ -45,9 +45,7 @@ class HuaweiSmartAXSSH(CiscoBaseConnection):
     def __send_command(
         self, command: str, delay_factor: float = 1.0, mode: Mode = Mode.BASE
     ) -> None:
-        """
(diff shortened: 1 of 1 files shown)


evilmonkey19:
I agree that maybe it isn't the best implementation as it breaks the standard way. I didn't think about do it the same way you suggest. Actually it is better also because we are not removing code. I move this PR to #3686 :)

PR #3686, opened 2025-05-29
Huawei SmartAX infoswitch off fix

ktbyers:
SmartAx defaults to sending logging/debugging to the terminal.

Disable this behavior during session_prepration method.

### DIFF
--- a/netmiko/huawei/huawei_smartax.py
+++ b/netmiko/huawei/huawei_smartax.py
@@ -23,6 +23,7 @@ class HuaweiSmartAXSSH(CiscoBaseConnection):
 
         self.set_base_prompt()
         self._disable_smart_interaction()
+        self._disable_infoswitch_cli()
         self.disable_paging()
 
     def strip_ansi_escape_codes(self, string_buffer: str) -> str:
@@ -40,6 +41,18 @@ class HuaweiSmartAXSSH(CiscoBaseConnection):
         log.debug(f"new_output = {output}")
         log.debug(f"repr = {repr(output)}")
         return super().strip_ansi_escape_codes(output)
+
+    def _disable_infoswitch_cli(self, command: str = "infoswitch cli OFF") -> None:
+        """SmartAX sends debugging output to the terminal by default--disable this."""
+        priv_escalation = False
+        # infoswitch cli OFF requires enable mode
+        if self.check_enable_mode():
+            self.enable()
+            priv_escalation = True
+        pattern = re.escape(self.base_prompt)
+        self._send_command_str(command, expect_string=rf"{pattern}#")
+        if priv_escalation:
+            self.exit_enable_mode()
 
     def _disable_smart_interaction(
         self, command: str = "undo smart", delay_factor: float = 1.0


evilmonkey19:
Hi!

I've tried and the implementation almost work. I did a small change to make it work. It's in #3687.

magnus-kr:
i think the initial claim of sendings logs to the terminal by default isn't technically correct

although infoswitch cli is on by default on every session, the device has to been previously configured to output to terminal

ktbyers:
@magnus-kr Okay, is there a change we should make here to the pull-request (or at least to the comment in the method/function)?

magnus-kr:
> @magnus-kr Okay, is there a change we should make here to the pull-request (or at least to the comment in the method/function)?

from what i could gather, the global config is set by "alarm output" and "undo alarm output" (and default is no output, you have to explicitly turn on)
while infoswitch is only opting out for the current session

documentation doesn't cover this in depth, but after checking a few hundred devices from different lines (MA56XX, MA58XX), models and firmwares the findings above were accurate

so calling  _disable_infoswitch_cli() when preparing the session isn't necessary for most use cases because global default values cover that

ktbyers:
@magnus-kr But it would ensure it was always disabled for the current session--correct?

magnus-kr:
@ktbyers yes, it unsubscribes for the current session even if the device does not output alarms to terminal (which is the default global value)

magnus-kr:
In our scenario, we mostly use config mode, this is what our logs look now:
```
with ConnectHandler(
                **self.device,
                session_log=file,
            ) as ssh:
                ssh.is_alive()
                ssh.find_prompt()
                ssh.enable()
                ssh.config_mode()
                ssh.send_command("undo interactive")


HUAWEI-5800>

HUAWEI-5800>

HUAWEI-5800>enable 

HUAWEI-5800#

HUAWEI-5800#

HUAWEI-5800#infoswitch cli OFF 

HUAWEI-5800#

HUAWEI-5800#disable 

HUAWEI-5800>

HUAWEI-5800>scroll 

HUAWEI-5800>

HUAWEI-5800>

HUAWEI-5800>

HUAWEI-5800>enable 

HUAWEI-5800#

HUAWEI-5800#

HUAWEI-5800#

HUAWEI-5800#config 

HUAWEI-5800(config)#

HUAWEI-5800(config)#

HUAWEI-5800(config)#undo interactive 
  Interactive function is disabled

HUAWEI-5800(config)#
```

we are entering and leaving enable mode to run infoswitch cli OFF, which does take longer for the total execution

evilmonkey19:
You are right, but the issue is that as you said it might be enabled in certain devices. If we want a reliable driver then we might account for this case. In the worst case, it will do nothing. Generally speaking, I don't think it might take more than a few hundred milliseconds to send the command. Most of the time usually is spent on the connection and disconnection and not sending the command itself.

Before, it was true this was a slow driver because of the regex pattern matching (see #3688). I you really need speed during connection I would recommend to use the mmi-mode, but it takes a bit more of time to establish the full setup as it needs to enter the config mode. This would be the case for mass provisioning of ONTs.

If you really need speed on the initial setup perhaps you could modify the netmiko driver in a local instance and comment the line.
This would be the case to show all ONTs in a single chassis with commands like "display ont info summary 0" (btw, if this is your case just do use_textfsm and you'll see all parsed ;) ).

magnus-kr:
@evilmonkey19 makes perfect sense! thanks for taking the time to go over it!

i will definitely run some tests with mmi-mode, since we don't really have any routines run outside config mode

evilmonkey19:
@magnus-kr You welcome! Let me know how it goes :)

Issue #3704, opened 2025-07-10
Enable() fails on switch with particular name


Martynas677:
Hi.

This issue persisted for me for quite some time, but only now I managed to find the time to troubleshoot it.

I sometimes run a simple Netmiko script that collects some information from 150+ switches, but my script fails to enter enable mode on this single particular switch. This switch is no different than the others, it runs same cisco IOS version, same hardware model as others, the only difference from other switches it has, is it's hostname "AS"

So from debugging logs I can see that the Netmiko doesn't enter my provided secret password for this switch. If I change the switch name to something else, somehow it starts working :)

My thoughts are that the hostname is somehow interfering with netmiko while it search "password" pattern while trying to enter enable mode.

I have several of examples

Trying to enter enable with hostname AS doesn't work. The script is not entering the secret for enable.

```

DEBUG:paramiko.transport:starting thread (client mode): 0xfda33230
DEBUG:paramiko.transport:Local version/idstring: SSH-2.0-paramiko_3.5.1
DEBUG:paramiko.transport:Remote version/idstring: SSH-2.0-Cisco-1.25
INFO:paramiko.transport:Connected (version 2.0, client Cisco-1.25)
DEBUG:paramiko.transport:=== Key exchange possibilities ===
DEBUG:paramiko.transport:kex algos: diffie-hellman-group-exchange-sha1, diffie-hellman-group14-sha1
DEBUG:paramiko.transport:server key: ssh-rsa
DEBUG:paramiko.transport:client encrypt: aes128-ctr, aes192-ctr, aes256-ctr
DEBUG:paramiko.transport:server encrypt: aes128-ctr, aes192-ctr, aes256-ctr
DEBUG:paramiko.transport:client mac: hmac-sha2-256, hmac-sha2-512, hmac-sha1, hmac-sha1-96
DEBUG:paramiko.transport:server mac: hmac-sha2-256, hmac-sha2-512, hmac-sha1, hmac-sha1-96
DEBUG:paramiko.transport:client compress: none
DEBUG:paramiko.transport:server compress: none
DEBUG:paramiko.transport:client lang: <none>
DEBUG:paramiko.transport:server lang: <none>
DEBUG:paramiko.transport:kex follows: False
DEBUG:paramiko.transport:=== Key exchange agreements ===
DEBUG:paramiko.transport:Kex: diffie-hellman-group-exchange-sha1
DEBUG:paramiko.transport:HostKey: ssh-rsa
[...]

PR #3707, opened 2025-07-19
Prevent Cisco password false positives, fix issue #3704

AlexMunoz905:
Fixes issue #3704.

When the Cisco devices hostname has "AS", "PAS", or "assw", the pattern search for "password" causes it to error out, and claim that a secret is not provided.

The pattern has been updated to ignore that, and prevent the false positives for "password".

### DIFF
--- a/netmiko/__init__.py
+++ b/netmiko/__init__.py
@@ -8,7 +8,7 @@ MIN_PYTHON_VER = "3.9"
 
 # Make sure user is using a valid Python version (for Netmiko)
 def check_python_version():  # type: ignore
-    python_snake = "\U0001f40d"
+    python_snake = "\U0001F40D"
 
     # Use old-school .format() method in case someone tries to use Netmiko with very old Python
     msg = """
--- a/netmiko/__init__.py
+++ b/netmiko/__init__.py
@@ -8,7 +8,7 @@ MIN_PYTHON_VER = "3.9"
 
 # Make sure user is using a valid Python version (for Netmiko)
 def check_python_version():  # type: ignore
-    python_snake = "\U0001F40D"
+    python_snake = "\U0001f40d"
 
     # Use old-school .format() method in case someone tries to use Netmiko with very old Python
     msg = """
--- a/netmiko/cisco_base_connection.py
+++ b/netmiko/cisco_base_connection.py
@@ -18,7 +18,7 @@ class CiscoBaseConnection(BaseConnection):
     def enable(
         self,
         cmd: str = "enable",
-        pattern: str = "ssword",
+        pattern: str = r"(?:^|\n)[Pp]assword:",
         enable_pattern: Optional[str] = None,
         check_state: bool = True,
         re_flags: int = re.IGNORECASE,
--- a/netmiko/utilities.py
+++ b/netmiko/utilities.py
@@ -137,7 +137,7 @@ def load_devices(file_name: Union[str, bytes, "PathLike[Any]", None] = None) -> 
 
 
 def find_cfg_file(
-    file_name: Union[str, bytes, "PathLike[Any]", None] = None,
+    file_name: Union[str, bytes, "PathLike[Any]", None] = None
 ) -> Union[str, bytes, "PathLike[Any]"]:
     """
     Search for netmiko_tools inventory file in the following order:
@@ -191,7 +191,7 @@ def display_inventory(my_devices: Dict[str, Union[List[str], Dict[str, Any]]]) -
 
 
 def obtain_all_devices(
-    my_devices: Dict[str, Union[List[str], Dict[str, Any]]],
+    my_devices: Dict[str, Union[List[str], Dict[str, Any]]]
 ) -> Dict[str, Dict[str, Any]]:
     """Dynamically create 'all' group."""
     new_devices = {}
--- a/netmiko/utilities.py
+++ b/netmiko/utilities.py
@@ -137,7 +137,7 @@ def load_devices(file_name: Union[str, bytes, "PathLike[Any]", None] = None) -> 
 
 
 def find_cfg_file(
-    file_name: Union[str, bytes, "PathLike[Any]", None] = None
+    file_name: Union[str, bytes, "PathLike[Any]", None] = None,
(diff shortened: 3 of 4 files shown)


AlexMunoz905:
These pictures are tests with hostname "AS" and hostname "SW-1" on a Cisco Catalyst 3850 running IOS-XE 16.12.07

Before Change:
image


After Change:
image

ktbyers:
@AlexMunoz905 On quick glance, I don't think this fixes the issue referenced. 

The failure is because the hostname is a subset of the `password` string and Netmiko's logic is basically saying search for "ssword" string OR the trailing prompt.

So Netmiko starts reading "Password" and it gets to "as" which then matches the devices prompt and it stops reading.

It then goes to the next line and checks was this "ssword" and sees it wasn't (i.e. "as" doesn't match "ssword") so it doesn't actually send the "enable" secret.

Note, your test above fails because you didn't define the "secret" with ConnectHandler (i.e. it is failing for a reason unrelated to what you are testing).

ktbyers:
This part of my above statement is wrong...

```
Note, your test above fails because you didn't define the "secret" with ConnectHandler (i.e. it is failing for 
a reason unrelated to what you are testing).
```

ktbyers:
FYI, my testing of your fix confirmed that it would still fail when the prompt as a substring of the specified 'password' pattern.

ktbyers:
@AlexMunoz905 It is not clear to me why your testing worked?

But here is what I saw when I implemented your fix (note, I implemented your pattern directly in base_connection.py, but this should have the same effect when called by anything using 'cisco_base_connection.py').

This is with the host prompt set to "AS>" on a Cisco IOS-XE.

```
ipdb> p output
'enable'
ipdb> n
> /home/kbyers/VENV/py3_venv/lib/python3.11/site-packages/netmiko/base_connection.py(2052)enable()
   2051             output += self.read_until_prompt_or_pattern(
-> 2052                 pattern=pattern, re_flags=re_flags
   2053             )

ipdb> n
> /home/kbyers/VENV/py3_venv/lib/python3.11/site-packages/netmiko/base_connection.py(2051)enable()
   2050             # Search for trailing prompt or password pattern
-> 2051             output += self.read_until_prompt_or_pattern(
   2052                 pattern=pattern, re_flags=re_flags

ipdb> n
> /home/kbyers/VENV/py3_venv/lib/python3.11/site-packages/netmiko/base_connection.py(2056)enable()
   2055             # Send the "secret" in response to password pattern
-> 2056             if re.search(pattern, output):
   2057                 self.write_channel(self.normalize_cmd(self.secret))
```

Note here you can see it only read up to "Pas" (which is what I would expect from the code). Basically it kept reading until it matched the "AS" hostname (ignoring capitalization).

At the point, Netmiko will Not send the enable secret and the `enable()` call will fail (basically because 'pattern' is not in the output, only the subset of pattern up until hostname is present.

```
ipdb> p output
'enable\nPas'
ipdb> p patter
*** NameError: name 'patter' is not defined
ipdb> p pattern
'ssword'
ipdb> c
  File "/home/kbyers/EP/test1.py", line 12, in <module>
    nc.enable()
  File "/home/kbyers/VENV/py3_venv/lib/python3.11/site-packages/netmiko/cisco_base_connection.py", line 27, in enable
    return super().enable(
           ^^^^^^^^^^^^^^^
  File "/home/kbyers/VENV/py3_venv/lib/python3.11/site-packages/netmiko/base_connection.py", line 2065, in enable
    raise ValueError(msg)
ValueError: Failed to enter enable mode. Please ensure you pass the 'secret' argument to ConnectHandler.
```

ktbyers:
Here is where I implemented/tested your fix:

```python
2009     def enable(
2010         self,
2011         cmd: str = "",
2012         # pattern: str = "ssword",
2013         pattern: str = r"(?:^|\n)[Pp]assword:",
2014         enable_pattern: Optional[str] = None,
2015         check_state: bool = True,
2016         re_flags: int = re.IGNORECASE,
2017     ) -> str:
```

ktbyers:
FYI, I implemented an alternate fix here:

https://github.com/ktbyers/netmiko/pull/3708

ktbyers:
FYI, I realized I incorrectly tested your solution by implementing it in base_connection.py...so I am re-testing it.

AlexMunoz905:
Interesting that it failed on your end, but I'm glad you were able to get a fix implemented.

Thanks!

ktbyers:
I tested your solution more and it looks like it works, but it relies on the position of the character match to beat out the prompt. 

I was scratching my head a bit on why it worked (so I dug into it more)...this is mostly notes for myself, but you might also be interested.

So in a simplified representation (and ignoring case):

```
pattern = r"(as|password:"
```

And

```
data = "password:"
```

Then pattern of "password:" would win as it matches on the "p" in password (which is an earlier character than "as").


So that made me wonder why it wins if prompt is "PA" (once again ignoring case).

```
pattern = r"(PA|password:)
```

Since normally "PA" would win here (i.e. both the first character the first pattern in the logical-or wins).

```python
In [1]: pattern = r"(PA|Password:)"
   ...: 

In [2]: import re

In [3]: re.search(pattern, "Password:", flags=re.I)
Out[3]: <re.Match object; span=(0, 2), match='Pa'>
```

But that is where you are counting on leading newlines in order to match earlier...

```python
ipdb> output
'\nPassword: '
ipdb> re.search(pattern, output, flags=re_flags)
<re.Match object; span=(0, 10), match='\nPassword:'>
ipdb> pattern
'(?:PA|(?:^|\\n)[Pp]assword:)'
ipdb> re_flags
re.IGNORECASE
ipdb> repr(output)
"'\\nPassword: '"
```

I think the other solution in PR #3708 is probably better...as I don't really want to count on the leading newlines being there.

Looking into this, did lead me to another bug in the `enable()` method which is here--this re.search doesn't use the re.flags so it won't ignore case.

```python
2055             # Send the "secret" in response to password pattern
2056             if re.search(pattern, output):
2057                 self.write_channel(self.normalize_cmd(self.secret))
2058                 output += self.read_until_prompt()
```

PR #3712, opened 2025-07-22
Add support to connect to HPE 1920 and 1950 OfficeConnect switches

spamik:
HPE 1920 and 1950 switches has standard Comware5 / Comware7 CLI but it needs to be enabled with special command sent to the CLI. This patch add new hpe_officeconnect_model argument to HPComwareBase class. When this argument is populated with hpe1920 or hpe1950 string value, it send magic command immediately after establishing connection with switch and therefore these switches can be handled as standard Comware switches.

### DIFF
--- a/netmiko/hp/__init__.py
+++ b/netmiko/hp/__init__.py
@@ -1,4 +1,7 @@
 from netmiko.hp.hp_procurve import HPProcurveSSH, HPProcurveTelnet
 from netmiko.hp.hp_comware import HPComwareSSH, HPComwareTelnet
+from netmiko.hp.hp_comware1920 import HPComware1920SSH, HPComware1920Telnet
+from netmiko.hp.hp_comware1950 import HPComware1950SSH, HPComware1950Telnet
 
-__all__ = ["HPProcurveSSH", "HPProcurveTelnet", "HPComwareSSH", "HPComwareTelnet"]
+__all__ = ["HPProcurveSSH", "HPProcurveTelnet", "HPComwareSSH", "HPComwareTelnet", "HPComware1920SSH",
+           "HPComware1920Telnet", "HPComware1950SSH", "HPComware1950Telnet"]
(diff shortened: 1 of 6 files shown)


evilmonkey19:
I can confirm that this is a [special case](https://networkguy.de/enabling-advanced-cli-on-hp-v1910-switches/) those switches :+1:

evilmonkey19:
In this case instead of adding some argument to the credentials i would suggest creating a new device type or issue the command version on startup and based on that decide. If you go for the second option I can help you creating the template in [NTC-Templates](https://github.com/networktocode/ntc-templates)

spamik:
@evilmonkey19 issuing display version command was my first attempt to keep it simple without configuration. But problem is that even display version is not possible without enabling CLI :-/ In basic version there were some commands from which output is possible to determine what type switch is. But I've ended on that problem that these commands has long output which not fit on one screen (----more---- placer is returned to CLI). Command which disable paging is also available only after enabling full CLI. And even pipe (something like "| include 1920") is also available only with full CLI. So I didn't find the way how to detect switch model before enabling full CLI with something which not lead to terminal paging issue :-/ (If you have some idea, I can test it, i have bunch of both types of these switches).

evilmonkey19:
Then the best solution is to create a separated device_type for these type of switches and there is no way to detect them beforehand.

spamik:
OK, I've reverted previous commit and I've added 2 new device types - one for 1920 series, second for 1950 series.

evilmonkey19:
Awesome! IMHO, It does look much cleaner not the codebase but also the usability for the user. 😃

Soullego:
@spamik i would suggest also add 1910 model.

spamik:
@Soullego Do you have some info what steps are needed on 1910 to enable full cli? I unfortunately didn't have any 1910 on stock to try it.

Soullego:
@spamik yeah, same as 1920 but different password
```
_cmdline-mode on
y
512900
```

spamik:
I've added new class for HPE 1910, so if someone can test it...

ktbyers:
I am going to close this one as I don't really know what to do about the hard-coded backdoor passwords (and everytime I look at it, I basically defer/punt it).

PR #3720, opened 2025-08-20
Fix typing hints for timeouts: change from int to float

jonglezb:
Checking both Paramiko and Telnetlib, all those parameters are actually float.

Only self.timeout seems to be used as an int in Netmiko itself, so it's left unchanged.


jonglezb:
The CI fails when type-checking keepalive: `netmiko/base_connection.py:1212: error: Argument 1 to "set_keepalive" of "Transport" has incompatible type "float"; expected "int"  [arg-type]`

I believe this is a mistake in typeshed, because keepalive can definitely be a float (it is compared to `time.time()`)

I just opened https://github.com/python/typeshed/pull/14596 to fix this in typeshed.

jonglezb:
With the new release 4.0.0.20250821 of types-paramiko, everything is green!

PR #3725, opened 2025-09-01
Nokia isam

Noppes:
Driver for Nokia isam devices


ktbyers:
@Noppes will try to look at this in the next few days. Feel free to ping me, if this slips through the cracks.

netmiko/nokia/nokia_isam.py
```diff
@@ -0,0 +1,64 @@
...
+        if match:
+            # strip off >... from base_prompt; strip off leading *
+            self.base_prompt: str = match.group(1)
+
+        return self.base_prompt
+
+    def cleanup(self, command: str = "logout") -> None:
+        """Gracefully exit the SSH session."""
+        # Always try to send final 'logout'.
+        self.write_channel(command + self.RETURN)
```
ktbyers, line 43: This doesn't seem sufficient?

Probably need more like:

```python
    def cleanup(self, command: str = "logout") -> None:
        """Gracefully exit the SSH session."""
        try:
            if self.check_config_mode():
                self.exit_config_mode()
        except Exception:
            pass
        # Always try to send final command
        if self.session_log:
            self.session_log.fin = True
        self.write_channel(command + self.RETURN)
```

netmiko/nokia/nokia_isam.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def check_config_mode(
+        self,
+        check_string: str = "configure#",
+        pattern: str = "",
+        force_regex: bool = False,
+    ) -> bool:
+        """Use equivalent enable method."""
+        return super().check_config_mode(check_string=check_string)
```
ktbyers: Is this right? It seems inconsistent with `set_base_prompt` and what I am seeing on the CLI behavior (but really hard for me to find documentation on this platform so maybe I am wrong here).

Noppes: default promp is name># and when you go into config mode it becomes name>configure#

ktbyers: And what happens when to the prompt when you traverse down the configuration tree?

I am having a lot of trouble finding documentation on this platform, but it looks to me like it does something like:

```
name>configure#

And

configure>system>security#
```

Note, I definitely could be wrong given the difficult obtaining documentation on this platform.

ktbyers: @Noppes Any update on my question here?

Noppes: From the logging the customer send, I only saw `name>configure#`, but its an easy change to just check for >configure instead, see new commit

PR #3734, opened 2025-09-30
Add Perle SSH class

bewing:
NB:  This class is for connecting to / configuring Perle SCS/SCR devices, not for connecting to reverse telnet/SSH devices managed by them.

`send_command_timing` is the only supported command method, as there is a paginator that cannot be disabled.


bewing:
Apologies, refactored per mypy

ktbyers:
Superseded by:

https://github.com/ktbyers/netmiko/pull/3738

PR #3741, opened 2025-10-23
fix(mikrotik): fix check_file_exists mikrotik output handling

arthurguyon:
Mikrotik `/file print detail` command output added flags in latest firmware versions, this causes problems during scp file transfer


ktbyers:
Superseded by https://github.com/ktbyers/netmiko/pull/3813

PR #3745, opened 2025-10-29
ansi_escape_codes can be configured when creating a connection object

rqx110:


ktbyers:
@rqx110 I am not seeing the purpose of this? Why care about needing this argument at connect? As opposed to setting it post connection?

rqx110:
Actually, I didn't have a specific purpose. It's just that when using this library, I wanted to remove ANSI code, and I found it wasn't easy to find directly in the parameters. I only found this property after reviewing the source code. Therefore, if this property is in the initialization parameter list, it will appear in the generated API documentation.

PR #3754, opened 2025-12-12
nokia_isam improvements

Noppes:
It looks like `environment mode batch` was made to disable paging and other interactive elements on a nokia_isam. Plus an exit to go out of the >environment 'node'

```
device>#environment mode ?
batch                 ! non-interactive mode intended for scripting tools
interactive           ! interactive mode intended for human operators

device>#environment mode batch ?
[no] prompt           : the specification of the prompt
                        default = ""
[no] terminal-timeout : specify the inactivity timeout of the terminal
                        default = timeout:30
[no] print            : specifies the way the output is printed on the terminal
                        default = default
[no] screen-length    : specifies the number of lines to be printed on the
                        terminal for paging
                        default = 0

[no] inhibit-alarms   : disables the alarm reporting in the current  session
```

For the exit config, `exit all`, exits out of all 'nodes'. Which is better incase they are deep into a bunch 'nodes'.

```
device>#exit ?
[no] all              : set the current position to the top node
[no] one              : set the current position to the parent node ```

### DIFF
--- a/netmiko/nokia/nokia_isam.py
+++ b/netmiko/nokia/nokia_isam.py
@@ -12,8 +12,8 @@ class NokiaIsamSSH(BaseConnection, NoEnable):
         self._test_channel_read()
         self.set_base_prompt()
         commands = [
-            "environment inhibit-alarms",
-            "environment screen-length 0",
+            "environment mode batch",
+            "exit",
         ]
         for command in commands:
             self.disable_paging(command=command, cmd_verify=True, pattern=r"#")
@@ -60,7 +60,7 @@ class NokiaIsamSSH(BaseConnection, NoEnable):
             config_command=config_command, pattern=pattern, re_flags=re_flags
         )
 
-    def exit_config_mode(self, exit_config: str = "exit", pattern: str = "") -> str:
+    def exit_config_mode(self, exit_config: str = "exit all", pattern: str = "") -> str:
         return super().exit_config_mode(exit_config=exit_config)
 
     def save_config(

### CODE REVIEW

**Comment by ktbyers** [comment]
@Noppes FYI, I am fixing the GitHub actions runner issue(s)

**Comment by Noppes** [comment]
> @Noppes FYI, I am fixing the GitHub actions runner issue(s)

Yea I figured it was something like that. Best wishes for the new year!

**Comment by ktbyers** [comment]
Comment from earlier ISAM issue:

```
In my testing of netmiko with Nokia ISAM products, I was able to get some limited functionality to work 
when the device_type = autodetect. Currently there is no vendor support for it with netmiko.

The Nokia 7360 OLT, actually runs two different operating systems. The main os is the isam.

The 2nd operating, which does the switching, is similar to SROS.

For these Nokia product lines you probably want use the following disable paging command

"environment print no-more inhibit-alarms mode batch"

There are multiple types of ISAM products, and they all operate a little bit differently etc.
```

Reference:

https://github.com/ktbyers/netmiko/issues/1988

**Comment by Noppes** [comment]
From my understanding, the `environment mode batch` also includes the `print no-more` and `inhibit-alarms`, though it doesnt hurt to add them, if it works slightly different per ISAM product

**Comment by ktbyers** [comment]
@Noppes Thanks, yes that was what I was having difficulty determining (especially for the `inhibit-alarms`).

File: netmiko/nokia/nokia_isam.py [hunk 1]
```diff
@@ -12,8 +12,8 @@ def session_preparation(self) -> None:
         self._test_channel_read()
         self.set_base_prompt()
         commands = [
-            "environment inhibit-alarms",
-            "environment screen-length 0",
+            "environment mode batch",
+            "exit",
```
ktbyers, line 16: @Noppes Is the `exit` needed? I suspect this will cause you to logoff the device (as you are not entering config mode here).

Noppes: Whenever a environment command is used the base changes to: `base_name>environment#` with exit, you exit out of the >environment

PR #3777, opened 2026-02-26
Added cisco_wlc_ios

Noppes:
New cisco wlc devices run on cisco ios-xe, like the cisco c9800.

Since ios-xe is basically just ios, I named it wlc_ios instead of wlc_xe, but either would be fine with me.

### DIFF
--- a/netmiko/ssh_dispatcher.py
+++ b/netmiko/ssh_dispatcher.py
@@ -234,6 +234,7 @@ CLASS_MAPPER_BASE = {
     "cisco_tp": CiscoTpTcCeSSH,
     "cisco_viptela": CiscoViptelaSSH,
     "cisco_wlc": CiscoWlcSSH,
+    "cisco_wlc_ios": CiscoIosSSH,
     "cisco_xe": CiscoIosSSH,
     "cisco_xr": CiscoXrSSH,
     "cloudgenix_ion": CloudGenixIonSSH,
@@ -353,6 +354,7 @@ FILE_TRANSFER_MAP = {
     "cisco_asa": CiscoAsaFileTransfer,
     "cisco_ios": CiscoIosFileTransfer,
     "cisco_nxos": CiscoNxosFileTransfer,
+    "cisco_wlc_ios": CiscoIosFileTransfer,
     "cisco_xe": CiscoIosFileTransfer,
     "cisco_xr": CiscoXrFileTransfer,
     "dell_os10": DellOS10FileTransfer,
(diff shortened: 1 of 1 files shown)


ktbyers:
@Noppes Why do we need this, just use the `ios` or `ios-xe` device_type?

Noppes:
> @Noppes Why do we need this, just use the `ios` or `ios-xe` device_type?

I could have probably worded that better, it does have specific commands only those wlc ios devices have, so specific textfsm templates are needed for their devices types.

ktbyers:
Why can't these just be in `ios` or `ios-xe` in TextFSM templates (they are 'ios-xe' commands...sure, they are for access-points, but they are just 'ios/ios-xe' commands)?

What is an example command(s)?

I am worried about breaking all of the other IOS templates for anyone that uses the device_type (i.e. all the standard templates for IOS would break). Also I don't really like duplicating device_types unnecessarily.

pietos:
@ktbyers Problem we had is that people are using the cisco_wlc driver for the Cisco Wireless Lan Controllers. Only Cisco decided to move for the newer models from Aironet to IOS. But the commands on the IOS are not the same as the commands for Cisco IOS (routers and switches). So the prompt is identical, but the commands like show ap summery not.

Therefor we suggest to make a new WLC mapping toward the IOS driver, similar to what you did for cisco_xe. This way it's easier to distinguish that there is a cisco_wlc (aironet) and a cisco_wlc_ios (IOS).

This way we can also move the WLC templates from IOS towards WLC_IOS and is it a bit more clear what templates from NTC are tested on Routers & Switches and what commands are tested on the WLC.

Also the existing WLC driver we don't really dare to touch, it's quite a "challenging" code on the driver.

**Old WLC SNMP sysdescription output**

    Cisco 5520 Wireless Controller

**New WLC SNMP sysdescription  output**

 Cisco IOS Software [Dublin], C9800 Software (C9800_IOSXE-K9), Version 17.12.4, RELEASE SOFTWARE (fc3)\r\nTechnical Support: http://www.cisco.com/techsupport\r\nCopyright (c) 1986-2024 by Cisco Systems, Inc.\r\nCompiled Tue 23-Jul-24 09:44 by mcpre

ktbyers:
@pietos @Noppes Yuck, but okay (yuck with Cisco overloading commands between 9800 and other IOS platforms, but not surprising).

We can do `cisco_ioswlc`.

Yeah, the previous WLC os was ugly and pretty unpleasant for creating a Netmiko driver.

ktbyers (approved):

PR #3780, opened 2026-03-02
calix axos added

Noppes:
Added support for Calix AXOS devices

### DIFF
--- a/netmiko/calix/calix_axos.py
+++ b/netmiko/calix/calix_axos.py
@@ -23,7 +23,7 @@ class CalixAxosBase(CiscoSSHConnection):
     def set_base_prompt(
         self,
         pri_prompt_terminator: str = "#",
-        alt_prompt_terminator: str = None,
+        alt_prompt_terminator: str = "#",
         delay_factor: float = 1.0,
         pattern: Optional[str] = None,
     ) -> str:
--- a/netmiko/calix/calix_axos.py
+++ b/netmiko/calix/calix_axos.py
@@ -1,0 +1,63 @@
+"""Calix Axos SSH Driver for Netmiko"""
+
+from typing import Any, Optional
+
+from netmiko.cisco_base_connection import CiscoSSHConnection
+
+
+class CalixAxosBase(CiscoSSHConnection):
+
+    def session_preparation(self) -> Any:
+        """Prepare the session after the connection has been established."""
+        self.ansi_escape_codes = True
+        self._test_channel_read(pattern=r"#")
+        self.set_base_prompt()
+        self.enable()
+        self.set_terminal_width(command="terminal screen-width 512", pattern="terminal")
+        self.disable_paging(command="terminal screen-length 0")
+
+    def check_enable_mode(self, check_string: str = "#") -> bool:
+        """Check if in enable mode. Return boolean."""
+        return True
+
+    def set_base_prompt(
+        self,
+        pri_prompt_terminator: str = "#",
+        alt_prompt_terminator: str = None,
+        delay_factor: float = 1.0,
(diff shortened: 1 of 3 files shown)


netmiko/calix/calix_axos.py
```diff
@@ -0,0 +1,63 @@
...
+        self.ansi_escape_codes = True
+        self._test_channel_read(pattern=r"#")
+        self.set_base_prompt()
+        self.enable()
+        self.set_terminal_width(command="terminal screen-width 512", pattern="terminal")
+        self.disable_paging(command="terminal screen-length 0")
+
+    def check_enable_mode(self, check_string: str = "#") -> bool:
+        """Check if in enable mode. Return boolean."""
+        return True
```
ktbyers, line 21: Why do you have this hard-coded to True?

ktbyers (commented):
Everything besides the question on `check_enable_mode` looks good.

netmiko/calix/calix_axos.py
Noppes: Because you can't determine if you are in enabled or not, its always #, whether you have used enabled or not, so I figured since enabled is used in session_preparation, Ill just always have it return True.

I guess there are 2 alternatives, use a boolean and set it to true when enabled() is used and false with exit. Or in the check function have it use the help command and check if there are enabled only commands in the list.

ktbyers: @Noppes But you have to call `enable()` so there is a functional difference (but the prompt is unchanged)?

Noppes: Correct, the prompt is unchanged:
```login as: sysadmin
sysadmin@10.247.x.x's password: 
Last login: Wed Jan 28 09:44:52 2026 from 10.172.x.x
**** ACTIVE XG801 CARD  ****
Welcome to CLI
sysadmin connected from 10.172.x.x using ssh on E7-2
E7-2# enable
E7-2# configure
Enter configuration commands, one per line. End with CNTL/Z.
E7-2(config)# end
E7-2# show interface ethernet 1/1/q1 vlan
% No entries found.
E7-2# show interface ethernet 1/1/q1 bridge table
% No entries found.
E7-2# show interface lag la1 bridge table
INTERFACE  VLAN  CVLAN  MAC                SHELF  SLOT  AGE      LEARN    
--------------------------------------------------------------------------
la1        200   NA     04:bc:9f:ca:62:21  1      1     3451201  DYNAMIC  ```

ktbyers: Yes, but is there an actual change in state--in other words, do you have to make the .enable() call to go into configuration mode?

ktbyers: Okay, looking at this manual, it doesn't look to me like the enable() call is necessary (so we should remove that and add the `NoEnable` class to class inheritance). We should also remove that `check_enable_mode` method.

Once again, I am assuming this is not needed based on the docs and based on the device's behavior.

Here is the manual I was looking at:

https://www.calix.com/content/dam/calix/mycalix-misc/lib/iae/axos/21x/cli-o/index.htm?toc98408314.htm?92703.htm

Just let me know if this is not consistent with what you are observing.

Basically, what I don't like is we have a broken `enable()` mode implementation in this driver. If it supports `enable()` mode, then we need some way to detect it. If it doesn't support .enable() mode, then we should get rid of these code items.

ktbyers: @Noppes Any update on this? If you are not working on this any longer, we can also just close this PR.

ktbyers:
@Noppes I am going to close this as it doesn't seem like we are actively working on it. We can just re-open it, if there is active work on it.

PR #3798, opened 2026-04-08
Add Raisecom ROS Driver

DAQEM:
This diver is almost identical to the Raisecom ROAP driver but uses `config` or `config terminal` instead of `configure terminal` as can be seen on the [Scrapli Raisecom ROS Definition](https://github.com/carlmontanari/scrapli/blob/main/scrapli/definitions/raisecom_ros.yaml). It also uses `write` instead of `write startup-config` to save the config.

Tested on Raisecom ISCOM2948GF-4C-AC/D.

Didn't update the docs since that updated 250 other files as well.

### DIFF
--- a/PLATFORMS.md
+++ b/PLATFORMS.md
@@ -124,6 +124,7 @@
 - QuantaMesh
 - Rad ETX
 - Raisecom ROAP
+- Raisecom ROS
 - Silver Peak VXOA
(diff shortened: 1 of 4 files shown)


ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/3802

PR #3807, opened 2026-04-09
Fujitsu Sir driver

ktbyers:


ktbyers:
@caribouHY Note, I moved the `commit` command to the `save_config()` method and I also eliminated the explicit `commit` method.

I made this change as from review it sounds like the `commit` CLI command causes running-config to be saved to startup-config (essentially).

It also sounds like the `save` CLI command is more of an editing buffer flush (which you could just put on the end of your CLI commands in send_config_set)? 

Is this reasonable?

Note, I also renamed the driver / device_type to be consistent with Netmiko naming conventions.

ktbyers:
I am going to merge this and we can fix / edit any issues that need fixed.

caribouHY:
Thank you for merging the Si-R support!

 I noticed a compatibility issue with ntc-templates when using use_textfsm=True.

Netmiko device_type: fujitsu_sir
ntc-templates platform: fsas_sir

Because of this mismatch, TextFSM templates are not automatically applied.

ktbyers:
@caribouHY I will just add a second alias name using "fsas_sir".

PR #3809, opened 2026-04-10
Add fsas_sir as alias for fujitsu_sir device_type

ktbyers:
ntc-templates uses fsas_sir as the platform name; adding this alias ensures use_textfsm=True works properly


ktbyers:
FYI @caribouHY

PR #3812, opened 2026-04-13
Update a10 driver

ktbyers:


ktbyers:
@inaba-vdom-0 FYI, updated PR. FYI, I removed the override of `read_until_prompt()` methods as overriding that method should really ONLY happen if there are no other options.

PR #3827, opened 2026-04-14
Remove strip_command() override from FurukawaFitelnet driver

ktbyers:
The command echo should already be consumed by cmd_verify before strip_command() is called, so the override should be unnecessary.

### DIFF
--- a/netmiko/furukawa/furukawa_fitelnet.py
+++ b/netmiko/furukawa/furukawa_fitelnet.py
@@ -198,26 +198,6 @@ class FurukawaFitelnetBase(CiscoBaseConnection):
             )
         return output
 
-    def strip_command(self, command_string: str, output: str) -> str:
-        """Strip command echo from output.
-
-        FITELnet echoes the command with the prompt still on the same line
-        (e.g. ``TEST-HOST#show version`` or ``#show version``), so the base
-        implementation's ``output.startswith(cmd)`` check fails.  This
-        override additionally matches an echo line that ends with the command
-        after an optional prompt terminator (``>`` or ``#``).
-        """
-        cmd = command_string.strip()
-        if output.startswith(cmd):
-            return super().strip_command(command_string=command_string, output=output)
-
-        echo_pattern = rf"^.*?[>#]?{re.escape(cmd)}\s*$"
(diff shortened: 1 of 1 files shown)


ktbyers:
@SeiyaFunaokaJP FYI, I unintentionally merged your previous PR...after reviewing in more detail, the strip_command() method was the only part that I thought was problematic.

I am not seeing why overriding the parent method is necessary. Netmiko should have already read any prompt prior to the command being sent so I would think this all should be unnecessary.

If the prior prompt has not been read, then I think there is (likely) an issue elsewhere.

Can you show me real example of output you are seeing where the command is not properly being stripped (please include your Python code including CLI commands you are sending). I only need the code around where the issue happens. Also include output that is problematic (session_log would be helpful).

ktbyers:
@SeiyaFunaokaJP FYI, I merged this as I wanted the `develop` branch to be the state I thought it should be (and then we can fix with a subsequent PR, if necessary).

SeiyaFunaokaJP:
@ktbyers Thank you for the careful review and also for merging.
You are completely right — the strip_command() override is not necessary.

Sorry for the unnecessary code, and thank you for catching it quickly.
Please keep the current code as it is.


I re-tested on my real device with the current develop,                                             
and send_command("show version") returns clean output on SSH, Telnet and Serial.
                                                                                                    
I now understand the flow. I was confused about session_log and echo/strip flow:

<details>
<summary>My understanding</summary>

```
  ↓ Raw log (session_log)
  HOST-NAME#show version
  ↓ command_echo_read()  → strips "HOST-NAME#" here
  show version
  ↓ strip_command()      → only needs to strip the echo line (for special devices)

I expected HOST-NAME# to still be present when strip_command() is called,
but command_echo_read() has already removed it one step earlier.
```

</details>

FYI: Log files after latest commit
https://github.com/SeiyaFunaokaJP/netmiko-PR-log/blob/main/demo.py
https://github.com/SeiyaFunaokaJP/netmiko-PR-log/blob/main/demo_output.log
https://github.com/SeiyaFunaokaJP/netmiko-PR-log/blob/main/session_log_ssh.log

ktbyers:
@SeiyaFunaokaJP Okay, great and thanks for the contribution.

PR #3833, opened 2026-04-15
Add SSH and file transfer support for ZPE Nodegrid devices

jamestiotio:
Add the classes and helper functions to allow users to perform SSH connections and file transfers to ZPE Nodegrid devices.

These devices are using the Linux kernel. The only difference is that an additional step is needed to enter the Bash shell prompt from the CLI.

Tested with a real device and all tests in `test_netmiko_scp` are passing.

### DIFF
--- a/netmiko/ssh_dispatcher.py
+++ b/netmiko/ssh_dispatcher.py
@@ -184,6 +184,7 @@ from netmiko.vyos import VyOSSSH
 from netmiko.watchguard import WatchguardFirewareSSH
 from netmiko.yamaha import YamahaSSH
 from netmiko.yamaha import YamahaTelnet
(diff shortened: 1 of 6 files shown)


ktbyers:
@jamestiotio So it seems like the only purpose of this is for the file transfer? The SSH driver itself does not look to be fully implemented?

jamestiotio:
@ktbyers The main use case that we have is the file transfer functionality, so that was the part that I prioritized. If you prefer the SSH driver to be more fleshed out, I can do that.

ktbyers:
@jamestiotio PR superseded by: https://github.com/ktbyers/netmiko/pull/3835

PR #3835, opened 2026-04-15
Zpe nodegrid driver

ktbyers:


ktbyers:
@jamestiotio You probably should regression test that your SCP file transfer still works. 

Also let me know if you run into any issues with the normal SSH driver--I had to use the CLI behaviors of these types of devices without a device to test against.

ktbyers:
@jamestiotio FYI, I am doing some additional review. I will make a few more updates (shortly).

ktbyers:
@jamestiotio Okay, done reviewing...

ktbyers:
I am going to merge this and we can make modifications as needed.

PR #3840, opened 2026-04-24
Config mode fix in A10 driver

ktbyers:


ktbyers:
@inaba-vdom-0 Escaped the parenthesis so it would match literal "(config)" (since it is a regular expression pattern).

inaba-vdom-0:
@ktbyers Thank you very much! I tested it in my environment and confirmed that everything works without any issues.

PR #3841, opened 2026-04-24
Fix session_log flush so data is written immediately on write()

### LINKED ISSUE #3378: Session log file is empty for `fortinet` type devices until the remote connection is disconnected
### Description of Issue/Question

Since updating to 4.3.0 (from 4.2.1), we've noticed an issue wherein the session log file is not written to until the remote connection is disconnected. This issue has only been observed on `fortinet` device types so far (due to our environment, though it may occur on other device types). In the past, we would tail the session log to monitor the read/write activity while executing network automation scripts since it was a "live" view of the device CLI, and allowed our engineers to watch for any unexpected activity or errors.  

This issue has been reproduced on macOS 14.3, Ubuntu 22.04.3, and RHEL 7.9. Our current workaround has been to revert back to netmiko 4.2.1 for now.

### Netmiko version
(Paste verbatim output from `pip freeze | grep netmiko` between quotes below)

```
netmiko==4.3.0
```

### Netmiko device_type (if relevant to the issue)
(Paste `device_type` between quotes below)

```
fortinet
```

### Steps to Reproduce the Issue

```
Send any batch of commands and follow the session log with `tail -f <PATH-TO-LOG>`
```

ktbyers:
Previously, SessionLog.write() only buffered data in slog_buffer and relied on flush_session_log decorator or close() to persist it. This meant session log files were empty until disconnect. Moving flush() into write() ensures real-time logging. Removes the now-redundant flush_session_log decorator entirely.

Fixes #3378

### DIFF
--- a/netmiko/base_connection.py
+++ b/netmiko/base_connection.py
@@ -104,20 +104,6 @@ def lock_channel(func: F) -> F:
     return cast(F, wrapper_decorator)
 
 
-def flush_session_log(func: F) -> F:
-    @functools.wraps(func)
-    def wrapper_decorator(self: "BaseConnection", *args: Any, **kwargs: Any) -> Any:
-        try:
-            return_val = func(self, *args, **kwargs)
-        finally:
-            # Always flush the session_log
-            if self.session_log:
-                self.session_log.flush()
-        return return_val
-
-    return cast(F, wrapper_decorator)
-
-
 def log_writes(func: F) -> F:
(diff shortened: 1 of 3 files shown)


ktbyers:
The big issue here is the `no_log` behavior. Netmiko has to ensure we can't have things written into the session log that are specified as "no_log" and this implies it needs to be very careful with respect to no_log data that span multiple reads.

This was why the previous solution was implemented.

PR #3863, opened 2026-07-05
Fix hp_comware send_command empty output in Netmiko 4.x

fanbinbill:
Root cause: Netmiko 4.x send_command() relies on _prompt_handler() -> find_prompt() without the extra clear_buffer() safety net that existed in the 3.x flow. On H3C/HP Comware devices, find_prompt() can leave a residual <HOSTNAME> prompt fragment queued in the channel.

Reproduction: on hp_comware with auto_find_prompt enabled and cmd_verify disabled by default, send_command() may match the stale prompt before real command output arrives. strip_command()/strip_prompt() can then sanitize the buffer down to an empty string.

Fix: override hp_comware _prompt_handler() to clear_buffer() immediately after auto prompt detection. This keeps the change scoped to the Comware driver and avoids altering behavior for other platforms. Add unit tests covering auto_find_prompt on/off and find_prompt() fallback handling.,timeout:120}},{

Change-Id: I74d55cafef8890a8364109518484fdb52770615f

### DIFF
--- a/netmiko/hp/hp_comware.py
+++ b/netmiko/hp/hp_comware.py
@@ -84,6 +84,14 @@ class HPComwareBase(CiscoSSHConnection):
             bypass_commands=bypass_commands,
         )
 
+    def _prompt_handler(self, auto_find_prompt: bool) -> str:
+        prompt = super()._prompt_handler(auto_find_prompt)
+        if auto_find_prompt:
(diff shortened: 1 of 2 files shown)


ktbyers:
Superseded by https://github.com/ktbyers/netmiko/pull/3865

PR #3867, opened 2026-07-12
Fix Palo Alto command echo stripping

### LINKED ISSUE #3543: Palo Alto driver strip_command method is not doing a `strip()` call
This calls breakage for PA and telnet since <enter> might be \r\n and not just \n.

Should just get rid of the strip_command method from the driver and revert back to base_connection.py method (i.e. I can't see any reason to override the default method here).

H-TTTTT:
Fixes #3543

## Summary

- remove the Palo Alto-specific `strip_command` override
- use the shared base implementation for SSH and Telnet command echo handling
- cover LF/CRLF terminators and repeated command text

## Testing

- focused Palo Alto tests: 3 passed
- base connection unit tests: 35 passed
- feasible unit suite: 120 passed, 1 skipped, 5 dependency/environment deselections
- Ruff format and lint: passed
- mypy: passed for 270 source files

## Hardware

Not tested against a physical PAN-OS device.

### DIFF
--- a/netmiko/paloalto/paloalto_panos.py
+++ b/netmiko/paloalto/paloalto_panos.py
@@ -181,11 +181,6 @@ class PaloAltoPanosBase(NoEnable, BaseConnection):
             raise ValueError(f"Commit failed with the following errors:\n\n{output}")
         return output
 
-    def strip_command(self, command_string: str, output: str) -> str:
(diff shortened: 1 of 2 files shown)


ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/3873

Issue #3872, opened 2026-07-23
Telnet `read_very_eager()` is O(n²), making large telnet transfers extremely slow (~600x slower than SSH)


ga-sekiguchi:
## Summary

Reading large outputs over **telnet** is quadratically slow (O(n²)). A command
that returns a large amount of data takes **~20 minutes over telnet** but the
**same command over SSH finishes in ~2 seconds**.

The root cause is in the vendored telnetlib (`netmiko/_telnetlib/telnetlib.py`),
which was bundled into netmiko after `telnetlib` was removed from the Python
standard library in 3.13 (PEP 594). `read_very_eager()` accumulates data into
`cookedq` and re-copies the entire growing buffer on every loop iteration.

## Environment

- netmiko: 4.7.0
- Python: 3.13 (uses the vendored `netmiko/_telnetlib/telnetlib.py`)

## Reproduction

Read a large file (~136,000 lines) from a device over telnet:

```python
# telnet connection -> ~20 minutes for 136,411 lines
export_file_output = conn_telnet_test.send_command("cat test.csv", read_timeout=6000)
```

The identical read over an SSH connection to the same device completes in
**~2 seconds**. The larger the output, the worse the gap becomes (quadratic).

## Root cause

For telnet, `Channel.read_channel()` calls `read_very_eager()`:

```python
# netmiko/channel.py:122
return self.remote_conn.read_very_eager().decode(self.encoding, "ignore")
```

`read_very_eager()` loops calling `process_rawq()`, but never drains `cookedq`
until the final `read_very_lazy()`:

```python
# netmiko/_telnetlib/telnetlib.py:360
def read_very_eager(self):
    self.process_rawq()
    while not self.eof and self.sock_avail():
        self.fill_rawq()
        self.process_rawq()
    return self.read_very_lazy()
```

Each `process_rawq()` appends to `cookedq` with an immutable-`bytes` concat:

```python
# netmiko/_telnetlib/telnetlib.py:496
self.cookedq = self.cookedq + buf[0]
```

Because `cookedq` is **never emptied during the loop**, every iteration
re-copies the entire (and growing) `cookedq`. For `n` total bytes received in
chunks, the total copy work is `~n² / (2 * chunk)` = **O(n²)**.

Note: this is *independent* of the `recv(50)` chunk size. Shrinking `recv` does
not help, because `cookedq` keeps growing across loop iterations regardless of
how small each individual chunk is.

## Suggested fix

[...]

PR #3875, opened 2026-07-24
Fix telnet reads by looping read_eager() 

ktbyers:
TelnetChannel.read_channel() called telnetlib's read_very_eager(), which accumulates cooked data into a single immutable bytes buffer (cookedq) that is re-copied on every loop iteration. For n bytes received in chunks the total copy work is ~n^2/(2*chunk), making large telnet reads quadratic (GH #3872): a 136k-line read took ~20 min over telnet vs ~2 s over SSH.

Reimplement the read by looping the public read_eager() primitive, which drains cookedq on every call, and accumulate its results in a Python list joined once at the end. Each concat stays O(chunk) and the whole read is O(n). The vendored telnetlib is left untouched (pristine CPython copy).

read_eager() returns as soon as it has cooked data, so looping it until it yields nothing reproduces read_very_eager()'s greedy 'drain everything available' behavior. The EOFError contract (raised on a closed connection with no buffered data, relied on by telnet_login) is preserved by only re-raising when no data was collected.

Add tests/unit/test_telnet_channel.py: characterization tests that drive the real vendored Telnet parser via a fake socket and lock down multi-chunk draining, IAC option stripping/answering, IAC sequences split across recv() boundaries, the EOFError contract, and empty-read behavior.

### DIFF
--- a/netmiko/channel.py
+++ b/netmiko/channel.py
@@ -116,10 +116,41 @@ class TelnetChannel(Channel):
         raise NotImplementedError
 
     def read_channel(self) -> str:
-        """Read all of the available data from the channel."""
+        """Read all of the available data from the channel.
+
+        This recreates telnetlib's read_very_eager() behavior ("drain everything
+        currently available, without blocking") but as an O(n) operation.
+
+        read_very_eager() accumulates cooked data into a single immutable bytes
+        buffer (cookedq) that it re-copies on every loop iteration, which is
+        O(n^2) for large reads (see netmiko GH #3872). Instead we loop the
+        public read_eager() primitive, which drains cookedq on every call, and
+        accumulate its results in a Python list, joining once at the end. Each
+        concat therefore stays O(chunk) and the whole read is O(n).
+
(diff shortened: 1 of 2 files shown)


ktbyers:
Fixes https://github.com/ktbyers/netmiko/issues/3872

ktbyers:
@ga-sekiguchi Do you want to test this fix?

Note, I didn't want to make any changes to the vendored version of _telnetlib so I moved the fix into the Netmiko code.

ga-sekiguchi:
Thanks so much for putting this together, and for moving the fix into the Netmiko code instead of touching the vendored _telnetlib — that keeps the CPython copy pristine and is a much cleaner approach than my original suggestion.

  I tested the fix against my original repro (the ~136k-line cat over telnet that previously took over 20 minutes), and it now completes in about 10 seconds. Huge improvement — thank you! 🎉

  The characterization tests around IAC handling and the EOFError contract look great too. This is good to go from my side.

PR #3877, opened 2026-08-16
Make ruamel.yaml optional

jbemmel:
Netmiko currently declares ruamel.yaml as a mandatory dependency even though it is only required by specific functionality. This unnecessarily forces consumers such as Netlab to install ruamel when using Netmiko for normal network-device configuration. Make the dependency optional while keeping it available to the functionality that requires it. Update the development/test dependencies so Netmiko's own CI and pre-commit tests continue to run.

Unrelated, but needed to make the CI/CD checks pass: Update cryptography to 50.0.0. Can split that into a separate PR if desired


ipspace:
To add a bit of background:

> This unnecessarily forces consumers such as Netlab

We plan to use netmiko instead of Ansible whenever possible. This has been on the backburner since Ansible release 12, but I hope to get it done this summer. However, we would like to avoid a mandatory switch to ruamel. Details in https://github.com/ipspace/netlab/issues/3780

ktbyers:
Yes, this should be fine. 

I will review this and merge it early next week

ktbyers:
Superseded by: https://github.com/ktbyers/netmiko/pull/3878

PR #3878, opened 2026-08-18
Make ruamel.yaml optional

ktbyers:


ktbyers:
@jbemmel @ipspace I expanded the testing a bit (mostly tests on the original code).

PR #3888, opened 2026-09-21
Fortinet version detect whitespace issue

ktbyers:


belongOps:
Thanks so much. 
That is awesome.

